PR #4, opened 2025-01-28
Fix recognition of intrinsic gas error

jenikd:
This PR fixes situation, when gas estimation failed because of `intrinsic gas too low` error. Error was not correctly recognized, because raised error was from `go-ethereum` libary, not the one from `sonic` repository.

### DIFF
--- a/ethapi/api_test.go
+++ b/ethapi/api_test.go
@@ -2,20 +2,24 @@ package ethapi
 
 import (
 	"context"
+	"math/big"
+	"testing"
+
 	cc "github.com/0xsoniclabs/carmen/go/common"
 	"github.com/0xsoniclabs/carmen/go/common/amount"
 	"github.com/0xsoniclabs/carmen/go/common/immutable"
 	"github.com/0xsoniclabs/carmen/go/common/witness"
 	"github.com/0xsoniclabs/sonic/inter/state"
+	"github.com/0xsoniclabs/sonic/opera"
 	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/holiman/uint256"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
-	"math/big"
-	"testing"
 
 	"github.com/0xsoniclabs/sonic/evmcore"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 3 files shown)


HerbertJordan (dismissed):
Is there a way to add a unit test for this making sure we will not break it again?

evmcore/error.go
```diff
@@ -60,7 +62,7 @@ var (
 
 	// ErrIntrinsicGas is returned if the transaction is specified to use less gas
 	// than required to start the invocation.
-	ErrIntrinsicGas = errors.New("intrinsic gas too low")
+	ErrIntrinsicGas = fmt.Errorf("%w", core.ErrIntrinsicGas)
```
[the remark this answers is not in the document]
jenikd: Fixed

HerbertJordan (approved):
Thanks for adding a test!

ethapi/api_test.go
```diff
@@ -228,3 +232,57 @@ func getTestData() (*evmcore.EvmHeader, *types.Transaction, types.Receipts, erro
 	}
 	return header, transaction, receipts, nil
 }
+
+func TestEstimategas(t *testing.T) {
```
HerbertJordan, line 236: nit:
```suggestion
func TestEstimateGas(t *testing.T) {
```

ethapi/api_test.go
```diff
@@ -228,3 +232,57 @@ func getTestData() (*evmcore.EvmHeader, *types.Transaction, types.Receipts, erro
...
+	data := hexutil.Bytes(dataBytes)
+	transactionArgs := TransactionArgs{
+		From: &addr,
+		To:   &addr,
+		Data: &data,
+	}
+
+	gas, err := api.EstimateGas(context.Background(), transactionArgs, &blkNr)
+	require.NoError(t, err, "failed to estimate gas")
+	require.Greater(t, gas, uint64(0))
```
HerbertJordan, line 287: This test actually passes even if the change of the error definition is undone.
To match the PR, it should reproduce the issue fixed by this PR.

HerbertJordan (changes requested):

HerbertJordan (approved):

HerbertJordan (approved):

PR #8, opened 2025-02-03
Update version handling

HerbertJordan:
This PR introduces a few suggested changes to #6:
- the version string created in the `Makefile` is validated and parsed by a regex, improving resilience
- the version string created in the `Makefile` uses ":" instead of "-" to separate the version from the dirty and dev parameters to avoid mixing up meta strings with these flags
- the `Version` string was made private
- a `Version` struct allowing to access the parsed version information was added
- the version number features required by the `versionWatcher` (fixed a speller issue) got moved to the respective package
- the version number is now sorting pre-releases and releases into the right order
- the git commit and date values are now also tracked by the `version` package instead of the `config` package
- the git commit and date are now package private

Also, if a development version is detected, a minor version number one higher than the last tag is used. This is to fix the problem that a version `v2.3.4-dev` is considered less than a version `v2.3.4`. Thus, if the tag is `v2.3.4` and the build has extra commits, the version number `v2.4.0-dev` is used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,9 +9,9 @@ sonicd:
 	GIT_TAG=`echo $(call get_git_tag)` && \
 	GOPROXY=$(GOPROXY) \
 	go build \
-	    -ldflags "-s -w -X github.com/0xsoniclabs/sonic/config.GitCommit=$${GIT_COMMIT} \
-				        -X github.com/0xsoniclabs/sonic/config.GitDate=$${GIT_DATE} \
-						-X github.com/0xsoniclabs/sonic/version.Version=$${GIT_TAG}" \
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -91,10 +91,14 @@ define get_git_tag
         DIRTY_STATE=$$(git status --porcelain 2>/dev/null); \
         FINAL_TAG=$${GIT_TAG}; \
         if [ "$${COMMITS_SINCE}" -ne 0 ]; then \
-            FINAL_TAG=$${FINAL_TAG}-dev; \
+            FINAL_TAG="$${FINAL_TAG}:dev"; \
```
facuMH, line 94: the `:` could be appended in line 92, and then the else can be removed

facuMH (approved):

HerbertJordan:
Replaced by #14.

PR #9, opened 2025-02-04
configures ipc endpoint to explicitly point to /tmp

kjezek:
This PR updates integration test framework to explicitly point the IPC endpoint file to the `/tmp` directory. 

Unfortunately, the default temp directory on MacOS is somewhere else. The resulting path is too long, and the test fails on MacOS, because the endpoint cannot be created. The error message is:

```
WARN [02-04|09:18:07.985] The ipc endpoint is longer than 103 characters.  endpoint=/var/folders/xw/04df617n7rsfr5__h58dvzpr0000gp/T/TestSetStorage_PreExisting_Contract_Storage_Temporarily_Overridden3278411628/001/state/sonic.ipc
```

Since the` /tmp` dir is also available on MacOS, this PR explicitly sets the endpoint to be stored in this directory.


kjezek:
> I am a little conflicted about this change. I understand that it is needed, but is this the right place to implement it? The same issue would happen for the client run.
> 
> Would it be a better place to change this, at the point where the ipcpath variable is defaulted if not tested? wdyt?

I think for standard client run it is (usually) not a problem, because the IPC endpoint targets to the `datadir` directory. If one runs the client e.g. from his home, it is usually a short enough name. 

The issue is really only for integration tests where `datadir` is stored in a test temp dir, which is on MacOS under a pretty long path. 

Also the IPC endpoint path can be set as a program argument (as it is done in this change), i.e. users can alway reconfigure it.

tests/integration_test_net.go
```diff
@@ -250,6 +250,13 @@ func (n *IntegrationTestNet) start() error {
 		originalArgs := os.Args
 		defer func() { os.Args = originalArgs }()
 
+		// MacOS uses other temporary directories than Linux, which is a too long name for the Unix domain socket.
+		// Since /tmp is also available on MacOS, we can use it as a short temporary directory.
+		tmp, err := os.MkdirTemp("/tmp", "sonic_integration_test_*")
```
LuisPH3, line 255: I am a little reluctant to hardcode paths. 
Have you tried https://cs.opensource.google/go/go/+/go1.23.5:src/os/file.go;l=460 ?

kjezek: I think this would return the long path on MacOS, which I wanted to avoid. 

I used` /tmp `as it exists both on MacOs and Linux. It is not ideal, it would not work on Windows. What else could we use? Maybe a home dir? It should exist and be short enough everywhere?

LuisPH3 (approved):

tests/integration_test_net.go
HerbertJordan, line 255: From the documentation of [`MkdirTemp`](https://pkg.go.dev/os#MkdirTemp):

> It is the caller's responsibility to remove the directory when it is no longer needed.

I guess it would be good to delete the directory once it is no longer needed.

HerbertJordan (dismissed):

tests/integration_test_net.go
kjezek: oh good point. I was thinking about it, and verified that the files were indeed deleted, but I should have checked the documentation as well . Will update this.

LuisPH3 (approved):

HerbertJordan (approved):

PR #10, opened 2025-02-04
Replace library to obtain Total System Memory with simpler CGO call

HerbertJordan:
This PR replaces the [memory](https://github.com/pbnjay/memory) library, imported from the [github.com/pbnjay/memory](https://github.com/pbnjay/memory) repository by a simpler, local implementation.

This exchange is conducted to address issues identified by an external audit team. In particular, incorrect memory sizes have been reported. Furthermore, the library depends on the outdated [syscall](https://go.googlesource.com/proposal/+/refs/heads/master/design/freeze-syscall.md) library.

The new solution avoids the usage of the (unmaintained) `memory` package and the `syscall` package.

This change was tested on:
- linux (Ubuntu 22.04)
- darwin (macOS 14.6.1)


LuisPH3 (dismissed):
The code proposed queries available memory.
The routine used is posix, behavior in different distros should be equivalent.
No overflow check is done, that could happen in machines with exceptional large memory.

LuisPH3 (approved):

PR #11, opened 2025-02-04
handle unhandled errors and enable errcheck

facuMH:
This PR is part of https://github.com/Fantom-foundation/sonic-admin/issues/90

It contributes by handling errors that do not need the caution util, nor occur in tests. These could have a real production impact, so they should be thoroughly tested much like https://github.com/Fantom-foundation/Sonic/pull/378

- [x]  run make test
 Norma test
- [x]  demonet/static.yml
- [x]  demonet/dynamic.yml

- testing coverage https://docs.google.com/spreadsheets/d/1wma9IgKtWAePlPet0XdyoSQPRv5mU9tZspshbcd6jAc/edit?usp=sharing

this PR also enables `errcheck` as part of `make lint`

### DIFF
--- a/evmcore/tx_cacher.go
+++ b/evmcore/tx_cacher.go
@@ -20,6 +20,7 @@ import (
 	"runtime"
 
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/log"
 )
 
 // senderCacher is a concurrent transaction sender recoverer and cacher.
@@ -62,7 +63,9 @@ func newTxSenderCacher(threads int) *txSenderCacher {
 func (cacher *txSenderCacher) cache() {
 	for task := range cacher.tasks {
 		for i := 0; i < len(task.txs); i += task.inc {
-			types.Sender(task.signer, task.txs[i])
+			if _, err := types.Sender(task.signer, task.txs[i]); err != nil {
+				log.Warn("Failed to recover sender", "err", err)
+			}
 		}
 	}
 }
--- a/evmcore/tx_cacher.go
+++ b/evmcore/tx_cacher.go
@@ -20,7 +20,6 @@ import (
 	"runtime"
 
 	"github.com/ethereum/go-ethereum/core/types"
-	"github.com/ethereum/go-ethereum/log"
 )
 
 // senderCacher is a concurrent transaction sender recoverer and cacher.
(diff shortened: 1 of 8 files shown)


evmcore/tx_pool.go
```diff
@@ -1669,7 +1671,9 @@ func (pool *TxPool) demoteUnexecutables() {
 				log.Error("Demoting invalidated transaction", "hash", hash)
 
 				// Internal shuffle shouldn't touch the lookup set.
-				pool.enqueueTx(hash, tx, false, false)
+				if _, err := pool.enqueueTx(hash, tx, false, false); err != nil {
+					log.Warn("Failed to re-enqueue invalidated transaction", "hash", hash, "err", err)
```
LuisPH3, line 1675: for coherence with the previous log line, I would do the following:
- use log.Error just like in the previous line.
- Only log one of the two cases, it is either demoted (re-enqued) or dropped:
    - "Demoting invalidated transaction"
    - "Dropping invalidated transaction"

LuisPH3: I would not do any extra work here, since this is an inherited issue from [Geth](https://github.com/ethereum/go-ethereum/blob/e6f3ce7b168b8f346de621a8f60d2fa57c2ebfb0/core/txpool/legacypool/legacypool.go#L1697)

gossip/handler.go
LuisPH3, line 1: Gossip handler seems to report errors via log.Warning, :+1: 
I would say that this change is ok.

evmcore/tx_cacher.go
```diff
@@ -62,7 +63,9 @@ func newTxSenderCacher(threads int) *txSenderCacher {
 func (cacher *txSenderCacher) cache() {
 	for task := range cacher.tasks {
 		for i := 0; i < len(task.txs); i += task.inc {
-			types.Sender(task.signer, task.txs[i])
+			if _, err := types.Sender(task.signer, task.txs[i]); err != nil {
```
LuisPH3, line 66: This may be quite verbose and not very productive. 
I think it would be OK just to ignore this error (with a comment). Since the intended use is to cache values, if any to be cached. 
wdyt?

HerbertJordan: I agree, this might be a case where we can reasonably justify to ignore the error.

HerbertJordan (approved):
Looks good. Please address @LuisPH3 comments, then I think it is ready to merge.

evmcore/tx_cacher.go
```diff
@@ -62,7 +62,9 @@ func newTxSenderCacher(threads int) *txSenderCacher {
 func (cacher *txSenderCacher) cache() {
 	for task := range cacher.tasks {
 		for i := 0; i < len(task.txs); i += task.inc {
-			types.Sender(task.signer, task.txs[i])
+			// this error is ignored because we want to cache the sender if valid,
+			// and if not, it simply not cached
```
LuisPH3, line 66: Sender is cached inside transactions the first time it is queried.
Sender may fail if the signature verification fails, in this case the value will not be cached, any subsequent call to Sender will retrieve the same error, therefore it is safe to ignore errors here.

LuisPH3 (changes requested):

LuisPH3 (approved):
:pray:

PR #12, opened 2025-02-05
Replace best-effort port probing with OS selected port numbers

HerbertJordan:
We experience flaky tests due to the selection of port numbers in integration tests that have already been used by other processes on the system. This change fixes this by allowing the OS to assign any free port number.

**Background**
When passing the port number 0 through the CLI to the `sonicd` client, an arbitrary free port is assigned to server sockets (for the P2P service and the web service). This code change is using this mechanism to avoid port collisions.

However, to inform the integration test infrastructure about the selected port, a back-channel needed to be added. Through this `sonicd` instances are able to inform the test infrastructure about the selected service port.

**Detail**
While working on this PR it was discovered that the HTTP and WebSocket traffic can be channeled over the same server socket -- which is the default when no alternative socket is defined on the command line. Thus, the http-port and the ws-port got combined.

### DIFF
--- a/cmd/sonicd/app/app.go
+++ b/cmd/sonicd/app/app.go
@@ -2,10 +2,33 @@ package app
 
 import (
 	"os"
+
+	"gopkg.in/urfave/cli.v1"
 )
 
+// Run starts sonicd with the regular command line arguments.
 func Run() error {
-	initApp()
+	return RunWithArgs(os.Args, nil)
+}
+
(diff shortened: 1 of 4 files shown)


LuisPH3 (approved):
The changes proposed replace the best effort port selection mechanism with an OS decided port and a mechanism to propagate the choice back to the tests that require them.
A proxy app has been created to enable the feature only for tests.

PR #14, opened 2025-02-05
Align code version representation with release management policy

HerbertJordan:
Updates the internal representation of code versions to match the Sonic's Repository and Release Management policy.

Code versions are to be tracked by constants defined in [version.go](https://github.com/0xsoniclabs/sonic/blob/main/version/version.go). These constants cover the major, minor, and patch version, as well as an optional pre-release field. The following rules are enforced (by unit tests and code initialization code):
- the pre-release tag can only be empty, "dev", or "rc.N" where N is a natural number
- if the pre-release tag is "dev", the patch version must be 0 

#### Intended Use

When creating a new release branch, version constants are updated to match the first release candidate of the release branch. At the same time the minor version on the development branch should be incremented by 1.

After that, patch numbers and release candidates are incremented as needed on the release branch. Version number updates are not merged back into the main branch.

#### Additional code quality improvements
- the version number features required by the `versionWatcher` (fixed a speller issue) got moved to the respective package
- the version number is now sorting pre-releases and releases into the right order
- the git commit and date values are now also tracked by the `version` package instead of the `config` package
- the git commit and date are now package private


gossip/blockproc/verwatcher/version_number.go
```diff
@@ -0,0 +1,38 @@
+package verwatcher
```
facuMH, line 1: misssing license header.

gossip/blockproc/verwatcher/version_number_test.go
```diff
@@ -0,0 +1,66 @@
+package verwatcher
```
facuMH, line 1: missing license header

version/version.go
```diff
@@ -1,57 +1,127 @@
+// This package tracks Sonic's version number. The version number follows
```
facuMH, line 1: missing license header

version/version.go
```diff
@@ -1,57 +1,127 @@
...
-	return ToString(uint16((v/1e12)%1e6), uint16((v/1e6)%1e6), uint16(v%1e6))
+func (v Version) String() string {
+	res := fmt.Sprintf("%d.%d.%d", v.Major, v.Minor, v.Patch)
+	if v.PreRelease != "" {
+		res += "-" + v.PreRelease
+	}
+	return res
 }
+
+var preReleaseRE = regexp.MustCompile(`^(|dev|rc(\d+))$`)
```
facuMH, line 89: nit: declare all global variables together?

facuMH (commented):
looks good, just a couple of comments

gossip/blockproc/verwatcher/version_number.go
HerbertJordan: Most files in this project do not have license headers. I think those should be added consistently in an independent PR.

version/version.go
HerbertJordan: Since this is just a performance tweak for the function following immediately below, and not a global variable supposed to be used by anybody else, I think it should be defined just next to it. The section of constants defined on the top, to the contrary, are supposed to be public constants for users of the package. No need to mix them up.

gossip/blockproc/verwatcher/version_number.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func (v versionNumber) String() string {
+	version := version.Version{
+		Major: int(v>>48) & 0xffff,
+		Minor: int(v>>32) & 0xffff,
+		Patch: int(v>>16) & 0xffff,
+	}
+	if v&0xffff < 256 {
+		version.PreRelease = "pre"
```
LuisPH3, line 35: PR description says: "the pre-release tag can only be empty, "dev", or "rcN" where N is a natural number"
Is pre and dev mixed? 

(I think i prefer dev)

LuisPH3 (commented):
I think semver is a fairly stablish system and it is interesting to get it right for tooling compatibility.
I do not see the need of implementing our own rotuines given the existing pacakges:
https://pkg.go.dev/search?q=semver&m=

facuMH (dismissed):
after Luis' comment about pre-release, it seems good to merge

gossip/blockproc/verwatcher/version_number.go
HerbertJordan: The version-numbering was not able to track release candidates in the reviewed version, so `pre` was intended as a place-holder for version numbers presented to users that are either `dev` versions or release candidates. I originally considered adding support for release candidates in the version-watcher as out-of-scope.

However, I see that this is confusing, so I added support for this to this PR as well. Now the version tracker can differentiate between development versions, release candidates, and releases and orders them according to our release management policy (`dev` < `rc*` < releases).

HerbertJordan:
> I think semver is a fairly stablish system and it is interesting to get it right for tooling compatibility.
> I do not see the need of implementing our own rotuines given the existing pacakges:
> https://pkg.go.dev/search?q=semver&m=

There are a few things to be considered:
- the implementation for the version struct defined here restricts the version number to the subset of versions we are allowing on our repository (only `dev`, `rc*` and release versions); everything in the `version` package added in this PR aims to make sure that only such versions are created

- the version-watcher mechanism requires an `uint64` representation of a version number with a total order; for a general semantic version string (e.g. `1.2.3-rc3-dirty` or `1.2.3-altha.rc1`) this conversion would not be possible; the restriction to a subset of the version numbers (only `dev`, `rc*`, and releases) resolves this problem, since for those a numeric encoding can be easily found (see `versionNumber` type in the PR)

- a final problem is that the agreement in version-naming we have established has a subtile inconsistency compared to the official definition of semantic versioning [here](https://semver.org/); in particular, according to the official rules a version `1.2.3-rc12` would be less than `1.2.3-rc9` since the pre-release versions are expected to be compared lexicographically; to fix this, we would have to rename our tags to `1.2.3-rc.12` and `1.2.3-rc.9` (see code [example](https://go.dev/play/p/aQCeS0PecRt)).

LuisPH3 (approved):
:+1:

PR #21, opened 2025-02-06
Add feature flag infrastructure

HerbertJordan:
This PR introduces feature flags to safe-guard the development of new features.

Every new feature in the client should be guarded by a feature flag to facilitate backward compatibility. Only by enabling the feature, backward-compatibility may be sacrificed.

Feature flags are motivated by the need to be able to test for backward compatibility of the system. Furthermore, they provide an abstraction between network upgrades (aka. hard forks) and features. Upgrades group sets of features ready to be launched in one package, while others can be ignored. This way, features can be re-arranged between upgrades if the development progress is not matching the release cycle.


thaarok:
@HerbertJordan, are you sure we need this? We can add these features directly into Upgrades struct, at least as long as we fit into 64 bits/flags. (They are serialized into uint64 bitmap.) There is no limitation in order of "Upgrades" activation - all can be activated or deactivated by a transaction with no order limitations.

opera/features.go
```diff
@@ -0,0 +1,80 @@
+package opera
+
+import (
+	"slices"
+	"strings"
+)
+
+//go:generate stringer -type=Feature
```
LuisPH3, line 8: :+1:  very cool

LuisPH3 (dismissed):

HerbertJordan:
>@HerbertJordan, are you sure we need this? We can add these features directly into Upgrades struct, at least as long as we fit into 64 bits/flags. (They are serialized into uint64 bitmap.) There is no limitation in order of "Upgrades" activation - all can be activated or deactivated by a transaction with no order limitations.

@thaarok the motivation was two-fold: to keep the `Upgrades` struct clean, by only including hard-fork settings, and to reduce the network configuration complexity by reducing the number of switches that we can configure on the network. 

If we expose each feature individual, and allow to turn then on and off independently, we would have to make sure that any combination of those flags is supported -- by the current and any future client implementation.

I would hope that by only exposing the hard-forks in the upgrade flag list the configuration space remains limited. Originally I even tried to simplify the configuration to make the upgrade a numeric value instead (since modeling upgrades as booleans imposes the problem that it is unclear what should happen if Sonic is enabled but Berlin is not). However, this turned out to be a rabbit hole.

I think it is mandatory for us to assume that once a certain hard-fork / feature is enabled, in general, there is no way back. The hardfork may introduce state that older hard-forks can no longer handle (e.g. delegate designators in externally owned accounts).

LuisPH3 (commented):
I am not 100% sure how is this structure meant to be used. 
The current code base has a very fragmented source of rules, basically each module handles updated with its own set of boolean flags.

This PR presents a new tool, but does not integrate it. Generating yet another system to handle updates.

HerbertJordan:
Closed for being too late. Features are developed based on the `Allegoro` upgrade flag. Will resurrect the idea if there is renewed need for this.

PR #23, opened 2025-02-10
Update Github Actions workflows

thaarok:
* GITHUB_TOKEN is not needed to do a release (when running in the same repo) (currently is release [failing](https://github.com/0xsoniclabs/sonic/actions/runs/13198761118/job/36845805842#step:7:18) because of invalid token)
* No private repositories are in dependencies anymore - removing ACCESS_TOKEN/GOPRIVATE
* Upgrade Go in workflow to 1.23
* Removed disabled unit tests (already set to run 0-times, because moved out from github)

### DIFF
--- a/.github/workflows/release_build.yml
+++ b/.github/workflows/release_build.yml
@@ -12,22 +12,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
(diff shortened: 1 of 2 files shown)


thaarok:
@wsodsong pattern updated, rewritten into glob syntax.

PR #24, opened 2025-02-12
Update go-ethereum dependency to v1.15.0

HerbertJordan:
This PR updates Sonic's go-ethereum dependency to v1.15.0 in preparation for the Prague fork.

This PR updates the `go-ethereum`, `Tosca`, and `lachesis-base` dependencies in order to facilitate the usage of the updated go-ethereum version.

The main issues addressed during the update have been:
- the `state.StateDB` interface now includes a `Finalise` function that collided with the `Finalise` function included in Sonic's `state.StateDB` definition in the `inter` package. To resolve the conflict, Sonic's version was renamed to `EndTransaction`
- a range of methods of the `state.StateDB` interface got extended by new tracing flags and/or return values
- the `SkipAccountChecks` flag in the `core.Message` was split into a `SkipNonceChecks` and `SkipFromEOACheck` flags; these flags are required to exempt internal transactions and transactions received through RPC calls from some sets of pre-checks; the new flags have been adjusted accordingly
- the utilities function `math.BigMin` and `math.BigMax` got removed in go-ethereum [here](https://github.com/ethereum/go-ethereum/commit/a5fe7353c) in favor of an explicit comparison; this PR followed this recommendation;
- support for an override flag to allow the unlocking of an account was removed from go-ethereum [here](https://github.com/ethereum/go-ethereum/commit/a5f00018451e7f5eab156785d93de18244a27b66#diff-f38342eea15df77a3f465f7fddc5ea244be644e3b9c48191cfd3287fbf92b5ad); the corresponding account manager configuration flag `InsecureUnlockAllowed` is thus always considered to be false; a corresponding update was applied to the RPC code

Tasks to be completed before merge:
- [ ] have Aida updated to geth 1.15.0 to facilitate nightly tests


LuisPH3 (dismissed):

utils/rate/gauge.go
```diff
@@ -8,17 +8,17 @@ import (
...
 }
 
 // NewGauge constructs a new Gauge and launches a goroutine.
 // Be sure to call Stop() once the meter is of no use to allow for garbage collection.
 func NewGauge() *Gauge {
 	return &Gauge{
-		input: metrics.NewMeterForced(),
-		count: metrics.NewMeterForced(),
+		input: metrics.NewMeter(),
+		count: metrics.NewMeter(),
```
thaarok, line 21: The reason for NewMeterForced() is (or at least was), we need this gauge for emitter to work.
Regular NewMeter will not work if metrics collection is not enabled by using `--metrics` in os.Args.
(see https://github.com/0xsoniclabs/go-ethereum/blob/main/metrics/metrics.go#L28)
If we don't need Gauge for the emitter anymore, we can remove this Gauge struct completely.

HerbertJordan: The metric monitoring was refactored in go-ethereum, and if I understand correctly, a single metric processing go-routine is now managing metric updates in the background. Force-enabling this for this Gauge is not trivial.

Considering that this Gauge is no longer used in the emitter on the current network (we use a fixed-rate emitter logic), I would suggest to remove the Gauge all-together and eliminate this dependency of the emitter code to metrics.

I will prepare a PR testing out how extensive the changes would be to remove the Gauge from the code base.

HerbertJordan: The removal of the old heuristic is under evaluation in #27. Looks like this is not too difficult, but it requires some more evaluations.

LuisPH3 (dismissed):
I don't see obstacles, other than the ongoing discussion.

tests/state_test.go
```diff
@@ -37,6 +38,10 @@ var (
 		"Byzantium":         {},
 		"Frontier":          {},
 		"Homestead":         {},
+		"Istanbul":          {},
+		"Berlin":            {},
+		"London":            {},
+		"Prague":            {}, // < TODO: enable once supported (issue #65)
```
LuisPH3, line 44: missing cancun?

opera/rules.go
```diff
@@ -224,31 +224,30 @@ type UpgradeHeight struct {
...
+	IstanbulBlock:           big.NewInt(0),
+	MuirGlacierBlock:        big.NewInt(0), // EIP-2384: Muir Glacier Difficulty Bomb Delay - relevant for ethereum only
+	BerlinBlock:             nil,           // to be overwritten in EvmChainConfig
+	LondonBlock:             nil,           // to be overwritten in EvmChainConfig
+	ArrowGlacierBlock:       nil,           // EIP-4345: Difficulty Bomb Delay - relevant for ethereum only
+	GrayGlacierBlock:        nil,           // EIP-5133: Delaying Difficulty Bomb - relevant for ethereum only
+	MergeNetsplitBlock:      nil,
+	ShanghaiTime:            nil, // to be overwritten in EvmChainConfig
+	CancunTime:              nil, // to be overwritten in EvmChainConfig
+	PragueTime:              nil,
```
LuisPH3, line 246: PragueTime: nil, // to be overwritten in EvmChainConfig

tests/state_test.go
HerbertJordan: This list is unfortunately set up as the exclusion list ... so everything except `Cancun` and `Shanghai` is filtered out at the moment.

opera/rules.go
HerbertJordan: done

LuisPH3 (approved):

PR #25, opened 2025-02-12
Introduce BigMin / BigMax utils

HerbertJordan:
In geth 1.15 the former `math.BigMin` and `math.BigMax` utility functions got removed by this [PR](https://github.com/ethereum/go-ethereum/commit/a5fe7353c) due to its name-space collision with the standard-libary `math` module. The suggested replacement is to inline the min/max decision.

However, this results in weakened readability, as seen in #24. By inlining the code, the intention of computing the maximum or minimum of two integers is harder to derive. Also, mixing up the argument order or a `< 0` for a `> 0` for the result of the `Cmp()` function is an easy mistake (as has happened to the author of #24). 

To counter that, this PR introduces generalized replacements for these functions in the `utils` module to retain readability.

### DIFF
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -1349,7 +1349,7 @@ func newRPCTransaction(tx *types.Transaction, blockHash common.Hash, blockNumber
 		// if the transaction has been mined, compute the effective gas price
 		if baseFee != nil && blockHash != (common.Hash{}) {
 			// price = min(tip, gasFeeCap - baseFee) + baseFee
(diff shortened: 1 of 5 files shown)


LuisPH3 (dismissed):
The changes proposed solve the issue produced by the namespaces clash.
The naming is not ideal, can be a little misleading, but I cannot find a better solution.

Confusion comes from go itself:
The math namespace include maxType and MinType for all the standard types, but such are constants and not functions. The min/max operations equivalent to the ones used here are provided by a built-in overload, for lack of better expressivity of the language.

LuisPH3 (approved):

PR #27, opened 2025-02-14
Remove disabled event emitter heuristic

HerbertJordan:
As part of the Sonic network launch the emitter policy was updated in the Sonic client. The original emitter heuristic was replaced by a fixed-rate emitter policy, where every validator is emitting a new event every `X` ms, where `X` can be configured  through network rules.

### Emitter Policy Change
To be on the safe side, the Sonic client still allows to fall back to the previous heuristic by setting `X` to `0`. However, in practice the new fixed-rate emitter policy has been adapted. Furthermore, the integration of the certification chain will require regular event-emitting steps from all validators, which would require an extensive review of the legacy emitter heuristic.

To reduce complexity, this PR is eliminating the fall-back heuristic in favor of a fixed-rate policy in all cases -- as it is de facto used on the Sonic network.

### Performance Impact
This change also facilitates the skipping of the event-creation process earlier in the emitter loop. This significantly reduces CPU load, as seen in the following profile comparisons collected using Norma:

Before (main):
image

After (this PR):
image

### Tasks
- [x] run integration tests in Norma
- [x] run performance impact evaluation


thaarok (approved):
great thanks!

PR #33, opened 2025-02-19
Add Member and Committee definitions

HerbertJordan:
This PR introduces data definitions for a Committee authorized to sign certificates.

A committee is composed of a group of members following these rules:
 - there must be at least one member
 - the public keys of all members must be unique
 - all members must provide a proof-of-possession for their public keys
 - all members must have a non-zero voting power
 - the sum of the voting power must not exceed 2^64-1 (maximum of uint64)

In addition to the type definitions, debug printing and serialization support is added.

### DIFF
--- a/scc/committee.go
+++ b/scc/committee.go
@@ -16,7 +16,7 @@ import (
 // Among those statements is the confirmation of the succeeding period's
 // committee.
 //
-// Committees are composed of a ordered list of Members, each weighted with a
+// Committees are composed of an ordered list of Members, each weighted with a
 // non-zero voting power. Members are identified by their public keys, for which
 // they are required to provide a proof of possession.
 type Committee struct {
--- a/scc/committee.go
+++ b/scc/committee.go
@@ -1,0 +1,172 @@
+package scc
+
+import (
+	"fmt"
+	"slices"
+	"strings"
+
+	"github.com/0xsoniclabs/sonic/scc/bls"
+)
+
+// Committee is a group of Members that can sign statements to produce
+// certificates.
+//
+// On the Sonic Certification Chain (SCC), time is divided into periods. For
+// each period, a committee is authorized to sign the statement for that period.
+// Among those statements is the confirmation of the succeeding period's
+// committee.
+//
+// Committees are composed of a ordered list of Members, each weighted with a
+// non-zero voting power. Members are identified by their public keys, for which
+// they are required to provide a proof of possession.
+type Committee struct {
+	members []Member
+}
+
+// MaxCommitteeSize is the maximum number of members that can be in a committee.
+// The number needs to be limited to prevent Committee certificates from becoming
+// too large.
+const MaxCommitteeSize = 512
+
+// MemberId is used to identify a member in a committee by its position in the
+// ordered list of members.
+type MemberId uint16
+
+// NewCommittee creates a new committee with the provided members. There must be
+// at least one member, all members must be valid, and no duplicates are allowed.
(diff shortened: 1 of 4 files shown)


scc/committee.go
```diff
@@ -0,0 +1,172 @@
...
+	if int(id) >= len(c.members) {
+		return Member{}, false
+	}
+	return c.members[id], true
+}
+
+// GetMemberId returns the id of the member with the given public key. If the
+// public key is not found, the second return value is false.
+func (c Committee) GetMemberId(publicKey bls.PublicKey) (MemberId, bool) {
+	for id, m := range c.members {
```
facuMH, line 62: nit: `slices.IndecFunc` ?

scc/committee.go
```diff
@@ -0,0 +1,172 @@
...
+	return slices.Clone(c.members)
+}
+
+// GetMember returns the member with the given id. If the id is out of bounds,
+// the second return value is false.
+func (c Committee) GetMember(id MemberId) (Member, bool) {
+	if int(id) >= len(c.members) {
+		return Member{}, false
+	}
+	return c.members[id], true
```
facuMH, line 56: maybe it's further down in the PR, but is it somewhere enforced that the `MemberId` is the same as the index ? what would happen if a member is removed and later re-added ?

scc/committee.go
```diff
@@ -0,0 +1,172 @@
...
+	return 0, false
+}
+
+// Validate checks that the committee is well-formed. For a committee to be well
+// formed, the following properties need to be satisfied:
+// - The committee must have at least one member.
+// - The committee size must not exceed the maximum committee size.
+// - All members must be valid.
+// - No two members can have the same public key.
+// - The sum of the voting power must not exceed 2^64 - 1.
```
facuMH, line 76: should there not also be a lower limit for the voting power ?

scc/member.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// EncodedMember is a fixed-length byte array that represents a serialized member.
+type EncodedMember [48 + 96 + 8]byte
+
+// Serialize serializes the member into a fixed-length byte array.
+func (m Member) Serialize() EncodedMember {
+	res := EncodedMember{}
+	*(*[48]byte)(res[0:]) = m.PublicKey.Serialize()
+	*(*[96]byte)(res[48:]) = m.ProofOfPossession.Serialize()
```
facuMH, line 53: this seems overly complicated for something that should not be. is it possible to the casting on the right side to byte slice?

facuMH (approved):
overall looks good. a couple of nit comments, but feel free to answer them and merge.

scc/committee.go
HerbertJordan: The lower limit is 1. A member is invalid if its voting power is 0.

scc/member.go
HerbertJordan: Then the assignment would not work. There is no assignment operator for byte slices.

scc/committee.go
HerbertJordan: Committees are immutable. Members can not be added or removed.

scc/committee.go
HerbertJordan: This would not really be shorter or more readable:
```
       res := slices.IndexFunc(c.members, func(m Member) bool {
		return m.PublicKey == publicKey
	})
	if res < 0 {
		return 0, false
	}
	return MemberId(res), true
```

facuMH (approved):

PR #34, opened 2025-02-19
Add certificate definitions

HerbertJordan:
This PR introduces definitions for certification chain statements, signatures, and certificates.


scc/cert/certificate.go
```diff
@@ -0,0 +1,45 @@
+package cert
+
+import "github.com/0xsoniclabs/sonic/scc"
+
+// Certificate is a message signed by a committee to certify the validity of a
+// statement. A certificate is a claim produced by a authorized committee that a
```
facuMH, line 6:
```suggestion
// statement. A certificate is a claim produced by an authorized committee that a
```

scc/cert/certificate.go
```diff
@@ -0,0 +1,45 @@
...
+// statement is true. In the certification chain, this is used to establish
+// facts like the hash of a block at a certain height, the certificate chain's
+// state at the begin of an epoch, or the composition of a committee for a
+// certain period.
+type Certificate[S Statement] struct {
+	Subject   S
+	Signature AggregatedSignature[S]
```
facuMH, line 13: should these two be private ?

scc/cert/certificate_test.go
```diff
@@ -0,0 +1,65 @@
...
+		err := cert.Verify(committee)
+		if i < 4 { // needs more than 2/3 to be valid
+			requires.Error(err)
+		} else {
+			requires.NoError(err)
+		}
+	}
+}
+
+func TestCertificate_VerifyAuthority_AcceptsDifferentAuthority(t *testing.T) {
```
facuMH, line 38: should this test not check that 2 out of 3 members of a committee are needed to validate the certificate ?

scc/committee.go
```diff
@@ -35,13 +35,20 @@ type MemberId uint16
 // NewCommittee creates a new committee with the provided members. There must be
 // at least one member, all members must be valid, and no duplicates are allowed.
 func NewCommittee(members ...Member) (Committee, error) {
-	res := Committee{members: slices.Clone(members)}
+	res := NewCommitteeForTests(members...)
```
facuMH, line 38: based on the documentation for `NewCommitteeForTests`, it should not be used for production code, which `NewCommittee` looks like, right?

scc/committee_test.go
```diff
@@ -79,6 +79,29 @@ func TestCommittee_GetMemberId_ReturnsNotFoundIfNotPresent(t *testing.T) {
 	require.False(t, found)
 }
 
+func TestCommittee_GetTotalVotingPower_ReturnsCorrectTotal(t *testing.T) {
+	total, overflow := Committee{}.GetTotalVotingPower()
```
facuMH, line 83: should this one and the test below use `NewCommitteeForTests` ?

facuMH (commented):
overall looks good. Just a couple of comments/questions. Once those are clarified, it should be good to merge.

scc/cert/aggregated_signature.go
```diff
@@ -0,0 +1,91 @@
...
+	authority scc.Committee, // < committee for which a 2/3 majority is required
+	producers scc.Committee, // < committee that signed the aggregated signature
+	statement S, // < the statement to be verified
+) error {
+
+	// The producer committee must be a valid committee. In particular, for each
+	// public key in the producer committee, a valid proof of possession must be
+	// present. Otherwise, a modified public key could be used to counterfeit
+	// the aggregated signature.
+	if err := producers.Validate(); err != nil {
```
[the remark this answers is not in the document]
HerbertJordan: I do not want the signature checking code to be distributed throughout the code base. It is critical that the producers are validated before a signature is checked, and this should be documented here.

Also, committees can be obtained multiple ways: by creating them from a list of members, by loading them from the data base, or by retrieving them through the network. We would have to make sure that in all those cases the the inputs are properly verified and not modified in-between.

Moving the validation out of this code to reduce computation load is an optimization step. Let focus on clean, simple code for now, and improve performance when we have real-world benchmarks demonstrating performance problems.

scc/cert/aggregated_signature.go
```diff
@@ -0,0 +1,91 @@
...
+		return fmt.Errorf("invalid producer committee: %w", err)
+	}
+
+	// The voting power of the authority must not exceed the maximum.
+	totalPower, overflow := authority.GetTotalVotingPower()
```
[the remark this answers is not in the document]
HerbertJordan: > if the can the total voting power change ?

I do not understand this. Can you clarify?

scc/cert/certificate.go
HerbertJordan: Ok, will add getters as needed.

scc/cert/certificate_test.go
HerbertJordan: The 2/3 majority is in detail tested by the function above and the unit-tests for the aggregated signature. All this test is covering is that the certificate handles `authorities` and `producers` correctly, and is not mixing them up.

scc/cert/aggregated_signature.go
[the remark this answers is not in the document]
facuMH: it started as "what happens if the voting power changes?", but then I realized that since the committee is created, it should not change, so the question became "can the total voting power change?"

scc/committee_test.go
HerbertJordan: All of the tests in this file are creating the `Committee{}` directly, and as these are the unit tests specifically for this type, I think it is OK for them to do it like this. It is also less opaque, since no side effects can be hidden in the function call.

facuMH (approved):

scc/cert/aggregated_signature.go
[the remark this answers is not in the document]
HerbertJordan: No, since the list of members and their voting power is fixed and the total voting power is just the sum of the members voting power.

PR #40, opened 2025-02-21
Extend event serializing test for type 4 tx.

LuisPH3:
New transaction type can be serialized into RLP encoding. This PR extends event serialization test with events containing all existing transaction types.


facuMH (approved):
changes look good. Just an idea for a future PR/issue, looking at `makeAllTransactionTypes`, could we make the tx types as aggregated types? as in, AccessListTx is a LegacyTx plus accessList; Dynamic is accessList plus free market values, blob is Dynamic plus blob flieds, SetCode is Dyanmic plys AuthList. wdyt ?

LuisPH3:
> changes look good. Just an idea for a future PR/issue, looking at `makeAllTransactionTypes`, could we make the tx types as aggregated types? as in, AccessListTx is a LegacyTx plus accessList; Dynamic is accessList plus free market values, blob is Dynamic plus blob flieds, SetCode is Dyanmic plys AuthList. wdyt ?

It is not that easy: types package is directly imported from Geth, we are currently using their infrastructure for transaction creation as well.

There have been a number of attempts to facilitate construction of txs: 
- https://github.com/0xsoniclabs/sonic/blob/e5d12a1a872faaa44c758438a60da0cac0d32079/tests/gas_price_suggestion_test.go#L107
- https://github.com/0xsoniclabs/sonic/blob/e5d12a1a872faaa44c758438a60da0cac0d32079/tests/transaction_gas_price_test.go#L268

And for the introduction of `SetCodeTx` we will yet need something new, because we need to import the ABI filled inputData from another contract into this Tx. For this reason, so far this is the most direct and easier way to construct transactions. 
The learned lessons will be incorporated into serse [tx_builder](https://github.com/0xsoniclabs/serse/blob/main/serse/transaction_builder.go)

HerbertJordan (approved):
Nice, thanks!

facuMH (approved):

PR #42, opened 2025-02-21
Add hex slice serialization

facuMH:
This PR adds a hex string to byte slice general type with its Unmarshalling methog, a couple of byte array instances that use it.
This new type should facilitate the conversion to/from JSON for bls keys and signatures, but that work is left for future PRs.

The bitset is also extended to be de/serialized as well as un/marshall using the introduced `serialization.Hexbyte`.

### DIFF
--- a/scc/cert/bitset.go
+++ b/scc/cert/bitset.go
@@ -62,3 +62,30 @@ func (b BitSet[T]) String() string {
 type unsigned interface {
 	~uint8 | ~uint16 | ~uint32 | ~uint64
 }
+
+func (b *BitSet[T]) Serialize() []byte {
+	result := []byte{}
+	for _, entry := range b.Entries() {
+		result = append(result, byte(entry))
+	}
+	return result
+}
+
+func (b *BitSet[T]) Deserialize(data []byte) {
+	for _, entry := range data {
+		b.Add(T(entry))
+	}
+}
+
+func (b *BitSet[T]) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf("\"0x%x\"", b.Serialize())), nil
+}
+
+func (b *BitSet[T]) UnmarshalJSON(data []byte) error {
+	var mask []byte
+	if _, err := fmt.Sscanf(string(data), "\"0x%x\"", &mask); err != nil {
+		return err
+	}
+	b.Deserialize(mask)
+	return nil
+}
--- a/scc/cert/bitset.go
+++ b/scc/cert/bitset.go
@@ -3,6 +3,8 @@ package cert
 import (
 	"fmt"
 	"strings"
+
+	"github.com/0xsoniclabs/sonic/scc/cert/serialization"
 )
 
 // BitSet is a variable-size bit-mask based unsigned integer set representation.
@@ -82,10 +84,12 @@ func (b *BitSet[T]) MarshalJSON() ([]byte, error) {
 }
 
 func (b *BitSet[T]) UnmarshalJSON(data []byte) error {
-	var mask []byte
-	if _, err := fmt.Sscanf(string(data), "\"0x%x\"", &mask); err != nil {
+	hexBytes := serialization.HexBytes{}
+	err := hexBytes.UnmarshalJSON(data)
+	if err != nil {
 		return err
 	}
-	b.Deserialize(mask)
+
+	b.Deserialize(hexBytes)
 	return nil
 }
--- a/scc/cert/serialization/types.go
+++ b/scc/cert/serialization/types.go
@@ -19,6 +19,9 @@ func (h *HexBytes) UnmarshalJSON(data []byte) error {
 		*h = nil
 		return nil
 	}
+	if !strings.HasPrefix(s, "0x") {
+		return fmt.Errorf("invalid hex string %s", s)
+	}
 	s = strings.TrimPrefix(s, "0x")
 	if len(s)%2 == 1 {
 		s = "0" + s
@@ -31,11 +34,22 @@ func (h *HexBytes) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+func UnmarshallFixLenghtHexBytes(data []byte, length int) (HexBytes, error) {
+	var h HexBytes
+	err := h.UnmarshalJSON(data)
+	if err != nil {
+		return nil, err
+	}
+	if len(h) != length {
+		return nil, fmt.Errorf("invalid length %d, expected %d", len(h), length)
+	}
+	return h, nil
+}
+
 type PublicKey [48]byte
 
 func (p *PublicKey) UnmarshalJSON(data []byte) error {
-	hexBytes := HexBytes{}
-	err := hexBytes.UnmarshalJSON(data)
+	hexBytes, err := UnmarshallFixLenghtHexBytes(data, 48)
 	if err != nil {
 		return err
 	}
@@ -50,8 +64,7 @@ func (p PublicKey) String() string {
 type Signature [96]byte
 
 func (s *Signature) UnmarshalJSON(data []byte) error {
-	hexBytes := HexBytes{}
-	err := hexBytes.UnmarshalJSON(data)
+	hexBytes, err := UnmarshallFixLenghtHexBytes(data, 96)
 	if err != nil {
 		return err
 	}
--- a/scc/cert/serialization/types.go
+++ b/scc/cert/serialization/types.go
@@ -1,0 +1,64 @@
+package serialization
+
+import (
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+type HexBytes []byte
+
+func (h *HexBytes) UnmarshalJSON(data []byte) error {
+	var s string
+	err := json.Unmarshal(data, &s)
+	if err != nil {
+		return err
+	}
+	if s == "null" {
+		*h = nil
+		return nil
+	}
+	s = strings.TrimPrefix(s, "0x")
+	if len(s)%2 == 1 {
+		s = "0" + s
+	}
+	raw, err := hex.DecodeString(s)
+	if err != nil {
+		return err
+	}
+	*h = raw
+	return nil
+}
+
+type PublicKey [48]byte
+
+func (p *PublicKey) UnmarshalJSON(data []byte) error {
+	hexBytes := HexBytes{}
+	err := hexBytes.UnmarshalJSON(data)
+	if err != nil {
+		return err
+	}
+	*p = PublicKey(hexBytes)
+	return nil
+}
+
+func (p PublicKey) String() string {
+	return fmt.Sprintf("0x%x", []byte(p[:]))
+}
+
+type Signature [96]byte
+
+func (s *Signature) UnmarshalJSON(data []byte) error {
+	hexBytes := HexBytes{}
+	err := hexBytes.UnmarshalJSON(data)
+	if err != nil {
+		return err
+	}
+	*s = Signature(hexBytes)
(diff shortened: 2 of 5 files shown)


scc/cert/serialization/types.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if len(h) != length {
+		return nil, fmt.Errorf("invalid length %d, expected %d", len(h), length)
+	}
+	return h, nil
+}
+
+type PublicKey [48]byte
```
HerbertJordan, line 49: There is nothing specific about this type that restricts its use to public keys. Anything that represents 48-byte could use this utility type for JSON serialization. I would thus suggest to use a more generic name like `HexBytes48`; same for the `Signature` below.

scc/cert/serialization/types_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestHexBytes_UnmarshalJSON_ValidHexString(t *testing.T) {
+	var h HexBytes
+	data := []byte(`"0x012abc"`)
+	require.NoError(t, h.UnmarshalJSON(data))
+	data = []byte(`"null"`)
+	require.NoError(t, h.UnmarshalJSON(data))
+}
```
HerbertJordan, line 15: This test misses a case with an odd number of hex-characters.

HerbertJordan (changes requested):
This PR needs some improvements. Please see the comments.

Also, please improve the PR description.

> This PR adds a hex string to byte slice general type with its Unmarshalling method, a couple of byte array instances that use it.

Ideally, the first line should be something somebody can understand when reviewing this in a year without prior knowledge what this PR is about.

scc/cert/bitset.go
```diff
@@ -62,3 +64,32 @@ func (b BitSet[T]) String() string {
 type unsigned interface {
 	~uint8 | ~uint16 | ~uint32 | ~uint64
 }
+
+func (b *BitSet[T]) Serialize() []byte {
+	result := []byte{}
+	for _, entry := range b.Entries() {
+		result = append(result, byte(entry))
+	}
```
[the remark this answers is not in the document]
facuMH: using b.mask as serialized result to avoid this issue.

scc/cert/serialization/types.go
facuMH: using `HexBytes48` and `HexBytes96`

scc/cert/serialization/types_test.go
facuMH: test case added

scc/cert/serialization/types.go
HerbertJordan, line 1: This package name should somehow provide a hint that it is about JSON encoding. Serialization is usually referring to a binary representation.

scc/cert/serialization/types.go
```diff
@@ -0,0 +1,111 @@
...
+// MarshalJSON converts the HexBytes into a JSON-compatible hex string with a "0x" prefix.
+func (h HexBytes) MarshalJSON() ([]byte, error) {
+	if h == nil {
+		return []byte("null"), nil
+	}
+	return []byte(fmt.Sprintf("\"0x%x\"", []byte(h))), nil
+}
+
+// UnmarshalJSON parses a JSON hex string into a HexBytes slice.
+// The input string must have a "0x" prefix and be an even-length hex string.
```
HerbertJordan, line 22: This is actually not true. The implementation accepts odd-length inputs, as well as the special case of a "null" value.

scc/cert/serialization/types.go
```diff
@@ -0,0 +1,111 @@
...
+		return err
+	}
+	*h = raw
+	return nil
+}
+
+// String returns the hex string representation of HexBytes.
+func (h HexBytes) String() string {
+	return fmt.Sprintf("0x%x", []byte(h))
+}
```
HerbertJordan, line 51: nit: is this required? (same for the other `String()` functions in this file)

Also: this one treats `nil` values different than the `MarshalJSON` function. This might lead to confusion.

scc/cert/serialization/types_test.go
```diff
@@ -0,0 +1,101 @@
...
+	h = nil
+	data, err = h.MarshalJSON()
+	require.NoError(t, err)
+	require.Equal(t, []byte("null"), data)
+}
+
+func TestHexBytes_UnmarshalJSON_ValidHexString(t *testing.T) {
+	var h HexBytes
+	data := []byte(`"0x012abc"`)
+	require.NoError(t, h.UnmarshalJSON(data))
```
HerbertJordan, line 23: This test should also check that h has the expected value after the unmarshalling.

scc/cert/serialization/types_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+func TestHexBytes48_MarshalJSON(t *testing.T) {
+	p := HexBytes48([48]byte{0x01})
+	data, err := p.MarshalJSON()
+	require.NoError(t, err)
+	expected := `"0x010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"`
+	require.Equal(t, []byte(expected), data)
+}
+
+func TestHexBytes48_UnmarshalJSON_ShortHexString(t *testing.T) {
```
HerbertJordan, line 54: By our test name [convention](https://github.com/Fantom-foundation/Tosca/issues/883) the name of a test like this should also state the intended behavior. Something like
```suggestion
func TestHexBytes48_UnmarshalJSON_TooShortHexStringIsRejected(t *testing.T) {
```

The other names of tests in this file would benefit from a similar renaming.

scc/cert/bitset.go
```diff
@@ -62,3 +64,31 @@ func (b BitSet[T]) String() string {
...
+// Serialize returns the byte representation of the BitSet.
+func (b *BitSet[T]) Serialize() []byte {
+	return b.mask
+}
+
+// Deserialize restores the BitSet from the given byte representation.
+func (b *BitSet[T]) Deserialize(data []byte) {
+	b.mask = make([]byte, len(data))
+	copy(b.mask, data)
+}
```
HerbertJordan, line 77: Are those needed?

scc/cert/bitset.go
```diff
@@ -62,3 +64,31 @@ func (b BitSet[T]) String() string {
...
+
+// MarshalJSON converts the BitSet into a JSON-compatible hex string.
+func (b *BitSet[T]) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf("\"0x%x\"", b.Serialize())), nil
```
HerbertJordan, line 81: For symmetry, this should also use the serialization utility, as the UnmarshalJSON is doing. In particular, this would introduce the special case support for `nil` values.

HerbertJordan (changes requested):

scc/cert/serialization/types_test.go
facuMH: that is done in line 28

scc/cert/serialization/types.go
facuMH: "required" might be a strong term, but it does facilitate testing.

scc/cert/json/types.go
```diff
@@ -0,0 +1,114 @@
...
+// MarshalJSON converts the HexBytes into a JSON-compatible hex string with a "0x" prefix.
+func (h HexBytes) MarshalJSON() ([]byte, error) {
+	if h == nil {
+		return []byte("null"), nil
+	}
+	return []byte(h.String()), nil
+}
+
+// UnmarshalJSON parses a JSON hex string into a HexBytes slice.
+// The input string must have a "0x" prefix.
```
HerbertJordan, line 22: nit: "null" does not have a "0x" prefix

scc/cert/bitset.go
```diff
@@ -62,3 +64,20 @@ func (b BitSet[T]) String() string {
...
+
+// UnmarshalJSON parses a JSON hex string into a BitSet.
+func (b *BitSet[T]) UnmarshalJSON(data []byte) error {
+	hexBytes := json.HexBytes{}
+	err := hexBytes.UnmarshalJSON(data)
+	if err != nil {
+		return err
+	}
+	b.mask = make([]byte, len(hexBytes))
+	copy(b.mask, hexBytes)
```
HerbertJordan, line 81: nit: no need to copy
```suggestion
	b.mask = hexBytes[:]
```

scc/cert/json/types.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+// HexBytes represents a byte slice that is serialized as a hexadecimal string with a "0x" prefix.
+type HexBytes []byte
+
+// MarshalJSON converts the HexBytes into a JSON-compatible hex string with a "0x" prefix.
+func (h HexBytes) MarshalJSON() ([]byte, error) {
+	if h == nil {
+		return []byte("null"), nil
+	}
```
HerbertJordan, line 17: not needed any more
```suggestion
```

scc/cert/json/types_test.go
```diff
@@ -0,0 +1,114 @@
...
+	data := []byte(`"0x012abc"`)
+	require.NoError(t, h.UnmarshalJSON(data))
+	data = []byte(`"null"`)
+	require.NoError(t, h.UnmarshalJSON(data))
+	data = []byte(`"0x12abc"`)
+	require.NoError(t, h.UnmarshalJSON(data))
+	require.Equal(t, HexBytes([]byte{0x1, 0x2a, 0xbc}), h)
+}
+
+func TestHexBytes_UnmarshalJSON_InvalidHeString_ProducesError(t *testing.T) {
```
HerbertJordan, line 35:
```suggestion
func TestHexBytes_UnmarshalJSON_InvalidHexString_ProducesError(t *testing.T) {
```

scc/cert/json/types_test.go
```diff
@@ -0,0 +1,114 @@
...
+	data := []byte(`"0x1234567"`)
+	err := p.UnmarshalJSON(data)
+	require.Error(t, err)
+}
+
+func TestHexBytes48_UnmarshalJSON_ValidHexString(t *testing.T) {
+	var p HexBytes48
+	data := []byte(`"0x0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"`)
+	err := p.UnmarshalJSON(data)
+	require.NoError(t, err)
```
HerbertJordan, line 78: it does not really check that the `p` contains the parsed value.

scc/cert/json/types_test.go
```diff
@@ -0,0 +1,114 @@
...
+	data := []byte(`0x1234567`)
+	require.Error(t, s.UnmarshalJSON(data))
+	data = []byte(`0x0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef`)
+	require.Error(t, s.UnmarshalJSON(data))
+}
+
+func TestHexBytes96_UnmarshallJSON_ValidHexString(t *testing.T) {
+	var s HexBytes96
+	data := []byte(`"0x0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"`)
+	require.NoError(t, s.UnmarshalJSON(data))
```
HerbertJordan, line 107: also here, no real check that the result is what was given as an input.

HerbertJordan (changes requested):

HerbertJordan (approved):

PR #43, opened 2025-02-21
Add distinct type to represent SCC periods

HerbertJordan:
This adds a distinct type for referencing SCC periods as a special version of a `uint64` to improve type safety when referencing to periods.


facuMH (dismissed):
nice addition, thanks.

facuMH (approved):

PR #45, opened 2025-02-21
Use pre-existing type for block number in certificates

HerbertJordan:
This PR replaces the basic `uint64` type used for identifying block numbers by the `idx.Block` type used in the rest of the project.


facuMH (approved):
simple replace. if tests pass, looks good.

PR #49, opened 2025-02-24
Update Go version to 1.24

HerbertJordan:
By updating to this Go version we get support for the new features of [1.23](https://tip.golang.org/doc/go1.23) and [1.24](https://tip.golang.org/doc/go1.24), including:
- [iterator](https://tip.golang.org/doc/go1.23#language) support in for-range loops
- extensions to the [slices](https://pkg.go.dev/slices) library
- generic type aliases
- improvements in various tools like `vet` or `cgo`


LuisPH3 (approved):
:+1:

PR #50, opened 2025-02-24
Add support for iterating through certificates to DB

HerbertJordan:
This PR introduces support for iterating through certificates in their native order to the certification store.


facuMH (approved):
nice use of itrs

PR #52, opened 2025-02-24
Add utility for encoding streams of objects

HerbertJordan:
This PR adds a utility for encoding a sequence of serializable objects into a stream.

When encoding sequences of SCC certificates into a stream, it is necessary to explicitly keep track of the boundaries of the encoding of individual entries. This is due to the fact that the utilized protobuf encoding does not encode data termination internally, and thus can not identify object boundaries implicitly.

This PR introduces a simple utility facilitating the encoding and decoding of a sequence dynamically sized elements into a single stream.


utils/objstream/objstream_test.go
```diff
@@ -0,0 +1,103 @@
...
+			break
+		}
+		require.NoError(err)
+		output = append(output, string(str))
+	}
+
+	require.Equal(input, output)
+}
+
+type msg string
```
facuMH, line 94:
```suggestion
type testMsg string
```
or something else that high lights that this type is for testing.

facuMH (commented):
Nice utility, just one minor change needed.

facuMH (approved):

PR #53, opened 2025-02-24
Add Member Json type

facuMH:
This PR adds a type `MemberJson` which serves as an intermediate step to converting members to json.


scc/cert/member_json.go
```diff
@@ -0,0 +1,48 @@
...
+type MemberJson struct {
+	PublicKey         jsonhex.Bytes48 `json:"publicKey" gencodec:"required"`
+	ProofOfPossession jsonhex.Bytes96 `json:"proofOfPossession" gencodec:"required"`
+	Weight            uint64          `json:"weight" gencodec:"required"`
+}
+
+// String returns the JSON string representation of the MemberJson.
+func (m MemberJson) String() string {
+	return fmt.Sprintf(`{"publicKey":"%v","proofOfPossession":"%v","weight":%d}`,
+		m.PublicKey, m.ProofOfPossession, m.Weight)
```
HerbertJordan, line 21: nit: this is likely equivalent to what I would get if I print the object using "%v" or "%+v" without a `String()` implementation.

scc/cert/member_json.go
HerbertJordan, line 1: Since the `Member` type is defined in the `scc` package, this file should likely also be moved there. Wdyt?

scc/cert/member_json_test.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+func TestMemberJson_String(t *testing.T) {
+	m := MemberJson{}
+	expected := fmt.Sprintf(`{"publicKey":"%v","proofOfPossession":"%v","weight":%d}`,
+		jsonhex.Bytes48{}, jsonhex.Bytes96{}, uint64(0))
+	require.Equal(t, expected, m.String())
+}
+
+func TestMemberJson_ToMember_InvalidPublicKey(t *testing.T) {
```
HerbertJordan, line 20:
```suggestion
func TestMemberJson_ToMember_FailsOnInvalidPublicKey(t *testing.T) {
```

scc/cert/member_json_test.go
```diff
@@ -0,0 +1,70 @@
...
+	m := MemberJson{
+		PublicKey:         jsonhex.Bytes48{},
+		ProofOfPossession: jsonhex.Bytes96(bls.Signature{}.Serialize()),
+		Weight:            uint64(0),
+	}
+	_, err := m.ToMember()
+	require.Error(t, err)
+}
+
+func TestMemberJson_ToMember_InvalidProofOfPossession(t *testing.T) {
```
HerbertJordan, line 30:
```suggestion
func TestMemberJson_ToMember_FailsOnInvalidProofOfPossession(t *testing.T) {
```

scc/cert/member_json_test.go
```diff
@@ -0,0 +1,70 @@
...
+	m := MemberJson{
+		PublicKey:         jsonhex.Bytes48(bls.NewPrivateKey().PublicKey().Serialize()),
+		ProofOfPossession: jsonhex.Bytes96{},
+		Weight:            uint64(0),
+	}
+	_, err := m.ToMember()
+	require.Error(t, err)
+}
+
+func TestMemberJson_ToMember_Valid(t *testing.T) {
```
HerbertJordan, line 40:
```suggestion
func TestMemberJson_ToMember_AcceptsValidInput(t *testing.T) {
```

HerbertJordan (changes requested):
Looks good, but it should be moved to a different package.

Also, we might just add a `MarshalJSON` and `UnmrashalJSON` to the `Member` type to make use easier. Wdyt?

scc/cert/member_json.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/0xsoniclabs/sonic/scc"
+	"github.com/0xsoniclabs/sonic/scc/bls"
+	"github.com/0xsoniclabs/sonic/utils/jsonhex"
+)
+
+// MemberJson is a JSON friendly representation of a scc.Member.
+type MemberJson struct {
+	PublicKey         jsonhex.Bytes48 `json:"publicKey" gencodec:"required"`
+	ProofOfPossession jsonhex.Bytes96 `json:"proofOfPossession" gencodec:"required"`
+	Weight            uint64          `json:"weight" gencodec:"required"`
```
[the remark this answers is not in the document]
facuMH: no, I added it on the first attempt to try it out. will remove next.

scc/cert/member_json.go
facuMH: almost. using "%+v" does not add the needed `"` for all the names and non numerical values, which we need. Also does not add a `,` after each value.

scc/cert/member_json.go
facuMH: moving it there.

HerbertJordan (changes requested):

facuMH:
This PR will not be merged. instead there will be a PR for bls.Signature encoding and then use that in a future work on certificate encoding.

PR #57, opened 2025-02-24
Add infrastructure for the passive generation of certificates

HerbertJordan:
This PR introduces the infrastructure needed for the passive generation of certificates.

The central element of this PR is a `Node` definition, which is going to be tasked with tracking the SCC state on each individual nodes. In this first iteration it is only prepared to track produced blocks locally and creating (unsigned) certificates for blocks and committees in its local store.

This is the first development step, aiming to obtain a setup in which certificates are created and available for RPC queries. Follow-up steps cover the active parts of creating block block and committee signatures as well as the passive aggregation thereof.


gossip/store_certificates_test.go
```diff
@@ -5,10 +5,13 @@ import (
 
 	"github.com/0xsoniclabs/sonic/scc"
 	"github.com/0xsoniclabs/sonic/scc/cert"
+	scc_node "github.com/0xsoniclabs/sonic/scc/node"
 	"github.com/Fantom-foundation/lachesis-base/inter/idx"
 	"github.com/stretchr/testify/require"
 )
 
+var _ scc_node.Store = (*Store)(nil)
```
facuMH, line 13: why is this dependency needed ?

scc/node/node.go
```diff
@@ -0,0 +1,44 @@
...
+// Node is a node in the Sonic Certification Chain. It is responsible for
+// handling the progression of the chain by responding to new block statements
+// and creating new certificates.
+type Node struct {
+	store Store
+	state State
+}
+
+// NewNode creates a new node with the given store and committee. The provided
+// committee is expected to be the active committee of the current block.
```
facuMH, line 19: expected ? There is no way of quickly verifying this, right ? just brainstorming here, but since this method is not something that would be called many times, it could perform a heavy task such as verifying the with a couple of RPC calls that the committee is the active one, no ?

facuMH (approved):
overall looks good. just a couple of questions.

gossip/store_certificates_test.go
HerbertJordan: In the near future the `gossip.Store` is supposed to serve as the `node.Store` for the one `Node` instance managing the SCC. Until this is integrated, there is no implicit check that makes sure that the `gossip.Store` is implementing this interface. With this line I add this requirement explicitly on the `gossip.Store` implementation.

scc/node/node.go
HerbertJordan: The source for this committee should be the local node's DB, which is -- from the perspective of the local node -- a trusted source, filled through one of many possible synchronization procedures:
- a full-sync from block 0
- a G-file based sync from a given block height
- a light-client protocol based sync

Any of these procedures should be carried out to fill the content of the local DB, before this new `Node` is created.

Also, this is not the final version of this function, as the factory parameters will evolve over the course of implementing this feature.

facuMH: aahh, I thought this node would be the one holding the state for the client to sync in.

PR #58, opened 2025-02-25
Improve error reporting in DB content iteration

HerbertJordan:
This PR improves the certification enumeration feature of the gossip store by adding support for reporting errors during the enumeration process.

To do so, this PR introduces an new, generic "Result" utility, inspired by similar features found in other languages.

### DIFF
--- a/utils/result/result.go
+++ b/utils/result/result.go
@@ -13,8 +13,8 @@ func New[V any](value V) T[V] {
 }
 
 // Error creates a new result with an error. This is the type of result that is
-// returned when the operation is unsuccessful. Note, passing nil will as a
-// an error will result in a non-error result containing the zero value.
+// returned when the operation is unsuccessful. Note, passing nil as an error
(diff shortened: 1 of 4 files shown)


utils/result/result.go
```diff
@@ -0,0 +1,32 @@
...
+
+// New creates a new result with a value. This is the type of result that is
+// returned when the operation is successful.
+func New[V any](value V) T[V] {
+	return T[V]{value: value}
+}
+
+// Error creates a new result with an error. This is the type of result that is
+// returned when the operation is unsuccessful. Note, passing nil will as a
+// an error will result in a non-error result containing the zero value.
```
facuMH, line 17: "passing nil will as a an error will result ..." ?

facuMH (dismissed):
there is confusing line in the documentation of a function, otherwise looks like a nice optional

facuMH (approved):

PR #59, opened 2025-02-25
add json encoding to public key and signature

facuMH:
This PR:

- Fixes a formatting issue in jsonhex formatting issue.
- Adds a test in jsonhex that uses `json.Marshall` and `json.Unmarshall` to ensure this issues are caught.
- Adds json encoding functions for `bls.PublicKey` and `bls.Signature`.

### DIFF
--- a/scc/bls/bls_test.go
+++ b/scc/bls/bls_test.go
@@ -1,6 +1,7 @@
 package bls
 
 import (
+	"encoding/json"
 	"regexp"
 	"testing"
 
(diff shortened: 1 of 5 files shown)


scc/bls/bls_test.go
```diff
@@ -162,6 +162,18 @@ func TestPublicKey_String_IsHexEncoded(t *testing.T) {
 	require.Regexp(t, regexp, key.String())
 }
 
+func TestPublicKey_CanBeMarshalAndUnmarshal(t *testing.T) {
+	require := require.New(t)
+	key := PublicKey{}
+	data, err := key.MarshalJSON()
```
HerbertJordan, line 168: This test should use `json.Marshal` and `json.Unmarshal`. Also the one for the signature below.

scc/bls/bls_test.go
```diff
@@ -162,6 +162,18 @@ func TestPublicKey_String_IsHexEncoded(t *testing.T) {
...
+	require.NoError(err)
+
+	var key2 PublicKey
+	err = key2.UnmarshalJSON(data)
+	require.NoError(err)
+	require.Equal(key, key2)
+}
+
```
HerbertJordan, line 176: There should be a test like this added (also for the signature) to verify that the encoding is indeed using the desired format:
```
func TestPublicKey_JsonEncodingUsesHexFormat(t *testing.T) {
	regexp := regexp.MustCompile(`^"0x[0-9a-f]{96}"$`)
	key := NewPrivateKey().PublicKey()
	data, err := json.Marshal(key)
	require.NoError(t, err)
	require.Regexp(t, regexp, string(data))
}
```

HerbertJordan (changes requested):

HerbertJordan (approved):

PR #61, opened 2025-02-25
Fix issue of error message not printed when running sonictool as validator with a rpc database. 

LuisPH3:
When attempting to run node as validator after having ran once as RPC, an error has to be printed. 
The function here modified had a bug, overwriting pointers captured to do cleanups after the execution. 
This change removes the overwriting of named returns in the return statement.

### DIFF
--- a/integration/assembly.go
+++ b/integration/assembly.go
@@ -69,14 +69,21 @@ func rawMakeEngine(gdb *gossip.Store, cdb *abft.Store, cfg Configs) (*abft.Lache
 	return engine, vecClock, blockProc, nil
 }
 
-func makeEngine(chaindataDir string, cfg Configs) (engine *abft.Lachesis, vecClock *vecmt.Index,
-	gdb *gossip.Store, cdb *abft.Store, blockProc gossip.BlockProc, dbsClose func() error, err error) {
+func makeEngine(chaindataDir string, cfg Configs) (
+	_ *abft.Lachesis,
+	_ *vecmt.Index,
+	_ *gossip.Store,
+	_ *abft.Store,
+	_ gossip.BlockProc,
+	_ func() error,
+	err error,
+) {
 	dbs, err := GetDbProducer(chaindataDir, cfg.DBs.RuntimeCache)
 	if err != nil {
 		return nil, nil, nil, nil, gossip.BlockProc{}, nil, err
 	}
 
-	gdb, cdb, err = getStores(dbs, cfg)
+	gdb, cdb, err := getStores(dbs, cfg)
 	if err != nil {
 		err = fmt.Errorf("failed to get stores: %w", err)
 		return nil, nil, nil, nil, gossip.BlockProc{}, nil, err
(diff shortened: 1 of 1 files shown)


integration/assembly.go
```diff
@@ -69,30 +69,29 @@ func rawMakeEngine(gdb *gossip.Store, cdb *abft.Store, cfg Configs) (*abft.Lache
...
+	gdb *gossip.Store,
+	cdb *abft.Store,
+	blockProc gossip.BlockProc,
+	dbsClose func() error,
+	err error,
+) {
 	dbs, err := GetDbProducer(chaindataDir, cfg.DBs.RuntimeCache)
 	if err != nil {
-		return nil, nil, nil, nil, gossip.BlockProc{}, nil, err
+		return
```
facuMH, line 83:
```suggestion
		return nil, nil, gdb, cdb,  gossip.BlockProc{}, nil, err
```
and similarly for the other returns should work as well, but making explicit that the those 2 variables are still needed.

LuisPH3: I do not agree with this solution. 
The reason is that you need to reason the returns on each of the return cases, and in each of them the valid solution is to return the actual contents of the variable at the line.  It is extra code, therefore prone to errors.

The real problem with this function that it returns 7 parameters including a lambda.

HerbertJordan: I would also prefer to avoid [naked returns](https://boldlygo.tech/archive/2024-08-02-naked-returns/). It makes it even harder to understand what is returned where.

LuisPH3: I have to insist here,

the original error only existed because of not using naked returns. Error management during deferred lambdas requires named returns, when using named returns, only naked returns should be used, to avoid un-intentional overwriting of the named variable. which is the original error, which passed review. 

Doing a normal return can very easily overseen, because the assignment to the captured variable is implicit. 
Under the scenario of named returns, only naked returns should be used. precisely to force reviewer to check the correctness of the variables.

HerbertJordan: "Named return values" and "naked returns" are two different things. You can use named return values also with regular returns. The values you are returning then are getting assigned to the named return values before processing the deferred calls.

LuisPH3: rewrote using _ as name for all not-error returns.

facuMH (dismissed):

integration/assembly.go
```diff
@@ -69,45 +69,44 @@ func rawMakeEngine(gdb *gossip.Store, cdb *abft.Store, cfg Configs) (*abft.Lache
...
 		}
 	}()
 
 	err = gdb.EvmStore().Open()
-	dbsClose = dbs.Close
+	dbsClose := dbs.Close
 	if err != nil {
 		err = fmt.Errorf("failed to open EvmStore: %v", err)
-		return nil, nil, nil, nil, gossip.BlockProc{}, dbsClose, err
+		return nil, nil, gdb, cdb, gossip.BlockProc{}, dbsClose, err
```
HerbertJordan, line 103: nit: If there was an error i think you do not want to return `gdb` and `cdb`, just nil.

HerbertJordan (dismissed):

integration/assembly.go
LuisPH3: yes. I just got too confused with the return assignment thing. It is no longer needed to return something ere.

HerbertJordan (approved):

PR #62, opened 2025-02-25
Update intrinsic gas computation and checks

LuisPH3:
This PR updates the intrinsic gas computation to cover the range of changes that affect sonic. 
- Any cost computation related to versions older than Istanbul is removed. 
- Added check eip-3860 (limit and meter init-code), introduced in Shanghai
- Added support for EIP-7702 transactions

Event check is updated to utilize the new version in a conservative manner, for correctness, but it should be upgraded to do a correct validation according to the current block height revision for robustness reasons.

### DIFF
--- a/eventcheck/basiccheck/basic_check.go
+++ b/eventcheck/basiccheck/basic_check.go
@@ -49,6 +49,9 @@ func validateTx(tx *types.Transaction) error {
 
 		// is eip-3860 (limit and meter init-code )
 		// Disable to get the lower intrinsic gas cost of both options
+		// FIXME: from allegro on, this flag should be enabled.
+		// It should have been enabled since shanghai, but it was not and changing
+		// it now would be a hard fork.
 		false,
 	)
 	if err != nil {
--- a/eventcheck/basiccheck/basic_check.go
+++ b/eventcheck/basiccheck/basic_check.go
@@ -41,7 +41,16 @@ func validateTx(tx *types.Transaction) error {
 		return ErrNegativeValue
 	}
 	// Ensure the transaction has more gas than the basic tx fee.
-	intrGas, err := evmcore.IntrinsicGas(tx.Data(), tx.AccessList(), tx.To() == nil)
+	intrGas, err := evmcore.IntrinsicGas(
+		tx.Data(),
+		tx.AccessList(),
+		tx.SetCodeAuthorizations(),
+		tx.To() == nil, // is contract creation
+
+		// is eip-3860 (limit and meter init-code )
+		// Disable to get the lower intrinsic gas cost of both options
+		false,
+	)
 	if err != nil {
 		return err
 	}
--- a/evmcore/state_transition.go
+++ b/evmcore/state_transition.go
@@ -1,62 +1,75 @@
-// Copyright 2015 The go-ethereum Authors
-// This file is part of the go-ethereum library.
-//
-// The go-ethereum library is free software: you can redistribute it and/or modify
-// it under the terms of the GNU Lesser General Public License as published by
-// the Free Software Foundation, either version 3 of the License, or
-// (at your option) any later version.
-//
-// The go-ethereum library is distributed in the hope that it will be useful,
-// but WITHOUT ANY WARRANTY; without even the implied warranty of
-// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
-// GNU Lesser General Public License for more details.
-//
-// You should have received a copy of the GNU Lesser General Public License
-// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
-
 package evmcore
 
 import (
+	"bytes"
 	"math"
 
 	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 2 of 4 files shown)


evmcore/state_transition_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"bytes"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestIntrinsicGas_IsCompatibleWithGethImplementation(t *testing.T) {
```
HerbertJordan, line 13: I agree that this test is very important to keep the check in the block processor aligned with the one in the event validation. 

However, another property we would have to make sure is that this version is 100% compatible with the previous version. One idea to do so is to move the old version into the test file and rename it, and use it as a another reference version in this test and in a potential fuzzer test. Wdyt?

HerbertJordan (commented):
Looks good, but we need to be extra careful with this code. So I added a few thoughts that concern me. Let me know what you think.

eventcheck/basiccheck/basic_check.go
```diff
@@ -41,7 +41,16 @@ func validateTx(tx *types.Transaction) error {
 		return ErrNegativeValue
 	}
 	// Ensure the transaction has more gas than the basic tx fee.
-	intrGas, err := evmcore.IntrinsicGas(tx.Data(), tx.AccessList(), tx.To() == nil)
+	intrGas, err := evmcore.IntrinsicGas(
```
HerbertJordan, line 44: To be extra conservative, we might do the following: have the old version of this function and a new version and switch based on the `Allegro` or another feature flag between the old and the new one.

We need to get access to this parameters anyway. Do not feel to be forced to make this work without those flags.

evmcore/state_transition_test.go
LuisPH3: This is guaranteed by review. 
The hardcoded version without eip-3860 is equivalent to the previously existing implementation.
This is transitively tested by comparing to geth using eip-3860:  false.

eventcheck/basiccheck/basic_check.go
LuisPH3: The new implementation of the function, using false for eip-3860, and with empty authList is equivalent to the previous code.

eventcheck/basiccheck/basic_check.go
```diff
@@ -41,7 +41,16 @@ func validateTx(tx *types.Transaction) error {
...
+		tx.Data(),
+		tx.AccessList(),
+		tx.SetCodeAuthorizations(),
```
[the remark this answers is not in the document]
LuisPH3: Yes it is.  
`Transaction` is a type elision over `TransactionData`. The interface of any transaction not being 7702 returns an empty list.
Only transactions of type SetCodeTx can have a 7702 `TransactionData`.
Events check allowed transaction types [here ](https://github.com/0xsoniclabs/sonic/blob/d5b74b03996bfd5b53d745ad17cb27dd6d6f25bb/eventcheck/epochcheck/epoch_check.go#L84)

Notice, this check is also to be updated in a future PR. In this check, we have access to the rules. 
Wdyt: can we merge them, or move the intrinsic gas check here ?

evmcore/state_transition.go
```diff
@@ -1,62 +1,75 @@
...
+			}
+			gas += lenWords * params.InitCodeWordGas
+		}
 	}
 	if accessList != nil {
 		gas += uint64(len(accessList)) * params.TxAccessListAddressGas
 		gas += uint64(accessList.StorageKeys()) * params.TxAccessListStorageKeyGas
 	}
+	if authList != nil {
+		gas += uint64(len(authList)) * params.CallNewAccountGas
```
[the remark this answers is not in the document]
LuisPH3: This is Geth code, I did not consider that. No overflow existed before for the access-list neither.

eventcheck/basiccheck/basic_check.go
LuisPH3: Instead of having two implementations, I will keep a test-reference implementation and check that the one production implementation remains compatible.

evmcore/state_transition.go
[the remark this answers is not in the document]
LuisPH3: I have been reading the code, access list has a cost of 2400 per entry and auth list 25k. 
The memory of any system will be full before this values overflow. 
Nevertheless none of the additions in the function are checked. this is odd

LuisPH3:
I begin to be concerned about the usefulness and correctness of this routines:
Validation called when a transaction is fetched from the local pool (therefore superfluous) but when received from the lachesis gossiping as well, where it is needed.
- I haven't found the place where the signature of the transaction embedded in an event is checked.

LuisPH3:
This pr is closed, cleanup of TX validation from lachesis events will happen here: https://github.com/0xsoniclabs/sonic-admin/issues/90

PR #67, opened 2025-02-26
Add support for fetching the latest certificates

HerbertJordan:
This PR extends the gossip DB interface to support fetching the latest committee and block certificates from the store. This is needed by the RPC service, as fetching for the `latest` certificate shall be supported.

Since the Ethereum DB interface does not support reverse-iteration through keys, a binary search is employed for finding the latest entries in the respective tables.

### DIFF
--- a/gossip/store_certificates.go
+++ b/gossip/store_certificates.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"iter"
+	"math"
 
 	"github.com/0xsoniclabs/sonic/scc"
 	"github.com/0xsoniclabs/sonic/scc/cert"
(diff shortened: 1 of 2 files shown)


gossip/store_certificates.go
```diff
@@ -120,17 +125,40 @@ func getCertificate[S cert.Statement](
 	return res, res.Deserialize(data)
 }
 
+// getLatestCertificate retrieves the latest certificate from the key/value store.
+// If no certificate is found, an error is returned.
+func getLatestCertificate[S cert.Statement](
+	table kvdb.Store,
+) (cert.Certificate[S], error) {
+	var res cert.Certificate[S]
```
facuMH, line 133: nit: res is not used until the deserialization, I would rather it was declared just above it, and return explicit nils in error cases.

facuMH (approved):

gossip/store_certificates.go
HerbertJordan: As discussed offline, this version is deemed more readable then the alternative where we would have to define zero values in each early exit branch.

PR #69, opened 2025-02-26
Accept missing SCC genesis in database during startup

HerbertJordan:
This change disables the SCC chain if there is no genesis certificate in the database.


LuisPH3 (approved):
The PR seems to avoid the error that prevented the client from booting.

PR #73, opened 2025-02-27
Transaction Scrambler: Add set code authentications

simonlechner:
The addition of authorizations that have been added with eip-7702, has introduces new dependencies between transactions that have to be considered in the transaction scrambler.

### DIFF
--- a/gossip/tx_scrambler.go
+++ b/gossip/tx_scrambler.go
@@ -3,10 +3,12 @@ package gossip
 import (
 	"bytes"
 	"cmp"
-	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/core/types"
 	"math/big"
 	"slices"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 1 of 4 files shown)


gossip/tx_scrambler.go
```diff
@@ -166,6 +147,124 @@ func analyseEntryList(entries []ScramblerEntry) ([]ScramblerEntry, [32]byte, boo
...
+	// if both nonce and gas prices are equal, sort by hash
+	// note: at this point, hashes can never be same - duplicates are removed
+	return a.Hash().Cmp(b.Hash())
+}
+
+// scrambleTransactions scrambles transactions by comparing its XORed hashes with salt
+func scrambleTransactions(list []ScramblerEntry, salt [32]byte) {
+	var aX, bX [32]byte
+	slices.SortFunc(list, func(a, b ScramblerEntry) int {
+		aX = xorBytes32(a.Hash(), salt)
```
jnixable, line 262: does it make sense to replace XOR with something stronger ?

simonlechner:
This version of the scrambler will not be updated.

PR #78, opened 2025-02-28
Add certificate json format tests

facuMH:
This PR solves https://github.com/0xsoniclabs/sonic-admin/issues/91 by adding a test for block and committee certificates that checks the expected format is used when converting either to JSON.


ethapi/block_cert_json_test.go
```diff
@@ -90,3 +92,93 @@ func TestBlockCertificateToJson(t *testing.T) {
...
+}
+
+func TestBlockCertificate_ContainsExpectedValues_EmptyCertificate(t *testing.T) {
+	require := require.New(t)
+	emptyCert := makeEmptyBlockCert()
+	data, err := json.Marshal(toJsonBlockCertificate(emptyCert))
+	require.NoError(err)
+	require.Contains(string(data), `"chainId":0`)
+	require.Contains(string(data), `"number":0`)
+	require.Contains(string(data), `"hash":"0x`)
```
HerbertJordan, line 137: nit: this is a bit misleading ... it actually contains a `"hash":"0x0000000000000000000000000000000000000000000000000000000000000000"`, but the test, at a first glance, makes it look like a `"hash":"0x"` is expected -- the missing trailing `"` is missing, which is very subtile. The same for the state root and signature below.

A readable alternative may be something like
```suggestion
	require.Contains(string(data), `"hash":"0x` + strings.Repeat("00", 32) + `"`)
```

ethapi/block_cert_json_test.go
```diff
@@ -90,3 +92,93 @@ func TestBlockCertificateToJson(t *testing.T) {
...
+	require.Contains(string(data), fmt.Sprintf(`"signers":%v`, string(signers)))
+	require.Contains(string(data), fmt.Sprintf(`"signature":"%v"`, agg.Signature()))
+}
+
+func makeEmptyBlockCert() cert.BlockCertificate {
+	return cert.NewCertificateWithSignature(
+		cert.NewBlockStatement(0, 0, common.Hash{}, common.Hash{}),
+		cert.NewAggregatedSignature[cert.BlockStatement](
+			cert.BitSet[scc.MemberId]{}, bls.Signature{}),
+	)
```
HerbertJordan, line 166: This should be the same as
```suggestion
	return cert.BlockCertificate{}
```
and then it is cleaner to just inline it.

ethapi/block_cert_json_test.go
```diff
@@ -90,3 +92,93 @@ func TestBlockCertificateToJson(t *testing.T) {
...
+
+func TestBlockCertificate_JsonEncodingMatchesExpectedFormat(t *testing.T) {
+	tests := map[string]cert.BlockCertificate{
+		"empty":     makeEmptyBlockCert(),
+		"non-empty": makeTestBlockCert(t),
+	}
+
+	for name, cert := range tests {
+		t.Run(name, func(t *testing.T) {
+			validateBlockCertJsonFormat(t, cert)
```
HerbertJordan, line 104: style consideration: since this function is only used once, it can be inlined here without risking to much readability issues ...

ethapi/block_cert_json_test.go
```diff
@@ -90,3 +92,93 @@ func TestBlockCertificateToJson(t *testing.T) {
...
+		t.Run(name, func(t *testing.T) {
+			require := require.New(t)
+			data, err := json.Marshal(toJsonBlockCertificate(cert))
+			require.NoError(err)
+			require.Regexp(regexp.MustCompile(regex), string(data))
+		})
+	}
+}
+
+func TestBlockCertificate_ContainsExpectedValues_EmptyCertificate(t *testing.T) {
```
HerbertJordan, line 130:
```suggestion
func TestBlockCertificate_EmptyCertificate_ContainsExpectedValues(t *testing.T) {
```

ethapi/block_cert_json_test.go
```diff
@@ -90,3 +92,93 @@ func TestBlockCertificateToJson(t *testing.T) {
...
+	require.NoError(err)
+	require.Contains(string(data), `"chainId":0`)
+	require.Contains(string(data), `"number":0`)
+	require.Contains(string(data), `"hash":"0x`)
+	require.Contains(string(data), `"stateRoot":"0x`)
+	require.Contains(string(data), `"signers":null`)
+	require.Contains(string(data), `"signature":"0x`)
+}
+
+func TestBlockCertificate_ContainsExpectedValues_NonEmptyCertificate(t *testing.T) {
```
HerbertJordan, line 143:
```suggestion
func TestBlockCertificate_NonEmptyCertificate_ContainsExpectedValues(t *testing.T) {
```

ethapi/committee_cert_json_test.go
HerbertJordan, line 1: The same feedback as for the block-certificate tests is also applicable on this file.

HerbertJordan (dismissed):
Looks good, just a few style comments.

HerbertJordan (approved):

PR #79, opened 2025-02-28
use json null instead of "null" in certificate encoding

facuMH:
This PR changes jsonhex.Byte encoding so that `nil` byte slices are encoded as the JSON null value instead of the string "null".

### DIFF
--- a/utils/jsonhex/types.go
+++ b/utils/jsonhex/types.go
@@ -49,9 +49,6 @@ func (h Bytes) String() string {
 	if h == nil {
 		return `null`
 	}
-	if len(h) == 0 {
-		return `"0x"`
-	}
 	return fmt.Sprintf(`"0x%x"`, []byte(h))
 }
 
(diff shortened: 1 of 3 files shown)


utils/jsonhex/types.go
```diff
@@ -23,7 +23,7 @@ func (h *Bytes) UnmarshalJSON(data []byte) error {
 	if err != nil {
 		return err
 	}
-	if s == "null"{
+	if s == `` {
```
HerbertJordan, line 26: I think the empty string should be treated like a parsing error. After all, there is no `Bytes` instance that could be marshaled into an empty string.

HerbertJordan (approved):

utils/jsonhex/types.go
facuMH: the problem is that the JSON encoding of "null" is unmarshaled to the empty string. This is why I had to make this change

facuMH: using `json.RawMessage` we can see the null instead of losing it. For unmarshaling of other types, using `json.RawMessage` and then `string.Trim(\`\"\`)` would allow us to accept data values with and without `""`.

PR #81, opened 2025-03-04
Implement EIP-7702 support in rpc calls.

LuisPH3:
This PR extends RPC transaction serialization to enable clients to query the transaction data after the transaction has included in the block. 

The PR introduces the changes in the serialization code, plus a json serialization test in the API code and an integration test for an End-to-end SetCode transactions lifetime in the client test.

### DIFF
--- a/ethapi/api_test.go
+++ b/ethapi/api_test.go
@@ -2,6 +2,8 @@ package ethapi
 
 import (
 	"context"
+	"crypto/ecdsa"
+	"encoding/json"
 	"fmt"
 	"math/big"
 	reflect "reflect"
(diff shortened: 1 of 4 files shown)


ethapi/api_test.go
```diff
@@ -399,3 +401,235 @@ func setExpectedStateCalls(mockState *state.MockStateDB) {
...
+			types.NewTx(tx),
+			types.NewPragueSigner(chainId),
+			from)
+		require.NoError(t, err)
+		return res
+	default:
+		t.Error("unsupported transaction type ", reflect.TypeOf(payload))
+		return nil
+	}
+}
```
facuMH, line 550: cosmtic suggestion, wdyt ? 
```suggestion
var signer types.Signer
	var txTyped *types.Transaction
	switch tx := payload.(type) {
	case *types.LegacyTx:
		txTyped = types.NewTx(tx)
		signer = types.NewEIP155Signer(chainId)
	case *types.AccessListTx:
		txTyped = types.NewTx(tx)
		signer = types.NewEIP2930Signer(chainId)
	case *types.DynamicFeeTx:
		txTyped = types.NewTx(tx)
		signer = types.NewLondonSigner(chainId)
	case *types.BlobTx:
		txTyped = types.NewTx(tx)
		signer = types.NewCancunSigner(chainId)
	case *types.SetCodeTx:
		txTyped = types.NewTx(tx)
		signer = types.NewPragueSigner(chainId)
	default:
		t.Error("unsupported transaction type ", reflect.TypeOf(payload))
		return nil
	}
	res, err := types.SignTx(
		txTyped,
		signer,
		from)
	require.NoError(t, err)
	return res
}
```

tests/gossip_store_test.go
```diff
@@ -0,0 +1,183 @@
...
+	// was executed and check if the transaction is present in the block and the
+	// values match, by comparing the hashes.
+
+	net, err := StartIntegrationTestNet(t.TempDir())
+	if err != nil {
+		t.Fatalf("Failed to start the fake network: %v", err)
+	}
+	defer net.Stop()
+
+	client, err := net.GetClient()
```
facuMH, line 29: is there a specific reason why this client is not closed ? `defer` or otherwise ?

LuisPH3: it would be nice to close it.  yes.

ethapi/api_test.go
LuisPH3: Here you can see the influence of functional programming. I developed a resistance to use mutable variables, specially when they are default initialized.
On the other hand, your code is smaller and maybe cleaner (the possibility of the nil txTyped after the switch still freaks me out)

facuMH: right. I did not consider the nil case the `default` case does an early return. You could also add a `require.NotEmpty(txType)` for safety

LuisPH3: The code was simplified quite a lot. The latest available signer is capable of signing all the previous transaction types (with the exception of very old ones before eip-155, which do not support)

LuisPH3: https://github.com/0xsoniclabs/sonic/pull/81/commits/e6f79b84bbeaeb796efbdf5df14001111e5c2948

facuMH (dismissed):

tests/gossip_store_test.go
HerbertJordan, line 1: The "gossip store" is retaining everything, from blocks, over transactions, receipts, and logs -- many of this is already covered in other tests. How about renaming this to something like "transaction_store_test.go"?

HerbertJordan (dismissed):
Very nice, thanks!

tests/gossip_store_test.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/81/commits/4e9df76eb2c784fc65dcc4a637a290c39fa6d12e

HerbertJordan (approved):

facuMH (approved):

PR #84, opened 2025-03-05
Introduce integration test network sessions to parallelize integration test subtests.

LuisPH3:
This PR introduces a proxy account, which can be used to issue transactions in the same manner as the `IntegrationTestNet` object did before. 

Each subtest has a sequence of events, when such events use a disjoint set of accounts from the network, they can be executed concurrently with other workload. This PR introduces a `IntegrationTestNetSession` interface, with the already existing interface to execute transactions and retrieve information from a running network, but who's transactions are backed by a newly created account. 

The existing tests remain compatible, as the `IntegrationTestNet` already implements the interface.

- Set_code_tx_test reduces execution time from 180s to 50s.
- selfdestruct test reduces execution time from 40s to 25s. 

Two challenges arised during this development:
- balance, all transactions are backed by some account. Existing infrastructure uses the validator account, which is provided with a very large balance. The current changes create a new account to pay for transactions with a transfer from that validator account. Unfortunately, uint64 limits are too close to real costs and tests would fail.  This required the migration to big.Int to generate very large accounts to back up sessions. 
- defer conflicts with parallel, as described here: https://brandur.org/fragments/go-prefer-t-cleanup-with-parallel-subtests

**Future work**: I believe that this technique could be scaled to be used by every integration test in their top level, being all of them spawns from a singleton integration test network. Such an implementation would run in parallel with the memory footprint of a single network, tests would behave identically when ran sequentially. But the newly introduced parallelism could lead to clashes of addresses used by tests. 

**Review**: The PR is organized by commits, It may be easier to review the changes when reading one commit at the time.

### DIFF
--- a/tests/integration_test_net.go
+++ b/tests/integration_test_net.go
@@ -5,11 +5,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math"
 	"math/big"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strconv"
+	"sync"
 	"syscall"
 	"testing"
 	"time"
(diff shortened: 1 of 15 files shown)


tests/integration_test_net_test.go
```diff
@@ -39,7 +40,7 @@ func TestIntegrationTestNet_CanFetchInformationFromTheNetwork(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Failed to start the fake network: %v", err)
 	}
-	defer net.Stop()
+	t.Cleanup(func() { net.Stop() })
```
HerbertJordan, line 43: idea: since the life-time of a integration test net needs to be coupled with the runtime of a test, not the test function, we could pass the `t *testing.T` to the `StartIntegrationTestNet()` function and register the cleanup inside that function. This will remove all those explicit registration calls. Wdyt?

tests/integration_test_net.go
```diff
@@ -31,9 +34,41 @@ import (
...
+	// GetTransactOptions provides transaction options to be used to send a transaction
+	// from the given account.
+	GetTransactOptions(account *Account) (*bind.TransactOpts, error)
+	// Apply sends a transaction to the network using the session account.
+	// and waits for the transaction to be processed. The resulting receipt is returned.
+	Apply(issue func(*bind.TransactOpts) (*types.Transaction, error)) (*types.Receipt, error)
+
+	// GetAccount returns the default account of the session. This account is used
+	// to sign transactions and pay for gas when using the apply method.
+	GetAccount() *Account
```
HerbertJordan, line 66: nit: from what I see in tests, in most contexts, `net.GetAccount()` does not provide a lot of context to identify what account is meant. How about `GetTreasury()` or `GetSessionSponsor()`?

tests/transaction_order_test.go
```diff
@@ -141,10 +141,17 @@ func TestTransactionOrder(t *testing.T) {
...
 	account := NewAccount()
-	receipt, err := net.EndowAccount(account.Address(), math.MaxInt64)
+
+	// endowments are backed up by the test net account, and therefore are
+	// bound by the (very large but not infinite) balance of the test net account.
+	receipt, err := net.EndowAccount(account.Address(),
+		new(big.Int).Mul(
+			big.NewInt(math.MaxInt64),
+			big.NewInt(1e6),
+		))
```
HerbertJordan, line 154: The documentation of this function specifies that the resulting account has a balance of `math.MaxInt64`, but this modification makes it a million times more ... should the documentation be updated or the value?

HerbertJordan (approved):
Thanks for this great improvement in development convenience!

I added a few suggestions that might be included here or in follow-up work - please have a look.

tests/integration_test_net_test.go
LuisPH3: this is a pretty good idea

tests/transaction_order_test.go
LuisPH3: I don't think the comment should say how much balance is it there, 
These numbers are astronomical and impossible to reason about. I think that backing balance could be changed anytime as the tests scale.

tests/integration_test_net_test.go
LuisPH3: here: https://github.com/0xsoniclabs/sonic/pull/86

LuisPH3:
Current build may fail until https://github.com/0xsoniclabs/sonic/pull/86 is merged

tests/integration_test_net.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/84/commits/17f1babba8bd0ffc4fadb435e4947d5c99f7bded
GetSessionSponsor

tests/transaction_order_test.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/84/commits/becaa19d9033daa434ba9dd047ab8bc0770e8b0a

HerbertJordan (approved):

PR #86, opened 2025-03-06
Change Integration test net constructors to improve test cleanup lifetimes. 

LuisPH3:
The issue when introducing parallel sub-tests using integration test networks is that cleanup would happen before sub-tests terminate, closing the network and crashing the tests. 
This change manages automatically cleanup registration in the context of t. 

- temp directory for client storage is generated from within the constructor.
- Net is automatically registered for cleanup in the context of t.
- Adding t as argument forces this code to be test only.


HerbertJordan (dismissed):
Nice!

facuMH (approved):

PR #87, opened 2025-03-06
Reduce block time in integration tests

HerbertJordan:
This PR switches the genesis default for integration tests from a `fakenet` genesis to a test-package specific JSON genesis.

The `fakenet` genesis is a general feature facilitating the low-effort setup of a test net of multiple nodes for evaluation purposes. It is, for instance, used by Norma for setting up and running scenarios. Due to its scope, it is set up with a default event-emitter delay of 600ms, causing a block time of ~1 second.

For testing the JSON genesis feature, which is used for starting long-running networks like the Sonic main-net or public test nets, the integration test infrastructure also supports a mode to start a network using JSON file based genesis. This, so far, was only used for very few tests.

With this PR, the JSON genesis mode is made the default, and the emitter time for all tests is configured to be a single millisecond. As a result, blocks are produced much faster in tests, reducing the test time.

For instance, the time for the `TestTransactionOrder` test is reduced from 31 to 13 seconds on my development machine.

### DIFF
--- a/cmd/sonictool/app/app_test.go
+++ b/cmd/sonictool/app/app_test.go
@@ -122,7 +122,7 @@ func TestSonicTool_account_ExecutesWithoutErrors(t *testing.T) {
 func TestSonicTool_genesis_ExecutesWithoutErrors(t *testing.T) {
 
 	// Create a history by running some transactions
-	net := tests.StartIntegrationTestNet(t)
+	net := tests.StartIntegrationTestNetWithFakeGenesis(t)
 	generateNBlocks(t, net, 2)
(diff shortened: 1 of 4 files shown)


LuisPH3 (approved):
The changes migrate every integration test to use a json-initialized database.
The constructor name is counter intuitive. because no json file is provided, it initializes a fake net using a sligtly different initialization procedure. 

StartIntegrationTestNetWithJsonGenesis (default net)  will reduce event emmiter interval (to 1 ms? )  to improve the thoughtput of the test transactions.

PR #92, opened 2025-03-07
Add feature set enumeration for configuration of fake-nets

LuisPH3:
This PR introduces an enumeration to groups feature sets under one single release name. 
This will be used to initialize fake-nets after a given release. 
The intention of this PR is to discuss naming and documentation.


opera/rules_test.go
```diff
@@ -0,0 +1,67 @@
...
+	tests := map[string]FeatureSet{
+		"sonic":   SonicFeatures,
+		"allegro": AllegroFeatures,
+		"unknown": FeatureSet(math.MaxInt),
+	}
+
+	for expected, fs := range tests {
+		if fs.String() != expected {
+			t.Errorf("Expected %s, got %s", expected, fs.String())
+		}
```
facuMH, line 21: nit:
```suggestion
		require.Equals(expected, fs.String())
```

opera/rules_test.go
```diff
@@ -0,0 +1,67 @@
...
+		AllegroFeatures: {
+			expectedUpgrades: Upgrades{
+				Berlin:  true,
+				London:  true,
+				Llr:     false,
+				Sonic:   true,
+				Allegro: true,
+			},
+		},
+		FeatureSet(math.MaxInt): {
```
facuMH, line 49: nit: error case could be it's own test

opera/rules.go
```diff
@@ -433,3 +434,59 @@ func (r Rules) String() string {
...
+
+func GetAllegroUpgrades() Upgrades {
+	return Upgrades{
+		Berlin:  true,
+		London:  true,
+		Llr:     false,
+		Sonic:   true,
+		Allegro: true,
+	}
+}
```
facuMH, line 492:
```suggestion
func GetAllegroUpgrades() Upgrades {
    upgrades := GetSonicUpgrades()
    upgrades.Allegro := true
	return upgrades
}
```
wdyt?  makes more explicit the change from sonic to Allegro, less error prone to set a "previous" flag wrong.

opera/rules_test.go
```diff
@@ -0,0 +1,67 @@
...
+		},
+		AllegroFeatures: {
+			expectedUpgrades: Upgrades{
+				Berlin:  true,
+				London:  true,
+				Llr:     false,
+				Sonic:   true,
+				Allegro: true,
+			},
+		},
```
facuMH, line 48:
```suggestion
		SonicFeatures: {
			expectedUpgrades: GetSonicUpgrades(),
		},
		AllegroFeatures: {
			expectedUpgrades: GetAllegroUpgrades(),
		},
```
do you think doing it like this would hurt readability ?

opera/rules.go
```diff
@@ -433,3 +434,59 @@ func (r Rules) String() string {
...
+	return Upgrades{
+		Berlin:  true,
+		London:  true,
+		Llr:     false,
+		Sonic:   true,
+		Allegro: false,
+	}
+}
+
+func GetAllegroUpgrades() Upgrades {
```
facuMH, line 484: should there be a test checking the flags are set correctly ?

opera/rules.go
```diff
@@ -433,3 +434,59 @@ func (r Rules) String() string {
 	b, _ := json.Marshal(&r)
 	return string(b)
 }
+
+// FeatureSet is an enumeration of different releases, each one enabling a
+// different set of features. These are an abstraction that allows to reason
+// about the different releases instead of isolated upgrades.
+// Feature sets are exclusive, and a (fake-)net cannot be configured with
+// more than one feature set at the time.
+type FeatureSet int
```
HerbertJordan, line 443: nit: this might deserve its own file in the same package. Wdyt?

opera/rules.go
```diff
@@ -433,3 +434,59 @@ func (r Rules) String() string {
...
+	case SonicFeatures:
+		return "sonic"
+	case AllegroFeatures:
+		return "allegro"
+	default:
+		return "unknown"
+	}
+}
+
+func GetSonicUpgrades() Upgrades {
```
HerbertJordan, line 474: very small nit: maybe we make this private to make something like `SonicFeatures.ToUpgrades()` the canonical way to retrieve the upgrade flags? Or maybe even inline these functions?

opera/rules.go
LuisPH3: it its the previous test.

opera/rules_test.go
LuisPH3: this does not test correctness of the function

opera/rules.go
LuisPH3: maybe, how should it be called?  I am resistant to have both opera and sonic packages. 
naming it.. featureSet, makes it of lower importance than "opera" or "sonic".
I spiral with this problem for a few days now.

opera/rules.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/92/commits/ac1ae6eb4e86e450e4bdf61363a1a2fccedc2a22

opera/rules_test.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/92/commits/ac1ae6eb4e86e450e4bdf61363a1a2fccedc2a22

opera/rules_test.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/92/commits/ac1ae6eb4e86e450e4bdf61363a1a2fccedc2a22

opera/rules.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/92/commits/ac1ae6eb4e86e450e4bdf61363a1a2fccedc2a22

opera/rules.go
HerbertJordan: Just a file, not a package.

HerbertJordan (approved):

opera/rules.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/92/commits/89e159d9f6f92ff5c027fd2bb58c1323a05b56f8

HerbertJordan (approved):

opera/feature_set.go
```diff
@@ -0,0 +1,51 @@
...
+			London:  true,
+			Llr:     false,
+			Sonic:   true,
+			Allegro: true,
+		}
+	default:
+		return res, fmt.Errorf("unknown feature set: %v", fs)
+	}
+	return res, nil
+}
```
facuMH, line 40:
```suggestion
func (fs FeatureSet) ToUpgrades() (Upgrades, error) {
	res := Upgrades{
				Berlin:  true,
				London:  true,
				Llr:     false,
			}
	switch fs {
	case SonicFeatures:
		res.Sonic := true
	case AllegroFeatures:
		res.Sonic = true
		res.Allegro = true
		}
	default:
		return Upgrade{}, fmt.Errorf("unknown feature set: %v", fs)
	}
	return res, nil
}
```

Considering that for Sonic the "old" 3 will always be the same. Wdyt ?

facuMH (approved):
Looks good, just a small suggestion.

opera/feature_set.go
LuisPH3: I would rather keep it declarative.  This change obscures the contents. 
If the code grows too much, then we can apply some refactoring.

PR #97, opened 2025-03-10
feat: datasemaphore tests

aksonic:
This PR adds tests to the `datasemaphore` utils package and adds comments explaining how `datasemaphore` works in detail - as it is a bit complicated.

The `Acquire` function has been rewritten because the previous implementation made it possible for `Acquire` to hang because the `sync.Cond` condition variable associated with the semaphore would never have `Broadcast()` called - while `Broadcast()` would potentially be triggered through the `Release` function called on the same semaphore instance - this is not guaranteed. A timer with a broadcast has now been added to the `Acquire` function and is called after the timeout to acquire the resources has elapsed to wake up **_all_** goroutines that may be waiting for a signal to retry - _including the same process that started the acquisition attempt_ so that `Acquire` does not hang.

Verification:

image

### DIFF
--- a/utils/datasemaphore/datasemaphore.go
+++ b/utils/datasemaphore/datasemaphore.go
@@ -7,14 +7,21 @@ import (
 	"github.com/0xsoniclabs/consensus/inter/dag"
 )
 
+// DataSemaphore implements a resource control mechanism for limiting concurrent processing
+// of DAG elements based on two resource dimensions: number of events and total size.
+// It allows callers to acquire resources up to configured limits and blocks if limits would be exceeded.
 type DataSemaphore struct {
-	processing    dag.Metric
-	maxProcessing dag.Metric
-
-	mu   sync.Mutex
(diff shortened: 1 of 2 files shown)


psubotic (approved):

Tjofil (approved):

utils/datasemaphore/datasemaphore.go
```diff
@@ -7,14 +7,21 @@ import (
...
+// It allows callers to acquire resources up to configured limits and blocks if limits would be exceeded.
 type DataSemaphore struct {
-	processing    dag.Metric
-	maxProcessing dag.Metric
-
-	mu   sync.Mutex
-	cond *sync.Cond
-
-	warning func(received dag.Metric, processing dag.Metric, releasing dag.Metric)
+	// github.com/0xsoniclabs/consensus/inter/dag
```
Tjofil, line 14: Is this comment like a reminder for the data structure ?

PR #103, opened 2025-03-11
Refactor logs initialization.

LuisPH3:
This PR introduces some changes in the logging initialization of the Sonic project.
- Introduces logging arguments into the sonictool. Log format, log level, per module verbosity, etc.
- Introduces an environment variable `SONIC_VERBOSITY` to tune log verbosity of sonictool, sonicd, and integration tests without recompiling.  
- Defaults to silent logs when running integration tests.

Example of use: 
`SONIC_VERBOSITY=3 go test -timeout 10m -run ^TestTransactionStore_CanTransactionsBeRetrievedFromBlocksAfterRestart$ github.com/0xsoniclabs/sonic/tests -v -count 1
`

This PR has been written with the "single point of responsibility" principle in mind, therefore the feature has been implemented in the command line arguments instead of duplicating the feature in the integration test tooling. 
For this reason, the complete debug namespace is utilized in sonictool as well, instead of cherry-picking the log level flag.

### DIFF
--- a/cmd/sonictool/app/main.go
+++ b/cmd/sonictool/app/main.go
@@ -5,6 +5,7 @@ import (
 	"sort"
 
 	"github.com/0xsoniclabs/sonic/config/flags"
+	"github.com/0xsoniclabs/sonic/debug"
 	"github.com/0xsoniclabs/sonic/version"
 	"gopkg.in/urfave/cli.v1"
 )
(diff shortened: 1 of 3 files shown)


facuMH (approved):
nice and simple way of changing integration test client log level !!

HerbertJordan (approved):
This PR unifies the handling of verbosity levels in various commands and adds the additional option to configure the verbosity using an environment variable.

Nice and clean solution.

PR #104, opened 2025-03-11
Change NoBaseFee vm config parameter for replaying transaction

jenikd:
This PR sets NoBaseFee parameter for vm config to true, to be able to replay transactions with gas price set to 0. This is usual for internal transactions.


thaarok (approved):
Thanks for this fix!

HerbertJordan (approved):

PR #105, opened 2025-03-11
Use members instead of committee for genesis json

facuMH:
Since the `GenesisJson` type used a `*scc.Committee` to store the genesis committee, `json.marshall` could not access the committee members, because this a private field, and so when exporting the genesis into a json file,  the committee was lost.

This PR adds 2 main changes, to fix this:
- First, two extra checks when decoding a json genesis, that the genesis committee has members, and these constitute a valid committee.
- Second, changes the `GenesisJson` member from `*scc.Committee` to `[]scc.Member`


facuMH:
replaced by https://github.com/0xsoniclabs/sonic/pull/106

PR #106, opened 2025-03-11
Add json encode/decode methods to `scc.Committee`

facuMH:
The `GenesisJson` type stores the genesis committee as a `*scc.Committee`, which contains a private field. As a result, when encoding with `json.Marshal`, reflection couldn't access the committee members, causing them to be lost when exporting the genesis to a JSON file.

This PR introduces two key changes to address this issue:
- Adds two extra validation checks when decoding a JSON genesis: ensuring that the genesis committee has members and that they form a valid committee.
- Implements `MarshalJSON` and `UnmarshalJSON` methods for `scc.Committee` to properly handle JSON serialization.


HerbertJordan (approved):
Thanks for fixing this. But heads-up: the committee structure in the JSON genesis and in the  `scc` package will diverge in the future.

LuisPH3 (approved):

PR #108, opened 2025-03-12
Add generate test data iterator

LuisPH3:
This PR adds testing tooling to facilitate the construction of inputs for tests. 

The use case will be to generate transaction types for validation or evaluation of execution based on different parametrization. 

The reason why the function requires a constructor is to guarantee stateless behavior during iteration. An alternative would have been to provide an `initial value` "a la fold". Nevertheless, to avoid deep-copy, the constructor function was introduced. 

Iterators are used instead of callbacks or slices to facilitate integration in loops and improve debuging of the tests using this code.

### DIFF
--- a/tests/generate_test_data_test.go
+++ b/tests/generate_test_data_test.go
@@ -1,0 +1,129 @@
+package tests
+
+import (
+	"iter"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// generateTestDataBasedOnModificationCombinations generates all possible versions of a
+// given type based on the combinations of modifications.
+// The iterator works around a function modify(T, []Piece) T, which shall modify
+// an newly constructed instance of T with the provided piece-modifiers.
+//
+// Arguments:
+//   - constructor: a function that constructs a new instance of T, for each version
+//     to be based on an unmodified instance.
+//   - pieces: a list of lists of pieces, where each list of pieces represents a
+//     domain of possible modifications.
+//   - modify: a function that modifies an instance of T with the provided pieces.
+//
+// Returns:
+// - an iterator that yields all possible versions of T based on the combinations
+func generateTestDataBasedOnModificationCombinations[T any, Piece any](
+	constructor func() T,
+	pieces [][]Piece,
+	modify func(tx T, modifier []Piece) T,
+) iter.Seq[T] {
+
+	return func(yield func(data T) bool) {
+		_cartesianProductRecursion(nil, pieces,
+			func(pieces []Piece) bool {
+				v := constructor()
+				v = modify(v, pieces)
+				return yield(v)
+			})
+	}
+}
+
+func TestCartesianProduct_CountInstantiations(t *testing.T) {
+
+	count := func(_ int, modifier []int) int {
+		var count int
+		for _, v := range modifier {
+			count += v
+		}
+		return count
+	}
+
+	countInstances := func(pieces [][]int, modifier func(int, []int) int) int {
+		var count int
+		makeZero := func() int { return 0 }
+		for _ = range generateTestDataBasedOnModificationCombinations(makeZero, pieces, modifier) {
+			count++
+		}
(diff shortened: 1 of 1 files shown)


tests/generate_test_data_test.go
```diff
@@ -0,0 +1,129 @@
...
+//
+// Returns:
+// - an iterator that yields all possible versions of T based on the combinations
+func generateTestDataBasedOnModificationCombinations[T any, Piece any](
+	constructor func() T,
+	pieces [][]Piece,
+	modify func(tx T, modifier []Piece) T,
+) iter.Seq[T] {
+
+	return func(yield func(data T) bool) {
```
HerbertJordan, line 30: It might be simpler to separate the computation of a cartesian product and the test data generation. The main advantages might be:
- easier to test
- a stand-alone cartesian product function may have other uses as well
- you could specialize the test-data generation to only accept modifications and "hard-code" the application of the modification; this should improve its usability;

An example how this could look like is [here](https://github.com/0xsoniclabs/sonic/blob/2235ff5c94e8a5b109a9e3e037a7366a649e502c/tests/generate_test_data_test.go#L23-L47). Wdyt?

tests/generate_test_data_test.go
```diff
@@ -0,0 +1,129 @@
...
+	countInstances := func(pieces [][]int, modifier func(int, []int) int) int {
+		var count int
+		makeZero := func() int { return 0 }
+		for _ = range generateTestDataBasedOnModificationCombinations(makeZero, pieces, modifier) {
+			count++
+		}
+		return count
+	}
+
+	assert.Equal(t, 0, countInstances(nil, count))
```
HerbertJordan, line 59: nit: formally, the size of the cartesian product of an empty list of element lists is 1 - the empty sequence. I think it also makes sense in the context of modifications, that if there are no modifications, the unmodified element should be returned.

HerbertJordan (dismissed):
I think the code could be improved, but I am fine with it as it is if you prefer to move on instead.

tests/generate_test_data_test.go
LuisPH3: this makes sense.

LuisPH3: https://github.com/0xsoniclabs/sonic/pull/108/commits/4e2870fcfedd45a9e4a7fbc28c8c3544c234ca25

tests/generate_test_data_test.go
LuisPH3: I did not migrate the proposed changes because I rather keep the modify function parameter. The reason is that when using this tool, the pieces that contribute to construct a test case may contain data, for example name of the rule, and functions to modify some data. 

```go
	generateTestDataBasedOnModificationCombinations(
		func() TestCase {
			return TestCase{
			        name: "" // < acyumulate name here
                                transaction:  new Tx
                                expectedCost: 21000
                         }
		},
		[][]CostDefinition{
			{
				{
					name:         "no calldata",
					modification: withCallData(0, 0),
					cost:         0,
				},
				{
					name:         "ones calldata",
					modification: withCallData(10, 1),
					cost:         (9*4 + 1) * 4,
				},
			},
			{
				{
					name:         "existing account",
					modification: withTo(&existingAccountAddress),
				},
				{
					name:         "non existing account",
					modification: withTo(&nonExistingAccountAddress),
				},
                        },	
		},
                func(tc TestCase, pieces []CostDefinition) TestCase {
		   // accumulate test case here
		 })
```			

By internalizing the modify function into the function loop, this flexibility is lost.
The modify function could be implemented as a member of T, and requiring the matching interface in the function. This approach would not allow using this generator for external types.

LuisPH3:
> I think the code could be improved, but I am fine with it as it is if you prefer to move on instead.

I agree that It could be cleaner, but for the time being, I think I will leave as it is, reason discussed in comment.

HerbertJordan (approved):

PR #111, opened 2025-03-13
Add `BlockQueryI` interface and implementation

facuMH:
This PR adds the `BlockQueryI` interface and first implementation, `BlockQuery`, along with unit tests and an integration test. 
`BlockQuery` is meant to facilitate fetching the information of an account at a specific block.
For now the interface only requires one method `GetAddressInfo` which takes an address and a block number, and returns a `ProofQuery`, which contains a state root hash (to be used as proof), next to the balance and nonce of the given address at the specified block number.


facuMH:
This PR is dropped. There will be another approach to this functionality

PR #115, opened 2025-03-13
TxScrambler: Consider authorizations in transaction order

simonlechner:
After adding the examples of the original eip-7702 txScrambler update, tests were failing. This was due to the fact that the authorizations have not been considered in the ordering of transactions.

### DIFF
--- a/gossip/scrambler/scrambler.go
+++ b/gossip/scrambler/scrambler.go
@@ -242,6 +242,7 @@ func pickOptimal(
 ) transaction {
 	var best transaction
 	bestNumberOfTransactions := 0
+	var numPassingAuths int
 	for _, tx := range ready {
 		// Copy the state to avoid modifying the original state.
 		noncesCopy := nonces.copy()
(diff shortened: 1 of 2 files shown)


gossip/scrambler/scrambler.go
```diff
@@ -242,16 +242,35 @@ func pickOptimal(
...
 		noncesCopy := nonces.copy()
 		noncesCopy.apply(tx)
 		numTransactions := 1 + len(getTransactionOrder(
 			transactions, noncesCopy, pickOptimal,
 		))
-		if numTransactions > bestNumberOfTransactions {
+
+		// Find the number of passing authorizations.
+		// Multiple authorizations with the same sender are not taken into account,
+		// there is no real world use case for it.
```
HerbertJordan, line 256: Is this derived from the restriction of a transaction pool only to contain one authorization per sender account?

If so, I think this assumption does not hold and there might be cases like this in the real world. Although each individual pool would not allow multiple authorizations for a single account, different pool instances of different nodes may contain distinct authorizations of the same sender, and they may get together in a single block.

Also, the transaction pool rules are a suggestion, not something that is enforced. Any network participant way chose to use arbitrary modified rules.

gossip/scrambler/scrambler.go
```diff
@@ -242,16 +242,35 @@ func pickOptimal(
...
+		authPassingCount := 0
+		for _, a := range tx.auth {
+			if nonces[a.sender] == a.nonce {
```
HerbertJordan, line 259: I think this would miss a case where the authorization is send by a sender that also has transaction in the list and the nonce of the transaction's nonce is one less than the authorization. In this case the transaction may be executed first, enabling the authorization to pass.

If I have more time I can try to construct such a test case.

HerbertJordan (changes requested):
Thanks for adding those examples and the extension of the optimization criteria. 

I agree, we should also strive for maximizing the number of executable authorizations as a second criteria after the number of processed transactions. However, I think we can not make the assumption that there is only a single authorization for each sender. Can you improve on that?

gossip/scrambler/scrambler.go
simonlechner: This is just regarding multiple authorizations with the same sender in the same transaction. The nonce would have to be increased inside the loop if the same address is there multiple times. There is no real world use case for multiple authorizations with the same address in one transaction.

gossip/scrambler/scrambler.go
simonlechner: I think this case is evaluated in a different recursive call since there is one for each ready transaction.

simonlechner:
This version of the scrambler will not be developed further.

PR #117, opened 2025-03-14
Integration tests: Workaround for missing blocks while fetching receipts.

LuisPH3:
Integration tests seem to have a Flaky behavior, 
Running `make test` shows the following error:
```
09:39:00  --- FAIL: TestTransactionOrder (8.44s)
09:39:00      transaction_order_test.go:75: 
09:39:00          	Error Trace:	/home/jenkins/workspace/_Request_Tests_0xSonicLabs_PR-93/tests/transaction_order_test.go:75
09:39:00          	Error:      	Received unexpected error:
09:39:00          	            	failed to get transaction receipt: block 0x9 not found
09:39:00          	Test:       	TestTransactionOrder
```

It seems that there is a synchronization issue in the get receipt function:

```go
// GetTransactionReceipt returns the transaction receipt for the given transaction hash.
func (s *PublicTransactionPoolAPI) GetTransactionReceipt(ctx context.Context, hash common.Hash) (map[string]interface{}, error) {
	tx, blockNumber, index, err := s.b.GetTransaction(ctx, hash)
	if tx == nil || err != nil {
		return nil, err
	}
	header, err := s.b.HeaderByNumber(ctx, rpc.BlockNumber(blockNumber)) // retrieve header to get block hash
	if header == nil || err != nil {
		return nil, err
	}
	receipts, err := s.b.GetReceiptsByNumber(ctx, rpc.BlockNumber(blockNumber))
	if receipts == nil || err != nil {
		return nil, err
	}
	if receipts.Len() <= int(index) {
		return nil, nil
	}
	return s.formatTxReceipt(header, tx, index, receipts[index]), nil
}
```

`s.b.GetTransaction(ctx, hash)` returns a valid blockNumber, but it is not yet available to ` s.b.HeaderByNumber(`  

**Changes in this PR**
The behavior of the `GetTransactionReceipt` may not be wrong, "block not found" is, after all, similar to the `ethereum.NotFound` error returned by the client side if receipt is not found:
```go
func (ec *Client) TransactionReceipt(ctx context.Context, txHash common.Hash) (*types.Receipt, error) {
	var r *types.Receipt
	err := ec.c.CallContext(ctx, &r, "eth_getTransactionReceipt", txHash)
	if err == nil && r == nil {
		return nil, ethereum.NotFound
	}
	return r, err
}
```

This PR adds code to retry receipt query if block is not found.

### DIFF
--- a/tests/integration_test_net.go
+++ b/tests/integration_test_net.go
@@ -749,22 +749,33 @@ func (s *Session) GetReceipt(txHash common.Hash) (*types.Receipt, error) {
 
 	// Wait for the response with some exponential backoff.
 	const maxDelay = 100 * time.Millisecond
+	aggregatedErrors := make([]error, 0)
+	const maxErrRetry = 5
 	now := time.Now()
 	delay := time.Millisecond
 	for time.Since(now) < 100*time.Second {
 		receipt, err := client.TransactionReceipt(context.Background(), txHash)
-		if errors.Is(err, ethereum.NotFound) {
-			time.Sleep(delay)
-			delay = 2 * delay
-			if delay > maxDelay {
-				delay = maxDelay
+		if err == nil {
+			return receipt, nil
+		}
+
+		// for any other error than NotFound, retry a number of times
+		if !errors.Is(err, ethereum.NotFound) {
+			aggregatedErrors = append(aggregatedErrors, err)
+
+			if len(aggregatedErrors) >= maxErrRetry {
+				return nil, fmt.Errorf("failed to get transaction receipt: %w",
+					errors.Join(aggregatedErrors...))
 			}
-			continue
 		}
-		if err != nil {
-			return nil, fmt.Errorf("failed to get transaction receipt: %w", err)
+
+		// If the receipt returned an error, retry after a delay.
+		// Block is not available immediately after the transaction is submitted.
+		time.Sleep(delay)
+		delay = 2 * delay
+		if delay > maxDelay {
+			delay = maxDelay
 		}
-		return receipt, nil
 	}
 	return nil, fmt.Errorf("failed to get transaction receipt: timeout")
 }


HerbertJordan (approved):
It looks like the source of the problem is [here](https://github.com/0xsoniclabs/sonic/blob/89a9d9970cb9b74e5f0c70312e48794d897c8163/gossip/c_block_callbacks.go#L369-L373). Whenever a block is processed, its transactions are stored before the block itself is stored.

In the RPC looking for receipts, it is first trying to load the [transaction](https://github.com/0xsoniclabs/sonic/blob/89a9d9970cb9b74e5f0c70312e48794d897c8163/ethapi/api.go#L1720) and then the [block header](https://github.com/0xsoniclabs/sonic/blob/89a9d9970cb9b74e5f0c70312e48794d897c8163/ethapi/api.go#L1724). If by chance the transaction is already found in the store but the block not yet, the block header lookup fails with an error [here](https://github.com/0xsoniclabs/sonic/blob/89a9d9970cb9b74e5f0c70312e48794d897c8163/gossip/ethapi_backend.go#L89).

#### Some ways to fix that would not work
- ideally, transactions and block headers are stored in an atomic transaction; this way, no intermediate access can see a partial state; however, DB infrastructure for the underlying store does not support atomic operations over multiple write operations
- we could store the block header before the transaction: this would fix the lookup for the receipts here, but other RPC functions retrieving blocks depend on the Tx-first order

#### A fix that could work
It seems like the possibility of missing data is anticipated in most RPC methods; the way to indicate missing data is to return `nil` with a `nil` error. Thus, the bug seems to be [here](https://github.com/0xsoniclabs/sonic/blob/89a9d9970cb9b74e5f0c70312e48794d897c8163/gossip/ethapi_backend.go#L89), where instead of an error due to missing data `nil, nil` should be returned.

Most users of the `HeaderByNumber()` function are already handling the possibility of a nil value correctly, but we should carefully check that we got all of them. An example change covering this is here 6c77e2f26e78bfa5771444723057f5e90cd6fa6c

@LuisPH3 can you check whether this fix would eliminate the flakiness?

tests/integration_test_net.go
```diff
@@ -749,22 +749,33 @@ func (s *Session) GetReceipt(txHash common.Hash) (*types.Receipt, error) {
...
+		if err == nil {
+			return receipt, nil
+		}
+
+		// for any other error than NotFound, retry a number of times
+		if !errors.Is(err, ethereum.NotFound) {
+			aggregatedErrors = append(aggregatedErrors, err)
+
+			if len(aggregatedErrors) >= maxErrRetry {
+				return nil, fmt.Errorf("failed to get transaction receipt: %w",
```
facuMH, line 767: have you made fail case test to see how this aggregated error would look like ? I wonder if 5 errors of retries in a row would make sense. what do you think about changing the error message to `failed to get transaction receipt after %v retries:` ?

LuisPH3: yes, I will have a look If I can remove the cause for this issue, but if I cannot get it done, I will update this

LuisPH3:
Overseen by #124

PR #118, opened 2025-03-17
feat: migrate workers & add tests

aksonic:
This PR migrates the `workers` utils from the `consensus` repository into the `sonic` repository since they are only used here and adds testing for 100% coverage. Workers is a utility to manage a pool of concurrent goroutines - please see code comments.

Verification:

image


LuisPH3 (changes requested):
If the time allows, I would like to have an iteration on tests.

PR #119, opened 2025-03-17
feat: migrate wthreshold & add tests

aksonic:
This PR migrates the `wmedian` utils from the `consensus` repository into the `sonic` repository (and renames it to the more appropriate `wthreshold` package, along with updating all relevant references, since it's not really a median) since they are only used here and adds testing for 100% coverage.

`wthreshold` exposes the `FindThresholdValue` function which iterates through a slice of `WeightedValues`, accumulating total weight and returns the first `WeightedValue` with which the total weight exceeds a given threshold - please see code comments - this function is required for gossip functionality.

Verification:

image

### DIFF
--- a/utils/wthreshold/wthreshold.go
+++ b/utils/wthreshold/wthreshold.go
@@ -1,0 +1,24 @@
+package wthreshold
+
+import (
+	"github.com/0xsoniclabs/consensus/inter/pos"
+)
+
+type WeightedValue interface {
+	Weight() pos.Weight
+}
+
+// FindThresholdValue iterates through a slice of WeightedValues, accumulating total weight
+// and returns the first WeightedValue with which the total weight exceeds a given threshold.
+// If cumulative weight of all values is less than the threshold, panic.
+func FindThresholdValue(values []WeightedValue, threshold pos.Weight) WeightedValue {
+	// Calculate weighted threshold value
+	var totalWeight pos.Weight
+	for _, value := range values {
+		totalWeight += value.Weight()
+		if totalWeight >= threshold {
+			return value
+		}
+	}
+	panic("invalid threshold value")
+}
--- a/utils/wthreshold/wthreshold_test.go
+++ b/utils/wthreshold/wthreshold_test.go
@@ -1,0 +1,114 @@
+package wthreshold
+
+import (
+	"testing"
+
+	"github.com/0xsoniclabs/consensus/inter/pos"
+)
+
+type testWeightedValue struct {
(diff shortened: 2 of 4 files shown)


utils/wthreshold/wthreshold.go
LuisPH3, line 1: is this file migrated?

utils/wthreshold/wthreshold_test.go
```diff
@@ -0,0 +1,114 @@
...
+			},
+			threshold: 1,
+			want:      "test16",
+			panic:     false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.panic {
```
LuisPH3, line 99: nit: for readability, we have a "tradition" of splitting the tests, successful cases and rejection cases. 
This removes the if inside the test case and usually makes code more readable.

utils/wthreshold/wthreshold.go
```diff
@@ -0,0 +1,24 @@
...
+func FindThresholdValue(values []WeightedValue, threshold pos.Weight) WeightedValue {
+	// Calculate weighted threshold value
+	var totalWeight pos.Weight
+	for _, value := range values {
+		totalWeight += value.Weight()
+		if totalWeight >= threshold {
+			return value
+		}
+	}
+	panic("invalid threshold value")
```
LuisPH3, line 23: Unfortunately. We can not allow a panic here. I know that this routine may have existed as described here before, but a panic in runtime has the following properties:
- It cannot be assets if the panic cannot be triggered during runtime, because is based on runtime values. 
- It may crash the program in critical regions, with unusual inputs/data sets.

For this reason, I strongly encourage you to investigate how difficult is to return an error if the threshold is not valid.

utils/wthreshold/wthreshold.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/0xsoniclabs/consensus/inter/pos"
+)
+
+type WeightedValue interface {
+	Weight() pos.Weight
+}
+
+// FindThresholdValue iterates through a slice of WeightedValues, accumulating total weight
+// and returns the first WeightedValue with which the total weight exceeds a given threshold.
+// If cumulative weight of all values is less than the threshold, panic.
```
LuisPH3, line 13: after removing panic, please rewrite doc

LuisPH3 (changes requested):
If times allows, I would like to remove this panic

PR #120, opened 2025-03-17
Add Light Client

facuMH:
This PR introduces the `LightClient`. Light clients store minimal state and operate without connecting to the peer-to-peer network. They don’t retain the chain’s head state or block history, relying instead on (untrusted) Archive servers for additional data, which can be verified using the current head block’s certificate.

This PR adds:
- The `LightClient` with a factory that takes a `Config` (pair of certificates provider and an initial committee) and a method to sync into the network.
- Unit tests for the `LightClient`

### DIFF
--- a/scc/light_client/light_client.go
+++ b/scc/light_client/light_client.go
@@ -1,0 +1,56 @@
+package light_client
+
+import (
+	"fmt"
+	"slices"
+
+	"github.com/0xsoniclabs/sonic/scc"
+	lc_state "github.com/0xsoniclabs/sonic/scc/light_client/light_client_state"
+	"github.com/0xsoniclabs/sonic/scc/light_client/provider"
+	"github.com/Fantom-foundation/lachesis-base/inter/idx"
+)
+
+// LightClient is the main entry point for the light client.
+// It is responsible for managing the light client state and
+// interacting with the provider.
+type LightClient struct {
+	provider provider.Provider
+	state    *lc_state.State
+}
+
+// Config is used to configure the LightClient.
+// It requires an url for the certificate providers and an initial committee.
+type Config struct {
+	Provider string
+	Genesis  scc.Committee
+}
+
+// NewLightClient creates a new LightClient with the given config.
+// returns an error if the config does not contain a valid provider or committee.
+func NewLightClient(config Config) (*LightClient, error) {
+	if err := config.Genesis.Validate(); err != nil {
+		return nil, fmt.Errorf("invalid committee provided: %w", err)
+	}
+	p, err := provider.NewServerFromURL(config.Provider)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create provider: %w\n", err)
+	}
+	return &LightClient{
+		state: lc_state.NewState(
+			scc.NewCommittee(
+				slices.Clone(config.Genesis.Members())...)),
+		provider: p,
+	}, nil
+}
+
+// Close closes the light client provider.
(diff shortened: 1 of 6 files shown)


scc/light_client/light_client.go
```diff
@@ -0,0 +1,56 @@
...
+		return nil, fmt.Errorf("invalid committee provided: %w", err)
+	}
+	p, err := provider.NewServerFromURL(config.Provider)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create provider: %w\n", err)
+	}
+	return &LightClient{
+		state: lc_state.NewState(
+			scc.NewCommittee(
+				slices.Clone(config.Genesis.Members())...)),
```
HerbertJordan, line 41: As discussed, some slice-cloning steps could be omitted here.

scc/light_client/light_client.go
```diff
@@ -0,0 +1,56 @@
...
+// interacting with the provider.
+type LightClient struct {
+	provider provider.Provider
+	state    *lc_state.State
+}
+
+// Config is used to configure the LightClient.
+// It requires an url for the certificate providers and an initial committee.
+type Config struct {
+	Provider string
```
HerbertJordan, line 24: nit: rename to "ProviderUrl" to indicate that this is expected to be a URL? Alternatively, use a URL type to indicate what this is supposed to be.

tests/integration_test_net.go
```diff
@@ -70,6 +71,9 @@ type IntegrationTestNetOptions struct {
 	NumNodes int
 	// ClientExtraArguments specifies additional arguments to be passed to the client.
 	ClientExtraArguments []string
+	// GenesisCommittee specifies the genesis committee to be used for the integration
+	// test network. If not set, a default committee is used.
+	GenesisCommittee scc.Committee
```
HerbertJordan, line 76: This can not be an option of the integration test network. The problem is that it would not be possible for the network nodes in the test network to figure out the private keys to act as the members of this committee.

Instead, the fake genesis is supposed to define a committee, start up the network using this committee, and the integration tests should retrieve the chosen genesis committee from the integration test network.

HerbertJordan (changes requested):
The light client looks good, but the integration test can not work like this. The genesis committee must come from the genesis process, and can not be imposed by the integration test setup.

Maybe split the PR into two, one for the light client files and one for the integration tests. The first can be merged, the second needs work.

tests/integration_test_net.go
facuMH: ok, I will drop the changes for the integration test net, and focus on the light client only for this PR

facuMH: https://github.com/0xsoniclabs/sonic/pull/120/commits/a6b4c6fec6b9b98d8f7358cecbed3f890e070ac8

scc/light_client/light_client.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/120/commits/700da771b2480c34fb638f60ece472a6ed674899

scc/light_client/light_client.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/120/commits/700da771b2480c34fb638f60ece472a6ed674899

HerbertJordan (approved):
Looks good. The PR description is still listing modifications to the genesis setup and the integration tests. This should be removed before closing this PR.

PR #122, opened 2025-03-17
Solve issue generating Tests flakiness. 

LuisPH3:
RPC queries are concurrent to block execution, the section modified is not atomic and queries may "see" partial results (for example, transactions having an assigned block, and the block is not yet available)

This new ordering solves integration test flakiness issues.

### DIFF
--- a/gossip/c_block_callbacks.go
+++ b/gossip/c_block_callbacks.go
@@ -342,6 +342,18 @@ func consensusCallbackBeginBlockFn(
 					bs.FinalizedStateRoot = hash.Hash(evmBlock.Root)
 					// At this point, block state is finalized
 
+					// Build index for not skipped txs
+					if txIndex {
+						// Index receipts
+						// Note: it's possible for receipts to get indexed twice by BR and block processing
+						if allReceipts.Len() != 0 {
+							store.evm.SetReceipts(blockCtx.Idx, allReceipts)
+							for _, r := range allReceipts {
+								store.evm.IndexLogs(r.Logs...)
+							}
+						}
+					}
+
 					bs.LastBlock = blockCtx
 					bs.CheatersWritten = uint32(bs.EpochCheaters.Len())
 					if sealing {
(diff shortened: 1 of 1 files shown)


LuisPH3:
Relates to  #122.
If this PR is accepted, #122 is no longer needed.

facuMH (dismissed):

gossip/c_block_callbacks.go
```diff
@@ -370,11 +353,34 @@ func consensusCallbackBeginBlockFn(
...
+							// have been stored, this prevents queries (RPC) executed
+							// concurrently from finding only one of the two components
+							// of this relationship.
+							store.evm.SetTxPosition(tx.Hash(), txPositions[tx.Hash()].TxPosition)
+						}
+
+						// Index receipts
+						// Note: it's possible for receipts to get indexed twice by BR and block processing
+						if allReceipts.Len() != 0 {
+							store.evm.SetReceipts(blockCtx.Idx, allReceipts)
```
HerbertJordan, line 377: The main use case for this index on receipts seems to be the `GetReceiptsByNumber` in the API [here](https://github.com/0xsoniclabs/sonic/blob/2b8135e7dbae50bed462784be4b929e40fac274e/gossip/ethapi_backend.go#L308).

This code fetches the block first and then retrieves all receipts. It thus seems to me that it depends on the receipts to be stored before the blocks. Otherwise an incorrect block without receipts may be returned. Right?

HerbertJordan (commented):
These changes need to be carefully studied. Unfortunately, there is no good mechanism to identify all the assumptions on the indexing order made by various parts of the code.

I tried to investigated some of the implications of this change and listed them into comments. Maybe you can have a look.

Since the flakiness of the tests has been traced back to another issue, the motivation for those changes may also be gone, right?

LuisPH3:
> Since the flakiness of the tests has been traced back to another issue, the motivation for those changes may also be gone, right?

Correct, once #124 is merged, we do not need this pr.

I think the core problem is the lack of atomicity on the store, this pr proposed a second-best option. PR #124 solves this issue from the consumer side: "if the answer is incomplete, assume not ready"

gossip/c_block_callbacks.go
LuisPH3: Right, I missed this one. The relationship is block -> receipts, it makes sense that those are stored first.

LuisPH3:
No longer needed

PR #112, opened 2025-03-13
Update referenced consensus module to 0xsoniclabs 

aksonic:
### Context:

This PR updates the usage of https://github.com/Fantom-foundation/lachesis-base to https://github.com/0xsoniclabs/consensus in sonic and updates all associated imports accordingly.

### Verification:

image

### Notes on versioning:

This PR integrates the consensus repository at commit https://github.com/0xsoniclabs/consensus/commit/76cd055a - the consensus repository at that commit hash has been updated to change all stale references of `Fantom-foundation/lachesis-base` to `0xsoniclabs/consensus`

The version of the consensus repository used will be updated with each PR - unfortunately this isn't too easily avoidable because some of the PRs involve migrating files from `consensus` over to `sonic` or have other breaking changes where a version bump is required. A (semver) versioned consensus release is anticipated but some refactoring work is to be conducted beforehand.

The image below is a rough reference of how the two will move with the `sonic` repository commits under our `consensus/refactor` branch linked above to indicate the anticipated changes and the matching `consensus` repository version that is planned to be used below for each PR.

image

PR #123, opened 2025-03-18
chore: revert consensus reference

aksonic:
This PR reverts the changes that were incorporated by PR https://github.com/0xsoniclabs/sonic/pull/112, in order to be as conservative as possible prior to the upcoming release. This PR reverts the referenced consensus module by `sonic` to use `Fantom-foundation/lachesis-base` and `Fantom-foundation/lachesis-base-sonic` exactly as they were used priorly - UNDOING the incorporated update to `0xsoniclabs/consensus` everywhere it is relevant.

**Please note:** In `go.mod`, the require `lachesis-base` followed by the `replace` directive to replace it with `lachesis-base-sonic` is exactly how it was done priorly 

Verification:

image


facuMH (approved):
This PR reverts the changes done before. It seems both go mod and sum are properly updated.

LuisPH3 (commented):
The code removes every use of 0xsoniclabs/consensus,
mod tidy is clean
search reveals not further uses.

PR #124, opened 2025-03-18
Fix rpc response when block not found

jenikd:
This PR fixes an issue, when error was returned instead of nil, when block was not found.
It is related to [this issue](https://github.com/0xsoniclabs/sonic-admin/issues/116) where problem is decribed

### DIFF
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -1747,19 +1747,15 @@ func (s *PublicTransactionPoolAPI) GetBlockReceipts(ctx context.Context, blockNr
 	if blockNr, ok := blockNrOrHash.Number(); ok {
 		number = blockNr
 		header, err = s.b.HeaderByNumber(ctx, number)
-		if err != nil {
+		if header == nil || err != nil {
 			return nil, err
 		}
 	} else if blockHash, ok := blockNrOrHash.Hash(); ok {
 		header, err = s.b.HeaderByHash(ctx, blockHash)
-		if err != nil {
+		if header == nil || err != nil {
 			return nil, err
 		}
 		number = rpc.BlockNumber(header.Number.Uint64())
-	}
-
-	if header == nil {
-		return nil, fmt.Errorf("block not found")
 	}
 
 	receipts, err := s.b.GetReceiptsByNumber(ctx, number)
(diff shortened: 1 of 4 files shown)


LuisPH3 (dismissed):
It seems complete to me:
`HeaderByNumber` returns nil instead of error when not found.

ethapi/block_overrides.go
```diff
@@ -78,7 +78,7 @@ func (context *chainContext) GetHeader(hash common.Hash, number uint64) *evmcore
 	// This method is called to get the hash for a block number when executing the BLOCKHASH
 	// opcode. Hence no need to search for non-canonical blocks.
 	header, err := context.b.HeaderByNumber(context.ctx, rpc.BlockNumber(number))
-	if err != nil || header.Hash != hash {
+	if header == nil || err != nil || header.Hash != hash {
```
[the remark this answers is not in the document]
jenikd: Done

ethapi/api.go
```diff
@@ -837,27 +837,27 @@ func (s *PublicBlockChainAPI) GetProof(ctx context.Context, address common.Addre
...
-	return nil, err
+	return s.getHeaderWithReceipts(ctx, header, rpc.BlockNumber(header.Number.Uint64()))
+}
+
+func (s *PublicBlockChainAPI) getHeaderWithReceipts(ctx context.Context, header *evmcore.EvmHeader, blkNumber rpc.BlockNumber) (*evmcore.EvmHeaderJson, error) {
+	receipts, err := s.getBlockReceipts(ctx, blkNumber)
+	if receipts == nil || err != nil {
+		return nil, err
+	}
+	return header.ToJson(receipts), nil
```
LuisPH3, line 860: :+1: thanks!

ethapi/block_overrides.go
```diff
@@ -78,7 +78,10 @@ func (context *chainContext) GetHeader(hash common.Hash, number uint64) *evmcore
 	// This method is called to get the hash for a block number when executing the BLOCKHASH
 	// opcode. Hence no need to search for non-canonical blocks.
 	header, err := context.b.HeaderByNumber(context.ctx, rpc.BlockNumber(number))
-	if header == nil || err != nil || header.Hash != hash {
+	if header == nil || err != nil {
+		return nil
+	}
+	if header.Hash != hash {
```
LuisPH3, line 84: This is now quite what I meant, sorry for not explaining myself. 
```go
if err == nil || header.Hash != hash {
    return nil
}
return header
```

This should be sufficient, if header is nil, it will already be returning nil.

LuisPH3 (approved):

ethapi/block_overrides.go
jenikd: But if header is nil, then header.Hash wil be nil pointer dereference

LuisPH3: right! my bad

PR #125, opened 2025-03-18
Fix originatedFees in epoch sealing - fixes 0xsoniclabs/sonic-admin#122

thaarok:
Fixes 0xsoniclabs/sonic-admin#122, fix is enabled by Allegro hardfork.

Update: Since receive.EffectiveGasPrice is not initialized in this part of the code, calculation from baseFee was added.

Successfully tested on local fakenet - originatedFees matches with EffectiveGasPrice in the receipt after this fix.

### DIFF
--- a/gossip/blockproc/drivermodule/driver_txs.go
+++ b/gossip/blockproc/drivermodule/driver_txs.go
@@ -128,7 +128,7 @@ func (p *DriverTxTransactor) PopInternalTxs(_ iblockproc.BlockCtx, _ iblockproc.
 	return internalTxs
 }
 
-func (p *DriverTxListener) OnNewReceipt(tx *types.Transaction, r *types.Receipt, originator idx.ValidatorID, baseFee *big.Int) {
+func (p *DriverTxListener) OnNewReceipt(tx *types.Transaction, r *types.Receipt, originator idx.ValidatorID, baseFee *big.Int, blobBaseFee *big.Int) {
 	if originator == 0 {
 		return
 	}
@@ -142,6 +142,13 @@ func (p *DriverTxListener) OnNewReceipt(tx *types.Transaction, r *types.Receipt,
 		gasPrice = tx.GasPrice()
 	}
 	txFee := new(big.Int).Mul(new(big.Int).SetUint64(r.GasUsed), gasPrice)
+
+	if r.BlobGasUsed != 0 && blobBaseFee != nil {
+		blobFee := new(big.Int).SetUint64(r.BlobGasUsed)
+		blobFee.Mul(blobFee, blobBaseFee)
+		txFee.Add(txFee, blobFee)
+	}
+
 	originated := p.bs.ValidatorStates[originatorIdx].Originated
 	originated.Add(originated, txFee)
 
--- a/evmcore/dummy_block.go
+++ b/evmcore/dummy_block.go
@@ -48,7 +48,8 @@ type (
 
 		WithdrawalsHash *common.Hash
 
-		BaseFee *big.Int
+		BaseFee     *big.Int
+		BlobBaseFee *big.Int // TODO issue #147
 
 		PrevRandao common.Hash // == mixHash/mixDigest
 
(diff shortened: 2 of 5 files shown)


gossip/blockproc/drivermodule/driver_txs.go
```diff
@@ -128,14 +128,20 @@ func (p *DriverTxTransactor) PopInternalTxs(_ iblockproc.BlockCtx, _ iblockproc.
...
 	if originator == 0 {
 		return
 	}
 	originatorIdx := p.es.Validators.GetIdx(originator)
 
 	// track originated fee
-	txFee := new(big.Int).Mul(new(big.Int).SetUint64(r.GasUsed), tx.GasPrice())
+	var gasPrice *big.Int
+	if p.es.Rules.Upgrades.Allegro {
+		gasPrice = effectiveGasPrice(tx, baseFee)
```
HerbertJordan, line 140: The `Receipt` contains the effective gas price of this transaction. To avoid redundancy, I would like to ask to take the price from there, such  that if in the future the gas price calculation is changed again, we do not risk to miss updating this function here.

HerbertJordan (changes requested):
Looks good, but we should avoid introducing a redundant computation of the effective gas price, as it can be easily overlooked if the gas-price computation changes again in the future. I added a corresponding comment, please have a look.

gossip/blockproc/drivermodule/driver_txs.go
thaarok: `receive.EffectiveGasPrice` is not initialized in this part of the code - I would have to call `Receipts.DeriveFields()` which seems unnecessary demanding for this purpose.
Other option is to publish effectiveGasPrice() from the go-ethereum-sonic, but that would be another code into our go-ethereum patch, which we prefer to avoid.

HerbertJordan: I spend an hour searching through the code to find a better way to support this than to avoid the re-implement the gas price computation here.

The relevant implementation seems to be [here](https://github.com/0xsoniclabs/go-ethereum/blob/86eca3554809383eb0068b6321221f3bfe9dbd6c/core/state_transition.go#L188-L193). In this code computing the effective gas price used in the transaction processor [here](https://github.com/0xsoniclabs/go-ethereum/blob/86eca3554809383eb0068b6321221f3bfe9dbd6c/core/state_transition.go#L262). Unfortunately, it is part of a function that performs a lot more operations, including the retrieval of the sender from the transaction, that makes it not easily reusable for our use case.

An idea: we factor out the gas-price computation from [here](https://github.com/0xsoniclabs/go-ethereum/blob/86eca3554809383eb0068b6321221f3bfe9dbd6c/core/state_transition.go#L188-L193) into a stand-alone function, which can then also be used by the `driver_txs.go`. If then, in the future, the gas-pricing would change again for whatever reason, both use cases would use the same implementation implicitly.

gossip/blockproc/drivermodule/driver_txs.go
```diff
@@ -128,14 +128,20 @@ func (p *DriverTxTransactor) PopInternalTxs(_ iblockproc.BlockCtx, _ iblockproc.
...
+	} else {
+		gasPrice = tx.GasPrice()
+	}
+	txFee := new(big.Int).Mul(new(big.Int).SetUint64(r.GasUsed), gasPrice)
```
HerbertJordan, line 144: Although we do not support it right now, the blob gas price should be considered here as well.

gossip/blockproc/drivermodule/driver_txs.go
HerbertJordan: I just found this [here](https://github.com/0xsoniclabs/go-ethereum/blob/86eca3554809383eb0068b6321221f3bfe9dbd6c/core/types/transaction.go#L374); This might allow us to compute the effective price without the need to do any patches like this:
```
func effectiveGasPrice(tx *types.Transaction, baseFee *big.Int) *big.Int {
	if baseFee == nil {
		return tx.GasPrice()
	}
	return new(big.Int).Add(baseFee, tx.EffectiveGasTipValue(baseFee))
}
```
Wdyt?

thaarok: It seemed not much straightforward when I was considering it, but you are true, this may be better.

HerbertJordan (approved):

PR #126, opened 2025-03-19
Add pre-deployed HistroyStorage Contract to fake nets started with Allegro

LuisPH3:
This PR adds the [EIP-2935](https://eips.ethereum.org/EIPS/eip-2935) pre-deployed contract to test infrastructure.

### DIFF
--- a/integration/makefakegenesis/genesis.go
+++ b/integration/makefakegenesis/genesis.go
@@ -12,6 +12,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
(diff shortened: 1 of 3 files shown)


HerbertJordan (approved):
Very nice, thanks!

simonlechner (approved):
Very nice that it is so easy to test with the test infrastructure.
Looks good, I expected a test checking that the contract is doing what it is supposed to do. But since it is taken from ethereum I think we can just take it as is and be ethereum compatible.

LuisPH3:
> Very nice that it is so easy to test with the test infrastructure. Looks good, I expected a test checking that the contract is doing what it is supposed to do. But since it is taken from ethereum I think we can just take it as is and be ethereum compatible.

Correct, this is one of several tests. There will be functional tests.

PR #128, opened 2025-03-19
Add Retry provider to handle request failures

facuMH:
This PR introduces the `retryProvider` type, which implements the `provider` interface while also wrapping an existing `provider` and automatically retries failed certificate fetch attempts. It follows the Decorator Pattern, ensuring that retry logic is separate from core provider implementations.
Additionally, the `retries` and `timeout` parameters are added to `light_client.Config`.

### DIFF
--- a/scc/light_client/light_client.go
+++ b/scc/light_client/light_client.go
@@ -22,9 +22,9 @@ type LightClient struct {
 type Config struct {
 	Url     *url.URL
 	Genesis scc.Committee
-	// unless specified only try once without delays
-	Attempts uint
-	Delay    time.Duration
+	// default values try only once without delays
+	Retries uint
+	Delay   time.Duration
 }
 
 // NewLightClient creates a new LightClient with the given config.
@@ -38,7 +38,7 @@ func NewLightClient(config Config) (*LightClient, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to create provider: %w\n", err)
 	}
-	p = newRetry(p, config.Attempts, config.Delay)
+	p = newRetry(p, config.Retries, config.Delay)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create retry provider: %w\n", err)
 	}
--- a/scc/light_client/light_client.go
+++ b/scc/light_client/light_client.go
@@ -22,7 +22,7 @@ type LightClient struct {
 type Config struct {
 	Url     *url.URL
 	Genesis scc.Committee
-	// default values try only once per request with 10 seconds timeout
+	// By default, requests are retried up to 1024 times to reach a 10-second timeout.
 	Retries uint
 	Timeout time.Duration
 }
--- a/scc/light_client/light_client.go
+++ b/scc/light_client/light_client.go
@@ -22,9 +22,9 @@ type LightClient struct {
 type Config struct {
 	Url     *url.URL
 	Genesis scc.Committee
-	// default values try only once without delays
+	// default values try only once per request with 10 seconds timeout
 	Retries uint
-	Delay   time.Duration
+	Timeout time.Duration
 }
 
 // NewLightClient creates a new LightClient with the given config.
@@ -38,10 +38,7 @@ func NewLightClient(config Config) (*LightClient, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to create provider: %w\n", err)
 	}
-	p = newRetry(p, config.Retries, config.Delay)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create retry provider: %w\n", err)
-	}
+	p = newRetry(p, config.Retries, config.Timeout)
 	return &LightClient{
 		state:    *newState(config.Genesis),
 		provider: p,
--- a/scc/light_client/light_client.go
+++ b/scc/light_client/light_client.go
@@ -3,6 +3,7 @@ package light_client
 import (
 	"fmt"
 	"net/url"
+	"time"
 
 	"github.com/0xsoniclabs/sonic/scc"
 	"github.com/Fantom-foundation/lachesis-base/inter/idx"
@@ -21,6 +22,9 @@ type LightClient struct {
 type Config struct {
 	Url     *url.URL
 	Genesis scc.Committee
+	// unless specified only try once without delays
+	Attempts uint
+	Delay    time.Duration
 }
 
 // NewLightClient creates a new LightClient with the given config.
@@ -29,9 +33,14 @@ func NewLightClient(config Config) (*LightClient, error) {
 	if err := config.Genesis.Validate(); err != nil {
 		return nil, fmt.Errorf("invalid committee provided: %w", err)
 	}
+	var p provider
 	p, err := newServerFromURL(config.Url.String())
 	if err != nil {
 		return nil, fmt.Errorf("failed to create provider: %w\n", err)
+	}
+	p = newRetry(p, config.Attempts, config.Delay)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create retry provider: %w\n", err)
 	}
 	return &LightClient{
 		state:    *newState(config.Genesis),
--- a/scc/light_client/retry.go
+++ b/scc/light_client/retry.go
@@ -11,38 +11,42 @@ import (
 )
 
 // retryProvider is used to wrap another provider and add a retry mechanism.
-// It is a provider that retries provider methods, that returned an error, a
-// specified number of times with some delay between retries up to a maximum timeout.
+// It is a provider that maxRetries provider methods, that returned an error, a
+// specified number of times with some delay between maxRetries up to a maximum timeout.
 //
 // Fields:
 // - provider: The underlying provider to retry requests on.
-// - retries: The maximum number of attempt to ask for certificates.
+// - maxRetries: The maximum number of attempt to ask for certificates.
 // - timeout: The time max time willing to wait for a request without error.
 type retryProvider struct {
-	provider provider
-	retries  uint
-	timeout  time.Duration
+	provider   provider
+	maxRetries uint
+	timeout    time.Duration
 }
 
 // newRetry creates a new retryProvider provider with the given provider and maximum
-// number of retries and total timeout.
+// number of maxRetries and total timeout.
 //
 // Parameters:
 //   - provider: The underlying provider to wrap with retry logic.
-//   - retries: The maximum number of attempt to ask for certificates.
+//   - maxRetries: The maximum number of attempt to ask for certificates. If it
+//     is zero value, then maxRetries is 1024.
 //   - timeout: The time max time willing to wait. if it is zero value,
 //     then timeout is 10 seconds.
 //
 // Returns:
 // - *retryProvider: A new retryProvider provider instance.
-func newRetry(provider provider, retries uint, timeout time.Duration) *retryProvider {
+func newRetry(provider provider, maxRetries uint, timeout time.Duration) *retryProvider {
 	if timeout == 0 {
 		timeout = 10 * time.Second
 	}
+	if maxRetries == 0 {
+		maxRetries = 1024
+	}
 	return &retryProvider{
-		provider: provider,
-		retries:  retries,
-		timeout:  timeout,
+		provider:   provider,
+		maxRetries: maxRetries,
+		timeout:    timeout,
 	}
 }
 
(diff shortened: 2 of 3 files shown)


scc/light_client/light_client.go
```diff
@@ -21,6 +22,9 @@ type LightClient struct {
 type Config struct {
 	Url     *url.URL
 	Genesis scc.Committee
+	// unless specified only try once without delays
```
HerbertJordan, line 25: nit: since technically these fields are always set to some value (=specified), it would be better to state how the zero value is interpreted.

scc/light_client/retry.go
```diff
@@ -0,0 +1,115 @@
...
+
+// retryP is a provider that retries requests a certain number of times on failed
+// requests for certificates with a determined wait between retries. It is used
+// to wrap other providers to add a retry mechanism.
+//
+// Fields:
+// - provider: The underlying provider to retry requests on.
+// - attempts: The maximum number of attempt to ask for certificates.
+// - delay: The time to wait between retries.
+type retryP struct {
```
HerbertJordan, line 21: nit: no appropriations; we have enough storage capacity and horizontal screen space to use full names like `retryProvider`.

scc/light_client/retry.go
```diff
@@ -0,0 +1,115 @@
...
+	"time"
+
+	"github.com/0xsoniclabs/sonic/scc"
+	"github.com/0xsoniclabs/sonic/scc/cert"
+	"github.com/Fantom-foundation/lachesis-base/inter/idx"
+)
+
+// retryP is a provider that retries requests a certain number of times on failed
+// requests for certificates with a determined wait between retries. It is used
+// to wrap other providers to add a retry mechanism.
```
HerbertJordan, line 15: After reading this and before looking into the implementation I was wondering: what is considered a failed request? If the request produces an error? If some requested certificates are missing? If some certificates are invalid?

I think a description of what is considered a failed request helps to understand better what can be expected from this decorator.

scc/light_client/retry.go
```diff
@@ -0,0 +1,115 @@
...
+// - []cert.CommitteeCertificate: A slice of committee certificates.
+// - error: An error if the provider failed to obtain the requested certificates.
+func (r retryP) getCommitteeCertificates(first scc.Period, maxResults uint64) ([]cert.CommitteeCertificate, error) {
+	result, err := retry(r.attempts, r.delay, func() (any, error) {
+		return r.provider.getCommitteeCertificates(first, maxResults)
+	})
+	if err != nil {
+		return nil, err
+	}
+	return result.([]cert.CommitteeCertificate), nil
```
HerbertJordan, line 68: The `retry` function is generic, so why mixing generics with reflection?
```suggestion
	return retry(r.attempts, r.delay, func() ([]cert.CommitteeCertificate, error) {
		return r.provider.getCommitteeCertificates(first, maxResults)
	})
```

scc/light_client/retry.go
```diff
@@ -0,0 +1,115 @@
...
+// Parameters:
+// - first: The starting block number for which to retrieve the block certificate.
+// - maxResults: The maximum number of block certificates to retrieve.
+//
+// Returns:
+//   - []cert.BlockCertificate: The block certificates for the given block number
+//     and the following blocks.
+//   - error: An error if the provider failed to obtain the requested certificates.
+func (r retryP) getBlockCertificates(first idx.Block, maxResults uint64) ([]cert.BlockCertificate, error) {
+	result, err := retry(r.attempts, r.delay, func() (any, error) {
```
HerbertJordan, line 83: Also here, there is no need for mixing `any` with generics.

scc/light_client/retry.go
```diff
@@ -0,0 +1,115 @@
...
+//     The joined error of all failed attempts if all attempts failed.
+func retry[C any](attempts uint, delay time.Duration, fn func() (C, error)) (C, error) {
+	var errs []error
+	for i := uint(0); i < attempts; i++ {
+		result, err := fn()
+		if err == nil {
+			return result, nil
+		}
+		errs = append(errs, err)
+		time.Sleep(delay)
```
HerbertJordan, line 110: Some "product level" considerations: the aim of retry mechanisms is to balance between incurred latency and introduced extra network load. While the given implementation is OK, a more common approach are exponential back-off periods, like for instance done [here](https://github.com/0xsoniclabs/sonic/blob/ab8a058aacc2a4e62b523b3787b34c2be29c9e09/tests/integration_test_net.go#L668-L687).

Exponential back-offs have the advantage that they minimize the delay if the error was caused by a sporadic event, while it quickly reduces the traffic load if there is a persistent problem.

From a user-perspective it might also be simpler to specify a total "deadline" for a request over a number of attempts and delays. We might even pass a [Context](https://pkg.go.dev/context#Context) parameter to the `Sync` and the provider functions to allow the user to set a deadline per call and to even cancel a sync call if desired. This might be an idea for a follow-up PR.

scc/light_client/retry_test.go
```diff
@@ -0,0 +1,142 @@
...
+	provider := NewMockprovider(ctrl)
+
+	// fail once
+	provider.EXPECT().getCommitteeCertificates(gomock.Any(), gomock.Any()).
+		Return(nil, fmt.Errorf("some error")).Times(1)
+	// then succeed
+	provider.EXPECT().getCommitteeCertificates(gomock.Any(), gomock.Any()).
+		Return([]cert.CommitteeCertificate{{}}, nil).Times(1)
+
+	delay := 200 * time.Millisecond
```
HerbertJordan, line 84: nit: not relevant for the test, could be 0

scc/light_client/retry_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+	start := time.Now()
+	certs, err := retry(3, delay, func() (any, error) {
+		return provider.getCommitteeCertificates(scc.Period(1), uint64(1))
+	})
+	duration := time.Since(start)
+
+	require.NoError(err)
+	require.NotNil(certs)
+	require.Less(duration.Milliseconds(), 2*delay.Milliseconds())
```
HerbertJordan, line 94: nit: not relevant for the objective of this test; should be removed

HerbertJordan (changes requested):
Looks good, but there are a few things that should be improved.

scc/light_client/retry.go
facuMH: well, I considering that what is a "failed" request depends on the implementation used for the internal provider. I think a more clear description would be to just clarify "retries requests that return an error"

scc/light_client/light_client.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/128/commits/91dfbe1c5d1a763e2d73c147f8e929977c9b9f00

scc/light_client/retry.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/128/commits/91dfbe1c5d1a763e2d73c147f8e929977c9b9f00

scc/light_client/retry.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/128/commits/91dfbe1c5d1a763e2d73c147f8e929977c9b9f00

scc/light_client/retry.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/128/commits/91dfbe1c5d1a763e2d73c147f8e929977c9b9f00

scc/light_client/retry_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/128/commits/91dfbe1c5d1a763e2d73c147f8e929977c9b9f00

scc/light_client/retry_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/128/commits/91dfbe1c5d1a763e2d73c147f8e929977c9b9f00

scc/light_client/retry.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/128/commits/de529598ba492eedce0360a7d906e4e0dd571af9

facuMH: timeout and incremental delays are implemented. Using a `Context` would be nice for a follow-up PR.

scc/light_client/light_client.go
```diff
@@ -21,6 +22,9 @@ type LightClient struct {
 type Config struct {
 	Url     *url.URL
 	Genesis scc.Committee
+	// default values try only once per request with 10 seconds timeout
+	Retries uint
+	Timeout time.Duration
```
HerbertJordan, line 27: nit: the `Retries` seems to be redundant and would require knowledge about the internal implementation details to know how to balance it against the `Timeout` parameter. Are 50 retries enough to run into a time-out of 15 seconds? Or do I need to set up at least 500 retries for that time frame?

However, I think the ultimate fix for this will be to use a `Context` as discussed, which will replace those anyway. So let's keep this for the `Context` adding PR.

scc/light_client/light_client.go
```diff
@@ -21,6 +22,9 @@ type LightClient struct {
 type Config struct {
 	Url     *url.URL
 	Genesis scc.Committee
+	// default values try only once per request with 10 seconds timeout
```
HerbertJordan, line 25: nit: it seems a bit odd that the default "retry" wrapper is essentially disabling itself. If you only try once, there is not much of a retry.

scc/light_client/retry.go
```diff
@@ -0,0 +1,121 @@
...
+//
+// Returns:
+//   - C: The result of the function if it succeeded.
+//   - error: Nil if at least one execution of fn returned without error.
+//     The joined error of all failed retries if all calls to fn failed.
+func retry[C any](retries uint, timeout time.Duration, fn func() (C, error)) (C, error) {
+	var errs []error
+	now := time.Now()
+	delay := time.Millisecond
+	maxDelay := 128 * time.Millisecond
```
HerbertJordan, line 102: nit: this timing is a bit aggressive when addressing a real web-server, as it would produce ~8 queries per second at the minimum. I would increase it to more like a second or two for the `maxDelay`. Also the initial delay should be increased to ~200 milliseconds, as otherwise it would just send the initial few queries in very quick succession.

Note: the maximum delay and latency in the unit test example I was providing a link to is intended to be very short as it only sends those requests to the local process; these requests do not even leave the address space, and thus can be set up at a higher rate to minimize delay during unit testing.

scc/light_client/retry.go
```diff
@@ -0,0 +1,121 @@
...
+	for i := uint(0); i < retries+1; i++ {
+		result, err := fn()
+		if err == nil {
+			return result, nil
+		}
+		errs = append(errs, err)
+		if delay < maxDelay {
+			delay *= 2
+		}
```
HerbertJordan, line 111: Unless the max delay is exactly a power-of-two times bigger than the minimum delay, this variant is safer:
```suggestion
        delay = 2*delay
		if delay > maxDelay {
			delay = maxDelay
		}
```
Otherwise, the actual `delay` could get up to 2x the desired `maxDelay`.

HerbertJordan (approved):
Looks alright, the remaining comments can be considered ideas for follow-up PRs.

scc/light_client/light_client.go
facuMH: https://github.com/0xsoniclabs/sonic-admin/issues/126

scc/light_client/light_client.go
facuMH: exactly, the idea for this was that one (as `LightClient` user) could disable the retry mechanism just by changing the config.

scc/light_client/retry.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/128/commits/1be35a75a4752a671ca398509b4002d0b30fac7f

scc/light_client/retry.go
facuMH: initial delayed changed to 200 miliseconds. max delay capped at 1 second. https://github.com/0xsoniclabs/sonic/pull/128/commits/1be35a75a4752a671ca398509b4002d0b30fac7f

scc/light_client/light_client.go
facuMH: after an offline discussion, the default behaviour will be similar to the timeout. when 0 the `newRetry` factory will retry 1024, which will most likely hit the timeout first. 
https://github.com/0xsoniclabs/sonic/pull/128/commits/1be35a75a4752a671ca398509b4002d0b30fac7f

HerbertJordan (approved):

PR #129, opened 2025-03-20
Integration tests: Allow use of custom configuration in test nodes.

LuisPH3:
This PR adds a mechanism to change the configuration of the integration test net nodes. This is required because some flags are not exposed via command line flags. 

The changes here presented compromises in the following:
- The configuration is saved from a constructed node, instead of saving the configuration loaded directly. This is done this way because the initialization of the node contains accesses to the command line arguments at several stages of initialization. Making quite difficult to identify when is the configuration complete. This code aims to reduce code changes in that section while introducing the new feature. 
- Not every configuration field can be modified using this mechanism. The configuration structure contains both, non-serializable objects, and explicitly ignored objects. This means that modifications may not be propagated into the integration test net node. 

This PR contributes to https://github.com/0xsoniclabs/sonic-admin/issues/124

### DIFF
--- a/cmd/sonicd/app/launcher.go
+++ b/cmd/sonicd/app/launcher.go
@@ -104,6 +104,7 @@ func initFlags() {
 		flags.ExitWhenEpochFlag,
 		flags.LightKDFFlag,
 		flags.ConfigFileFlag,
+		flags.DumpConfigFileFlag,
 		flags.ValidatorIDFlag,
 		flags.ValidatorPubkeyFlag,
 		flags.ValidatorPasswordFlag,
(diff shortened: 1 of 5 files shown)


LuisPH3:
After reconsidering using "sonictool dumpconfig" I decided not do so:
The problem is the same as the above described issue related to the initialization of the node configuration. The default configuration generated by `sonictool dumpconfig` defaults many parameters, and the integration network does not behave like other instances (tests timeout and other issues).

This Pr introduces a mechanism to retrieve the "effective" configuration file, after parsing all the command line arguments.

HerbertJordan (approved):
Nice addition to our infrastructure :)

thaarok (approved):

PR #130, opened 2025-03-20
Add test to check revision is forwarded correctly

simonlechner:
To ensure the revision is forward correctly to the interpreter an integration test is added. This checks that the access cost for the delegation designation address is taken into account if Prague is enabled.

### DIFF
--- a/tests/integration_test_net.go
+++ b/tests/integration_test_net.go
@@ -78,6 +78,8 @@ type IntegrationTestNetOptions struct {
 	// will not be saved into the toml file, modifications will be ignored.
 	// Zero value means no modification.
 	ModifyConfig func(*config.Config)
+	// Accounts to be deployed with the genesis.
+	Accounts []makefakegenesis.Account
 }
 
 // IntegrationTestNet is a in-process test network for integration tests. When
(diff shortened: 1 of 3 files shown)


tests/integration_test_net.go
```diff
@@ -169,6 +171,8 @@ func StartIntegrationTestNetWithJsonGenesis(
 		effectiveOptions.FeatureSet,
 	)
 
+	jsonGenesis.Accounts = append(jsonGenesis.Accounts, effectiveOptions.Accounts...)
```
HerbertJordan, line 174: In the function `StartIntegrationTestNetWithFakeGenesis` it should be tested that no extra accounts are specified. If so, the function should panic with a message telling the user that this feature is only supported for the JsonGenesis.

HerbertJordan (dismissed):
Nice test and integration test feature extension, thanks!

HerbertJordan (dismissed):

LuisPH3 (approved):
The test proposed checks that the delegate address resolved from a delegation designator is charged according to cold rule. 

Idea: because the test uses a accessListTx, we could pre-load address 42 to it, and the gas cost should be +2400+100, this would cover the warm case.

simonlechner:
> Idea: because the test uses a accessListTx, we could pre-load address 42 to it, and the gas cost should be +2400+100, this would cover the warm case.

This is already tested in the interpreter. For the aim of this test, to test that the transaction is processed with the right revision, it is sufficient to look at just one case.

LuisPH3 (approved):

LuisPH3 (approved):

PR #131, opened 2025-03-21
Test deployment of eip-2935

LuisPH3:
The HistoyStorage contract needs to be stored in a fix address, as decribed in https://eips.ethereum.org/EIPS/eip-2935.

> The bytecode will be deployed à la [EIP-4788](https://eips.ethereum.org/EIPS/eip-4788).

This means the use of a pre replay-protection transaction introduced in https://eips.ethereum.org/EIPS/eip-155.
This requires manual configuration of the node, for the RPC to accept this kind of transaction. 

This PR concludes https://github.com/0xsoniclabs/sonic-admin/issues/124

### DIFF
--- a/tests/eip2935_test.go
+++ b/tests/eip2935_test.go
@@ -2,16 +2,20 @@ package tests
 
 import (
 	"context"
+	"math/big"
 	"testing"
 
+	"github.com/0xsoniclabs/sonic/config"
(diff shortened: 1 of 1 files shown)


HerbertJordan (approved):
Great work, thanks!

PR #132, opened 2025-03-21
Proof of concept of onchain SCC light client

jenikd:
This PR is a try of BLS on chain signature verification using precompiled contract introduced in Prague/Pectra  with [eip-2537](https://eips.ethereum.org/EIPS/eip-2537)


tests/contracts/blsContracts/BLS.sol
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    function CheckAndUpdate(bytes memory pubKey, bytes memory signature, bytes memory message) public {
+        bool res  = CheckSignature(pubKey, signature, message);
+        if (res == true) {
+            Signature = signature;
+        }
+    }
+
+    function CheckAgregatedSignature(bytes memory pubKeys, bytes memory signature, bytes memory message) public view returns (bool){
```
HerbertJordan, line 31: nit:
```suggestion
    function CheckAggregatedSignature(bytes memory pubKeys, bytes memory signature, bytes memory message) public view returns (bool){
```

tests/contracts/blsContracts/BLS.sol
```diff
@@ -0,0 +1,47 @@
...
+    function EncodeToG2(bytes memory message) public view returns (bytes memory){
+        return BLSLibrary.EncodeToG2(message);
+    }
+
+    function CheckSignature(bytes memory pubKey, bytes memory signature, bytes memory message) public view returns (bool){
+        // check length of input parameters
+        require(pubKey.length == 128, "Invalid public key length");
+        require(signature.length == 256, "Invalid signature length");
+
+        // hash message and do pairing
```
HerbertJordan, line 19: nit:
```suggestion
        // hash message and check pairing
```

tests/contracts/blsContracts/BLSLibrary.sol
```diff
@@ -0,0 +1,235 @@
...
+    // Negated generator of G1 to be used for pairing function
+    bytes constant negG1 =
+        hex"0000000000000000000000000000000017f1d3a73197d7942695638c4fa9ac0fc3688c4f9774b905a14e3a3f171bac586c55e83ff97a1aeffb3af00adb22c6bb00000000000000000000000000000000114d1d6855d545a8aa7d76c8cf2e21f267816aef1db507c96655b9d5caac42364e6f38ba0ecb751bad54dcd6b939c2ca";
+
+    // This defines how many fp elements have to be created when hashing message
+    // For EncodeToG2 this is 2
+    // For HashToG2 this is 2*2
+    uint16 constant elCount = 2;
+
+    uint16 private constant L = 64;
```
HerbertJordan, line 22: nit: only used at one place and the name `L` does not tell a lot what it is for ... maybe inline this?

tests/contracts/blsContracts/BLSLibrary.sol
```diff
@@ -0,0 +1,235 @@
...
+    ) public view returns (bool) {
+        bytes memory res = precompilePair(
+                abi.encodePacked(negG1, signature, pubKey, messageHash)
+            );
+        require(res.length == 32, "Invalid result length");
+        return res[31] != 0;
+    }
+
+    // All input parameters are points on G1 or G2 as bytes
+    function PairAgregateSignature(
```
HerbertJordan, line 67: nit:
```suggestion
    function PairAggregatedSignature(
```

tests/contracts/blsContracts/BLSLibrary.sol
```diff
@@ -0,0 +1,235 @@
...
+                elArray[i].val[2],
+                elArray[i].val[1],
+                elArray[i].val[0]
+            );
+        }
+        return precompileMaptoG2(fpBytes);
+    }
+
+    // All input parameters are points on G1 or G2 as bytes
+    function Pair(
```
HerbertJordan, line 54: How about renaming this to "CheckSignature" or "IsValidSignature" if you want to stress that the function returns a boolean and does not revert the contract in case of a failure?

The function is internally using the `pair` function to check the the signature, but it is not actually returning the result of the pairing.

tests/contracts/blsContracts/BLSLibrary.sol
HerbertJordan, line 67: Maybe rename to "CheckAggregatedSignature" for the same reasons as for the "Pair" function?

tests/contracts/blsContracts/BLSLibrary.sol
```diff
@@ -0,0 +1,235 @@
...
+            res[i] = setBigIntToElement(v, modulo);
+        }
+
+        return res;
+    }
+
+    function setBigIntToElement(BigNumber memory v, BigNumber memory modulo) private view returns (Elements.Element memory) {
+        
+        // returning Elements.toMont(Elements.ElementFromBytes(vv.val));  //this is working to get correct field point but with Montgomery reduction
+        // returning Elements.ElementFromBytes(v.val); // correct field point without Montgomery reduction, but working for MapToG2
```
HerbertJordan, line 218: This seems to contain some important insights, but in its current form it looks like obsolete code that should be removed. Can you convert the insights covered in the comments into some code comment for future reference?

tests/contracts/blsContracts/BLS.sol
```diff
@@ -0,0 +1,47 @@
...
+        // check length of input parameters
+        require(pubKeys.length%128 == 0, "Invalid public keys length");
+        require(signature.length == 256, "Invalid signature length");
+
+        // hash message and do pairing
+        bytes memory msgHashG2 = BLSLibrary.EncodeToG2(message);
+        return BLSLibrary.PairAgregateSignature(pubKeys,signature,msgHashG2);
+    }
+
+    function CheckAndUpdateAgregatedSignature(bytes memory pubKeys, bytes memory signature, bytes memory message) public{
```
HerbertJordan, line 41:
```suggestion
    function CheckAndUpdateAggregatedSignature(bytes memory pubKeys, bytes memory signature, bytes memory message) public{
```

tests/contracts/blsContracts/BigNumber.sol
HerbertJordan, line 1: I assume this is from some external source? If so, can you add a link to the file header?

tests/contracts/blsContracts/Elements.sol
```diff
@@ -0,0 +1,411 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.26;
+
+library Elements {
```
HerbertJordan, line 4: Some documentation on what this library provides would be nice. In particular, some information on what an "Element" is, or some link to read up on this.

If this is taken from some third-party source, a line documenting the origin would be good as well.

tests/bls_onchain_test.go
```diff
@@ -0,0 +1,248 @@
...
+	blsContract, _, err := DeployContractWithOpts(net, blsContracts.DeployBLS, txOptions)
+	require.NoError(t, err, "failed to deploy contract; %v", err)
+
+	// Test different bls libraries verification
+	tests := []struct {
+		name     string
+		detaFunc func(pk bls.PrivateKey, message []byte) ([]byte, []byte, error)
+	}{
+		{"gnark", getGnarkData},
+		{"bls", getBlsData},
```
HerbertJordan, line 35: If I understand correctly, both of these data sources are required to produce exactly the same data for a given private key and message. Checking this, is beyond the scope of an integration test. This would be better placed in an unit test for the BLS library.

I think in this context you can pick one of these, simplifying the structure of this test.

tests/bls_onchain_test.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			runTest(t, net, blsContract, test.detaFunc)
```
HerbertJordan, line 39: I personally that factoring out test code like here makes them harder to understand. At the call side here, I can not understand what is being tested, and at the definition of the `runTest` function I lack the context of the setup.

Since `runTest` is also not reused, and is a function name in the namespace of this entire package, I would suggest to avoid such helper functions and inline the code here. Wdyt?

HerbertJordan (changes requested):
Looks good to me, thanks for building this end-to-end test for us.

I added a few suggestions, mostly of a cosmetic nature. Please have a look, let me know what you think.

A consistent spelling issue that should be fixed is "Agregate" => "Aggregate".

I also think that the problem you faced with the deployment of the contracts is a general issue that should be fixed for all deployment operations. I prepared a patch for this and some minor other changes to the test in #144. Please have a look.

tests/contracts/blsContracts/BLSLibrary.sol
```diff
@@ -0,0 +1,235 @@
...
+            assembly {
+                let chunkPtr := add(chunk, 0x20) // Points to the start of the `chunk` data
+                let srcPtr := add(add(pseudoRandomBytes, 0x20), chunkStart) // Points to the correct position in `pseudoRandomBytes`
+                mstore(chunkPtr, mload(srcPtr)) // Copy 32 bytes
+                mstore(add(chunkPtr, 0x20), mload(add(srcPtr, 0x20))) // Copy the next 32 bytes
+            }
+
+            BigNumber memory v = BigNumbers.init(chunk, false);
+            res[i] = setBigIntToElement(v, modulo);
+        }
```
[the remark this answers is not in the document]
jenikd: Added links to ietf specs

tests/contracts/blsContracts/BLSLibrary.sol
jenikd: Made it inline with comment

tests/bls_onchain_test.go
jenikd: Simplified test and made it more readable

tests/bls_onchain_test.go
```diff
@@ -0,0 +1,158 @@
...
+				require.NoError(t, err, "failed to get receipt; %v", err)
+				t.Logf("gas used for updating signature: %v", receipt.GasUsed)
+
+				updatedSignature, err := blsContract.Signature(nil)
+				require.NoError(t, err, "failed to get updated signature; %v", err)
+				require.Equal(t, signatureBytes, updatedSignature, "signature has to be updated")
+			})
+		})
+	}
+}
```
HerbertJordan, line 80: Nice test, great improvement, thanks :)

HerbertJordan (approved):
Thanks for factoring the feedback and for building this proof of concept!

PR #133, opened 2025-03-21
EIP-2935: Implement processor routine to call history storage contract

LuisPH3:
Add and test changes for the eip-2935 https://eips.ethereum.org/EIPS/eip-2935

The history storage, a feature with similar semantics to the `BLOCKHASH` opcode with the following differences:
- It is implemented in a contract, therefore it does not rely on built-in storage, but instead hashes are stored in the blockchain itself. 
- It has a history of 8191 elements. 

How it works:
- When the contract is called for the `system-address`, it is a write operation, and the expected argument is the block hash. The costs of this operation are ignored and do not account for any block limit nor consume any balance. 
- When called from any other address, it is a read operation, and the expected argument is the block number. 

These changes are ported from the equivalent feature in geth: https://github.com/ethereum/go-ethereum/commit/a223efcf39855c92730e0f6a269c90d1e89fd7e4


tests/contracts/read_history_storage/read_history_storage.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+// ReadHistoryStorage uses the history storage contract to read block hashes
+// https://eips.ethereum.org/EIPS/eip-2935
+contract ReadHistoryStorage {
+    event BlockHash(uint256 queriedBlock, bytes32 blockHash);
```
HerbertJordan, line 7: idea: if you think this is not too much extra, this might be an easy opportunity to cross-check the hashes with the hashes produced by the BLOCKHASH opcode. You could emit the block number, the hash according to the pre-deployed contract, and the hash according to the OpCode in this event. To access the OpCode hash, a build-in  [blockhash](https://docs.soliditylang.org/en/latest/units-and-global-variables.html#block-and-transaction-properties) function in Solidity can be used.

But feel free to ignore this if you want to keep the test focused.

tests/eip2935_test.go
```diff
@@ -79,3 +83,65 @@ func TestEIP2935_HistoryContractIsNotDeployedBeforePrague(t *testing.T) {
...
+
+	// Fist loop just issues transactions to create blocks
+	hashes := make([]common.Hash, 0, testIterations)
+	for range testIterations {
+		receipt, err = net.Apply(func(opts *bind.TransactOpts) (*types.Transaction, error) {
+			return readHistoryStorageContract.ReadHistoryStorage(opts, new(big.Int))
+		})
+		require.NoError(t, err)
+		require.Equal(t, types.ReceiptStatusSuccessful, receipt.Status)
+		hashes = append(hashes, receipt.BlockHash)
```
HerbertJordan, line 118: I was for a moment confused about this, since it seems to always read the hash of block 0, but in the loop below you are checking the hashes collected in this loop. Toke me some time that you are actually collecting the hashes of the `receipt`s.

Maybe the operation performed to produce block could be some simpler operation like an account endowment?

tests/eip2935_test.go
```diff
@@ -79,3 +83,65 @@ func TestEIP2935_HistoryContractIsNotDeployedBeforePrague(t *testing.T) {
...
+	}
+
+	// second loop queries block hashes of the blocks generated in the first loop
+	for i := range testIterations {
+		blockNumber := startBlock + uint64(i) + 1
```
HerbertJordan, line 123: It is probably not a problem, but this setup is depending on timing. It assumes that there is no block produced between the time when the `startBlock` is queried and the first blocks created by the loop above. Also, no empty block may be emitted in-between iterations of the generator loop above.

An easy fix for eliminating this timing dependency would be to store hashes in a map, mapping block numbers to hashes, instead of storing them in a list. This way, sporadic in-between blocks would not cause problems.

evmcore/state_processor.go
```diff
@@ -73,6 +73,11 @@ func (p *StateProcessor) Process(
 		blockNumber  = block.Number
 		signer       = gsignercache.Wrap(types.MakeSigner(p.config, header.Number, time))
 	)
+
+	if p.config.IsPrague(blockNumber, time) {
+		ProcessParentBlockHash(block.ParentHash, vmenv, statedb)
```
HerbertJordan, line 78: It would be good to test that this feature is only enabled if Allegro/Prague is enabled. Any ideas how to do this effectively without a lot of extra work?

HerbertJordan (commented):
Looks good, but there are a few more things that we should consider when integrating this feature. Please have a look.

evmcore/state_processor.go
LuisPH3: I think so,in sonic I can deploy the history storage contract and query it, it should retrieve empty hashes.

tests/eip2935_test.go
LuisPH3: yes, account endowment could do the trick.

tests/eip2935_test.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/133/commits/4da7db62c0dceb4fedf6263bb9f3719729055b96

tests/eip2935_test.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/133/commits/4da7db62c0dceb4fedf6263bb9f3719729055b96

evmcore/state_processor.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/133/commits/6e55361a766fd76fe439b89b1b628a323f418b82

tests/eip2935_test.go
```diff
@@ -79,3 +83,65 @@ func TestEIP2935_HistoryContractIsNotDeployedBeforePrague(t *testing.T) {
...
+		receipt, err = net.Apply(func(opts *bind.TransactOpts) (*types.Transaction, error) {
+			return readHistoryStorageContract.ReadHistoryStorage(opts, new(big.Int))
+		})
+		require.NoError(t, err)
+		require.Equal(t, types.ReceiptStatusSuccessful, receipt.Status)
+		hashes = append(hashes, receipt.BlockHash)
+	}
+
+	// second loop queries block hashes of the blocks generated in the first loop
+	for i := range testIterations {
```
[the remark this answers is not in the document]
LuisPH3: This was mentioned already today. 
I created this ticket to handle it: https://github.com/0xsoniclabs/sonic-admin/issues/128

tests/contracts/read_history_storage/read_history_storage.sol
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/133/commits/a1fef7fe387e3a3aefa1874217974af4e9bca459

evmcore/state_processor.go
```diff
@@ -156,6 +161,27 @@ func ApplyTransactionWithEVM(msg *core.Message, config *params.ChainConfig, gp *
 	return receipt, err
 }
 
+// ProcessParentBlockHash stores the parent block hash in the history storage contract
```
facuMH, line 164: considering it is only used in `Process()`, does it need to be an exported (public) method ?

facuMH (approved):

evmcore/state_processor.go
LuisPH3: It will be used in (many) more places here: https://github.com/0xsoniclabs/sonic-admin/issues/128

PR #134, opened 2025-03-21
Add Multiplexer Provider

facuMH:
This PR introduces the `Multiplexer` type, which implements the `Provider` interface while also wrapping a number of existing `Provider`s and automatically iterating through them when they try fetch certificate and return a non `nil` error. It follows the Decorator Pattern, ensuring that multiplexing logic is separate from core provider implementations.

A multiplexer provider is also used in the `LightClient` and the `Config` struct is adapted to reflect this, containing now a list of providers and failing when this list in empty, or if any of the provided ones are invalid.

### DIFF
--- a/scc/light_client/multiplexer.go
+++ b/scc/light_client/multiplexer.go
@@ -73,7 +73,7 @@ func (m *multiplexer) getBlockCertificates(first idx.Block, maxResults uint64) (
 	})
 }
 
-// tryAll executes a function on each provider in sequence until one returns nil error.
+// tryAll executes a function on each provider in sequence until one returns a nil error.
 //
 // This function is a generic helper that iterates over a list of providers,
 // calling the given function on each. If any function call succeeds, it immediately
@@ -84,7 +84,7 @@ func (m *multiplexer) getBlockCertificates(first idx.Block, maxResults uint64) (
 //
 // Parameters:
 //   - ps: A slice of provider instances to be tried.
-//   - fn: A function that takes a provider and returns a result of type T and an error.
+//   - fn: A function that takes a provider and returns a result of type C and an error.
 //
 // Returns:
 //   - C: The result of the first successful function execution.
--- a/scc/light_client/multiplexer.go
+++ b/scc/light_client/multiplexer.go
@@ -1,0 +1,104 @@
+package light_client
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/0xsoniclabs/sonic/scc"
+	"github.com/0xsoniclabs/sonic/scc/cert"
+	"github.com/Fantom-foundation/lachesis-base/inter/idx"
+)
+
+// multiplexer is a provider that distributes requests across multiple providers.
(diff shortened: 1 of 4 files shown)


scc/light_client/multiplexer.go
```diff
@@ -0,0 +1,104 @@
...
+//   - fn: A function that takes a provider and returns a result of type T and an error.
+//
+// Returns:
+//   - C: The result of the first successful function execution.
+//   - error: Nil if at least one execution of fn returned without error.
+//     The joined error of all failed attempts if all attempts failed.
+func tryAll[C any](ps []provider, fn func(provider) (C, error)) (C, error) {
+	var errs []error
+	for _, p := range ps {
+		res, err := fn(p)
```
HerbertJordan, line 96: idea: a nice feature of the multiplexer would be to remember the last successful source and use this one first the next time; maybe in a follow-up PR.

scc/light_client/multiplexer.go
```diff
@@ -0,0 +1,104 @@
...
+// This function is a generic helper that iterates over a list of providers,
+// calling the given function on each. If any function call succeeds, it immediately
+// returns the result. If all calls fail, it aggregates the errors and returns them.
+//
+// Type Parameters:
+//   - C: The type of the result returned by the function.
+//
+// Parameters:
+//   - ps: A slice of provider instances to be tried.
+//   - fn: A function that takes a provider and returns a result of type T and an error.
```
HerbertJordan, line 87:
```suggestion
//   - fn: A function that takes a provider and returns a result of type C and an error.
```

scc/light_client/multiplexer.go
```diff
@@ -0,0 +1,104 @@
...
+//   - cert.BlockCertificate: The block certificates for the given block number
+//     and the following blocks.
+//   - error: Not nil if the provider failed to obtain the requested certificates.
+func (m *multiplexer) getBlockCertificates(first idx.Block, maxResults uint64) ([]cert.BlockCertificate, error) {
+	return tryAll(m.providers, func(p provider) ([]cert.BlockCertificate, error) {
+		return p.getBlockCertificates(first, maxResults)
+	})
+}
+
+// tryAll executes a function on each provider in sequence until one returns nil error.
```
HerbertJordan, line 76:
```suggestion
// tryAll executes a function on each provider in sequence until one returns a nil error.
```

HerbertJordan (dismissed):
Looks good, thanks!

scc/light_client/multiplexer.go
facuMH: thought about it as well, task created: https://github.com/0xsoniclabs/sonic-admin/issues/130

scc/light_client/multiplexer.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/134/commits/d00a2c14817dd3c53bbe1456e22df082f102cc7b

HerbertJordan (approved):

PR #138, opened 2025-03-25
Update mock library

facuMH:
This PR removes the deprecated `github.com/golang/mock` and replaces for the recommended `go.uber.org/mock/gomock`
Most mocks and tests in the project are already using `go.uber.org/mock/gomock`. But a few cases remain where the deprecated library is still being used. This PR removes the remaining cases of it.


HerbertJordan (approved):
Thanks for the dependency cleanup!

PR #140, opened 2025-03-25
Update ethereum dependency to v1.15.5

simonlechner:
The go ethereum dependency is updated to v1.15.5. 
Receipt blooms are only computed once per receipt and then merged, these changes were introduced in geth [here](https://github.com/ethereum/go-ethereum/pull/31129).


HerbertJordan (dismissed):
Looks good, thanks!

PR #142, opened 2025-03-25
network rule validation

kjezek:
This PR introduces a network rule verification infrastructure and adds support for validating rule changes on the network.

The rule verification code covers basic rule properties by constraining the domain of various rule parameters. The main goal of those constraints is to prevent an accidental network stall. These validation steps are to be considered a last line of defense to safe-guard the network lifelines. However, they are necessarily incomplete. Any change of network rules should be tested extensively before being applied on production networks.

To facilitate a safe roll-out, the validation step is ignored by default. Only a rule-update will enable this features. To prevent a network split, the validation features should only be updated once sufficient (ideally all) nodes on the network have been updated to support this feature.

This PR follows-up on https://github.com/Fantom-foundation/Sonic/pull/383

### DIFF
--- a/gossip/blockproc/drivermodule/driver_txs.go
+++ b/gossip/blockproc/drivermodule/driver_txs.go
@@ -23,7 +23,8 @@ import (
 )
 
 const (
-	maxAdvanceEpochs = 1 << 16
+	internalTransactionsGasLimit = opera.MinimumMaxBlockGas / 2
+	maxAdvanceEpochs             = 1 << 16
 )
 
 type DriverTxListenerModule struct{}
@@ -66,7 +67,7 @@ func InternalTxBuilder(statedb state.StateDB) func(calldata []byte, addr common.
 		if nonce == math.MaxUint64 {
 			nonce = statedb.GetNonce(common.Address{})
 		}
-		tx := types.NewTransaction(nonce, addr, common.Big0, 500_000_000, common.Big0, calldata)
+		tx := types.NewTransaction(nonce, addr, common.Big0, internalTransactionsGasLimit, common.Big0, calldata)
 		nonce++
 		return tx
 	}
--- a/gossip/blockproc/drivermodule/driver_txs.go
+++ b/gossip/blockproc/drivermodule/driver_txs.go
@@ -23,7 +23,7 @@ import (
 )
 
 const (
-	internalTransactionsGasLimit = opera.MinimumMaxBlockGas / 10
+	internalTransactionsGasLimit = 500_000_000
 	maxAdvanceEpochs             = 1 << 16
 )
 
--- a/gossip/blockproc/drivermodule/driver_txs.go
+++ b/gossip/blockproc/drivermodule/driver_txs.go
@@ -23,6 +23,7 @@ import (
 )
 
 const (
+    // internalTransactionsGasLimit is the gas limit for internal transactions like epoch sealing. This constant MUST NOT be changed, as doing so will cause a network fork.
 	internalTransactionsGasLimit = 500_000_000
 	maxAdvanceEpochs             = 1 << 16
 )
--- a/gossip/blockproc/drivermodule/driver_txs.go
+++ b/gossip/blockproc/drivermodule/driver_txs.go
@@ -23,7 +23,7 @@ import (
 )
 
 const (
-	internalTransactionsGasLimit = opera.MinimumMaxBlockGas / 2
+	internalTransactionsGasLimit = opera.MinimumMaxBlockGas / 10
 	maxAdvanceEpochs             = 1 << 16
 )
 
--- a/opera/marshal.go
+++ b/opera/marshal.go
@@ -2,15 +2,19 @@ package opera
 
 import "encoding/json"
 
-func UpdateRules(src Rules, diff []byte) (res Rules, err error) {
+func UpdateRules(src Rules, diff []byte) (Rules, error) {
 	changed := src.Copy()
-	err = json.Unmarshal(diff, &changed)
+	err := json.Unmarshal(diff, &changed)
 	if err != nil {
-		return src, err
+		return Rules{}, err
 	}
 	// protect readonly fields
-	res = changed
-	res.NetworkID = src.NetworkID
-	res.Name = src.Name
-	return
+	changed.NetworkID = src.NetworkID
+	changed.Name = src.Name
+
+	// check validity of the new rules
+	if err = changed.Validate(); err != nil {
+		return Rules{}, err
+	}
+	return changed, nil
 }
--- a/opera/marshal.go
+++ b/opera/marshal.go
@@ -13,8 +13,10 @@ func UpdateRules(src Rules, diff []byte) (Rules, error) {
 	changed.Name = src.Name
 
 	// check validity of the new rules
-	if err = changed.Validate(); err != nil {
-		return Rules{}, err
+	if changed.Upgrades.CheckRuleChanges {
+		if err = changed.Validate(); err != nil {
+			return Rules{}, err
+		}
 	}
 	return changed, nil
 }
--- a/opera/marshal.go
+++ b/opera/marshal.go
@@ -5,16 +5,16 @@ import "encoding/json"
 func UpdateRules(src Rules, diff []byte) (Rules, error) {
 	changed := src.Copy()
 	if err := json.Unmarshal(diff, &changed); err != nil {
-	if err != nil {
 		return Rules{}, err
 	}
+
 	// protect readonly fields
 	changed.NetworkID = src.NetworkID
 	changed.Name = src.Name
 
 	// check validity of the new rules
 	if changed.Upgrades.Allegro {
-		if err = changed.Validate(); err != nil {
+		if err := changed.Validate(); err != nil {
 			return Rules{}, err
 		}
 	}
--- a/opera/marshal.go
+++ b/opera/marshal.go
@@ -4,7 +4,7 @@ import "encoding/json"
 
 func UpdateRules(src Rules, diff []byte) (Rules, error) {
 	changed := src.Copy()
-	err := json.Unmarshal(diff, &changed)
+	if err := json.Unmarshal(diff, &changed); err != nil {
 	if err != nil {
 		return Rules{}, err
 	}
--- a/opera/marshal.go
+++ b/opera/marshal.go
@@ -13,7 +13,7 @@ func UpdateRules(src Rules, diff []byte) (Rules, error) {
 	changed.Name = src.Name
 
 	// check validity of the new rules
-	if changed.Upgrades.CheckRuleChanges {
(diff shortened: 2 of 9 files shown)


opera/validate.go
```diff
@@ -0,0 +1,221 @@
...
+	issues = append(issues, validateGasRules(rules.Gas))
+	issues = append(issues, validateGasPowerRules("Economy.ShortGasPower", rules.ShortGasPower))
+	issues = append(issues, validateGasPowerRules("Economy.LongGasPower", rules.LongGasPower))
+
+	return errors.Join(issues...)
+}
+
+const (
+	// upperBoundForRuleChangeGasCosts is a safe over-approximation of the gas costs of a rule change.
+	upperBoundForRuleChangeGasCosts = 1_000_000 // < TODO: verify this number
```
LuisPH3, line 142: Please add issue number and add it to the release plan. So that we do not forget this task

LuisPH3 (commented):
I need some extra time to read the rules and their testing, in the meanwhile I have some minor comments.

opera/validate_test.go
```diff
@@ -0,0 +1,517 @@
...
+		},
+		"London upgrade requires Berlin": {
+			upgrade: Upgrades{London: true},
+			issue:   "London upgrade requires Berlin",
+		},
+		"Sonic upgrade is required": {
+			upgrade: Upgrades{},
+			issue:   "Sonic upgrade is required",
+		},
+		"Allegro upgrade is required": {
```
LuisPH3, line 494: question: once this change is deployed, only allegro updates will be allowed? 

Could it be that deploy takes a little longer and a rules update is needed in the meanwhile ? (with sonic features)

opera/validate.go
```diff
@@ -0,0 +1,221 @@
...
+)
+
+func validateGasRules(rules GasRules) error {
+	var issues []error
+
+	if rules.MaxEventGas < upperBoundForRuleChangeGasCosts {
```
LuisPH3, line 148: If there is a upperBoundForRuleChangeGasCosts constant, I would expect values to be always smaller than this value.
Should it be lowerBound instead?

opera/validate.go
LuisPH3: I am having some trouble to understand how this relates to other fields, like: rules.MaxEventGas

LuisPH3 (commented):
Unfortunately, I am not familiar enough to confirm this changes, please check my comments on your on criteria.  
One comment that I think is important is whereas we want to force allegro updates, once this changes go live.

opera/marshal.go
```diff
@@ -2,15 +2,21 @@ package opera
...
-func UpdateRules(src Rules, diff []byte) (res Rules, err error) {
+func UpdateRules(src Rules, diff []byte) (Rules, error) {
 	changed := src.Copy()
-	err = json.Unmarshal(diff, &changed)
+	err := json.Unmarshal(diff, &changed)
 	if err != nil {
-		return src, err
+		return Rules{}, err
 	}
 	// protect readonly fields
```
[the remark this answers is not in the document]
kjezek: It refers to lines: 
```

	changed.NetworkID = src.NetworkID
	changed.Name = src.Name

```
i.e. I think the comment is at the right place.

opera/validate_test.go
kjezek: Good point. The idea was that we never want to disable a hardfork, but since this code also validates generation of genesis, we should allow for generating a genesis data without Allegro. 

All in all, only transition from False to True should be prohibited.

kjezek: actually it should be ok, because the whole validation is enabled only when Allegro is `true`

kjezek: I noticed the check is missing for genesis generation for block zero - will add it, though it is probably not needed in real life as the network is already running, but it is needed at least for testing, perhaps for running a new fakenet, ot testnet, etc.

opera/validate.go
kjezek: I think this is a boundary for a rule change event only - i.e. an event that contains only one transaction that changes the network rule. An event can be bigger, but its minimal possible size must be to allow for updating network rules.

LuisPH3: I get it now. Next generation rules shall have a `MaxAventGas` larger than the cost of executing a rule change. To allow later changes. It is a complex concept.

LuisPH3 (dismissed):

gossip/blockproc/drivermodule/driver_txs.go
```diff
@@ -23,7 +23,8 @@ import (
 )
 
 const (
-	maxAdvanceEpochs = 1 << 16
+	internalTransactionsGasLimit = opera.MinimumMaxBlockGas / 10
```
HerbertJordan, line 26: This might already have been present in my original prototype, but thinking about it, this seems to be a not-so-great idea.

It is very important that the costs of internal transactions is a constant. If we would change this, we break the epoch sealing. So I suggest to put a fixed value of `500_000_000` here and have a static check that this is at most 10% of the `opera.MinimumMaxBlockGas` constant in some init code that panics if this is not the case.

My worry is that somebody may change `opera.MinimumMaxBlockGas` without being aware of this changing the internal transaction gas limit, and thereby indirectly breaking compatibility. Thus, actually this imposes a limit on the `opera.MinimumMaxBlockGas` constant, and the property of it being at least 10x larger than this constant should be checked wherever it is defined.

opera/rules.go
```diff
@@ -27,8 +28,9 @@ const (
 	sonicBit               = 1 << 3
 	allegroBit             = 1 << 4
 
-	defaultMaxBlockGas          = 1_000_000_000
-	defaultTargetGasRate        = 15_000_000 // 15 MGas/s
+	MinimumMaxBlockGas          = 5_000_000_000 // < must be large enough to allow internal transactions to seal blocks
```
HerbertJordan, line 31: As we are including these constants into the rule verification, these must never be altered again. If we would, potential future rule updates may be interpreted differently by different versions of the client. We should document this here, with a big block of comments including a "WARNING" sign.

We should be clear on which constants are included in validity tests, and are thus not be modified any more, and which constants may be modified as they are not part of the validity test.

opera/rules.go
```diff
@@ -420,9 +423,23 @@ func DefaultGasPowerRules() GasPowerRules {
 func (r Rules) Copy() Rules {
 	cp := r
 	cp.Economy.MinGasPrice = new(big.Int).Set(r.Economy.MinGasPrice)
+	cp.Economy.MinBaseFee = new(big.Int).Set(r.Economy.MinBaseFee)
```
HerbertJordan, line 426: This change is interesting. Basically, this means that before it, the `BaseFee` is not properly copied during rule updates, which could lead to unintended results. We should investigate the implications and warn the operations team.

opera/rules.go
```diff
@@ -420,9 +423,23 @@ func DefaultGasPowerRules() GasPowerRules {
...
 	return cp
 }
 
+// Validate checks the rules for consistency and safety. Rules are considered safe if
+// they do not risk stalling the network or preventing future rule updates.
+//
+// Note: the validation is very liberal to allow a maximum flexibility in the rules.
+// The merely check for the most critical configuration errors that may lead to network
```
HerbertJordan, line 434:
```suggestion
// It merely checks for the most critical configuration errors that may lead to network
```

opera/rules.go
```diff
@@ -27,8 +28,9 @@ const (
 	sonicBit               = 1 << 3
 	allegroBit             = 1 << 4
 
-	defaultMaxBlockGas          = 1_000_000_000
-	defaultTargetGasRate        = 15_000_000 // 15 MGas/s
+	MinimumMaxBlockGas          = 5_000_000_000 // < must be large enough to allow internal transactions to seal blocks
+	MaximumMaxBlockGas          = math.MaxInt64 // < should fit into 64-bit signed integers to avoid parsing errors in third-party libraries
```
HerbertJordan, line 32: nit: these constants should be moved to the `validate.go` file, wdyt?

tests/integration_test_net.go
```diff
@@ -188,7 +188,7 @@ func StartIntegrationTestNetWithJsonGenesis(
 	jsonGenesis.Accounts = append(jsonGenesis.Accounts, effectiveOptions.Accounts...)
 
 	// Speed up the block generation time to reduce test time.
-	jsonGenesis.Rules.Emitter.Interval = inter.Timestamp(time.Millisecond)
+	jsonGenesis.Rules.Emitter.Interval = inter.Timestamp(100 * time.Millisecond)
```
HerbertJordan, line 191: This seems unrelated and should be removed. It would slow down our integration tests.

HerbertJordan (changes requested):
Thanks for building this, looks good. My main concern is to document the "constness" of some of the constants in the code. We need to make sure that developers understand that some of those are not only constant for a binary build but over time. This should be documented accordingly, as described in the comments. Please have a look, let me know what you think.

opera/rules.go
kjezek: right, it was copied only as a pointer - i.e. the value could be potentially modified from the original object

opera/rules.go
kjezek: `MinimumMaxBlockGas` is used in both files `rules.go` and `validate.go`, i.e. not sure where it better belongs

tests/integration_test_net.go
kjezek: This is related because we introduced a minimum `Interval` 100ms, and when the interval is lower, the integration tests do not start due to not passing the validation. 

100ms minimum value was requested by @jmpike , I do not know if this is a hard constraint.

opera/rules.go
kjezek: Perhaps we should duplicate these constants in `validate.go`, where they would became non-updatable in this file, according to the comment above.

opera/rules.go
kjezek: ok, added warning to the validation.go header. I think this warning holds for the whole validation process, not only these constants.  Also these constants were duplicated and moved to validation.go to make them "private" for this file.

tests/integration_test_net.go
kjezek: ok, this change removed as well as the minimal threshold for `Interval`

opera/rules.go
HerbertJordan: This might be the reason why base-fee changes are effective immediately. Without this, it may take until the end of the epoch for those changes to become effective.

HerbertJordan: We should add a unit test for `Copy` that makes sure all fields are deep-copied.

kjezek: done in another PR

opera/validate.go
kjezek: I tried to test this property here: 
https://github.com/0xsoniclabs/sonic/pull/204
let me know what do you think

HerbertJordan (approved):

PR #144, opened 2025-03-26
Simplify BLS on-chain test

HerbertJordan:
A set of minor improvements in the implementation of the on-chain BLS verification test, mostly concerning the integration with the integration net infrastructure.


jenikd (approved):
Thanks for these changes

PR #145, opened 2025-03-26
Add `GetBalance` to `LightClient`

facuMH:
This PR adds the method `GetBalance` to the `LightClient` enabling it to ask for the balance of a given account, verifying it with the known state root certificate. 
This PR:
- adds the method `getAccountProof` which given an account, returns the witness proof for it, using the corresponding provider method.
- adds the method `GetBalance` which given an address, returns it's balance or an error. uses `getAccoutProof` to get a proof for the state's current head and the given address (of an account), and then validate it using the state's root hash.

### DIFF
--- a/scc/light_client/light_client.go
+++ b/scc/light_client/light_client.go
@@ -69,9 +69,12 @@ func (c *LightClient) Sync() (idx.Block, error) {
 }
 
 // getAccountProof retrieves and verifies the proof for the given address.
-// Returns an error if synchronization fails, if the proof cannot be obtained.
+// Returns an error if the light client has not been synced or if the proof cannot be obtained.
 func (c *LightClient) getAccountProof(address common.Address) (carmen.WitnessProof, error) {
-	proof, err := c.provider.getAccountProof(address, LatestBlock)
+	if !c.state.hasSynced {
+		return nil, fmt.Errorf("light client has not yet synced")
+	}
+	proof, err := c.provider.getAccountProof(address, c.state.headNumber)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get account proof: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


scc/light_client/light_client.go
```diff
@@ -64,3 +67,78 @@ func (c *LightClient) Close() {
...
+// getAccountProof retrieves and verifies the proof for the given address.
+// It first ensures the client is synchronized before querying for the proof.
+// Returns an error if synchronization fails, if the proof cannot be obtained.
+func (c *LightClient) getAccountProof(address common.Address) (carmen.WitnessProof, error) {
+	// always sync before querying
+	_, err := c.Sync()
+	if err != nil {
+		return nil, fmt.Errorf("failed to sync: %w", err)
+	}
+	proof, err := c.provider.getAccountProof(address, LatestBlock)
```
HerbertJordan, line 80: This may simply fail because "LatestBlock" has moved since the completion of the `Sync` call. In fact, it is quite likely that the certification chain is lagging behind the change by a few blocks. So this fetch should not ask for the "LatestBlock" but the block you just synced to.

scc/light_client/light_client.go
```diff
@@ -64,3 +67,78 @@ func (c *LightClient) Close() {
 func (c *LightClient) Sync() (idx.Block, error) {
 	return c.state.sync(c.provider)
 }
+
+// getAccountProof retrieves and verifies the proof for the given address.
+// It first ensures the client is synchronized before querying for the proof.
+// Returns an error if synchronization fails, if the proof cannot be obtained.
+func (c *LightClient) getAccountProof(address common.Address) (carmen.WitnessProof, error) {
+	// always sync before querying
+	_, err := c.Sync()
```
HerbertJordan, line 76: One could argue that it should be the users choice to sync or not to sync. After all, it is offered as an extra function.

scc/light_client/light_client.go
facuMH: right. Here it should use the latest block known to the light client state.

scc/light_client/light_client.go
```diff
@@ -64,3 +67,45 @@ func (c *LightClient) Close() {
...
+}
+
+// GetBalance returns the balance of the given address.
+// It returns an error if the light client has not been synced,
+// the balance could not be proven or there was any error
+// in getting or verifying the proof.
+func (c *LightClient) GetBalance(address common.Address) (*uint256.Int, error) {
+	if !c.state.hasSynced {
+		return nil, fmt.Errorf("light client has not yet synced")
+	}
```
HerbertJordan, line 94: nit: redundant, since it is also checked inside the `getAccountProof` call following right after.

HerbertJordan (approved):

scc/light_client/light_client.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/145/commits/7d809671c18e57960b435da8d81a12baf2d0737c

HerbertJordan (approved):

PR #146, opened 2025-03-26
Modify RPC callbacks to execute EIP-2935 History Storage Contract.

LuisPH3:
Issue https://github.com/0xsoniclabs/sonic-admin/issues/128 describes a list of RPC methods which can execute a transaction in the RPC.
All these methods are calling one of the following internal routines: 
- `DoCall`, executes one transaction in a read-only mode. It is used by `DoEstimateGas` which ends up being used in many rpc methods.
- `StateAtTransaction`, executes a subset of one block transactions and returns the intermediate state. Used in debug and tracing methods.
- trace: `replayBlock`, similar to the previous, executes a subset of transactions included in a block. Used by other trace methods.
- `AccessList`, this is a tracing tool that helps pre-filling the access list for a transaction. This function exists 1:1 in [geth](https://github.com/ethereum/go-ethereum/blob/a775e68421595d9c3807e68cce7ff2037991a781/internal/ethapi/api.go#L1139), but it hasn't been upgraded with the history storage semantics (it should happen [here](https://github.com/ethereum/go-ethereum/blob/a775e68421595d9c3807e68cce7ff2037991a781/internal/ethapi/api.go#L1224C1-L1224C82)). 

Therefore, only  `DoCall`,  `StateAtTransaction`, and `replayBlock` have been instrumented with the new semantics. 

The methodology has been to locate all calls to `core.ApplyMessage` from the RPC, as this is the entry point to the processor. 

Unit testing checks that the side effect state calls, happening as a result of the invocation of the History Storage contract, happen only when `Allegro` is enabled. 
This test does not check that the RPC state access is read-only, as this is in a much larger scope (that applies to the complete RPC usage of the state by any callback)

### DIFF
--- a/ethapi/api_test.go
+++ b/ethapi/api_test.go
@@ -17,6 +17,8 @@ import (
 	"github.com/0xsoniclabs/sonic/inter/state"
 	"github.com/0xsoniclabs/sonic/opera"
 	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/consensus"
+	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/eth/tracers"
 	"github.com/holiman/uint256"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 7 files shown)


ethapi/api_test.go
```diff
@@ -596,6 +598,325 @@ func TestTransactionJSONSerialization(t *testing.T) {
...
+				mockState.EXPECT().SubBalance(sender, gomock.Any(), gomock.Any())
+
+				mockState.EXPECT().Prepare(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any())
+				mockState.EXPECT().SetNonce(sender, uint64(2), gomock.Any())
+				mockState.EXPECT().GetCode(recipient).Return([]byte{})
+				mockState.EXPECT().Snapshot()
+				mockState.EXPECT().Exist(recipient)
+				mockState.EXPECT().GetRefund().Times(2)
+				mockState.EXPECT().EndTransaction().Times(2)
+				mockState.EXPECT().TxIndex()
```
facuMH, line 814: this could be abstracted as `expectedTraceReplayBlockPostTransaction` or something similar and used in both `sonic` and `allegro` cases. wdy ?

ethapi/api_test.go
```diff
@@ -596,6 +598,325 @@ func TestTransactionJSONSerialization(t *testing.T) {
...
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+
+			ctrl := gomock.NewController(t)
+
+			blockOrHash := rpc.BlockNumberOrHashWithNumber(0)
+
+			header := evmcore.EvmHeader{
+				// return any block but 0, 0 is genesis and has special semantics
```
facuMH, line 828: for the same reason of this comment, shouldn't `blockOrHash` also be different than `0` ?

ethapi/api_test.go
```diff
@@ -596,6 +598,325 @@ func TestTransactionJSONSerialization(t *testing.T) {
...
+			[]opera.UpgradeHeight{
+				{Upgrades: opera.SonicFeatures.ToUpgrades(), Height: 0},
+			})
+	} else if features == opera.AllegroFeatures {
+		chainConfig = opera.MainNetRules().EvmChainConfig(
+			[]opera.UpgradeHeight{
+				{Upgrades: opera.SonicFeatures.ToUpgrades(), Height: 0},
+				{Upgrades: opera.AllegroFeatures.ToUpgrades(), Height: 1},
+			})
+	}
```
facuMH, line 887: nit: silently returning config with zero value of `Upgrade` seems like not ideal for testing of future upgrades. What about `else{ fmt.Printf("unknown feature set %v", feature) }` ? or passing `t` as a parameter and using `t.Logf`?

HerbertJordan (dismissed):
Looks good, thanks!

ethapi/api_test.go
LuisPH3: yes, it may look cleaner.
I was getting a little tired of this monster test, i could not decide what deserves a name and what does not.

ethapi/api_test.go
LuisPH3: yes

ethapi/api_test.go
```diff
@@ -596,6 +598,325 @@ func TestTransactionJSONSerialization(t *testing.T) {
...
+			senderKey)
+
+		block.Transactions = []*types.Transaction{tx1, tx2}
+		txHash := tx2.Hash()
+		_, err := api.replayBlock(t.Context(), block, &txHash, &[]hexutil.Uint{hexutil.Uint(0)})
+		require.NoError(t, err)
+	}
+
+	expectedCallsFromTxCall := func(mockState *state.MockStateDB) {
+		mockState.EXPECT().Prepare(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any())
```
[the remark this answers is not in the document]
LuisPH3: I think that I will keep it in mind for the next time.

ethapi/api_test.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/146/commits/b010ad5402d118a4b0eeb32f833d0397e3d3b957

ethapi/api_test.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/146/commits/6c952c26bc125322cbfa1679dca2bc962a9a7d0a

ethapi/api_test.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/146/commits/37a882325e04238128a999cfa3b2a4af62f804d8

facuMH (approved):

jenikd (approved):

PR #148, opened 2025-03-27
Reject gapped tx from delegated account

facuMH:
This PR updates the transaction validation logic to ensure proper ordering of transactions with authorization.

The previous check rejected a transaction if all the following conditions match:
- The sender had a delegation or pending delegation.
- There was at least one queued or pending transaction from the same sender.
- None of those transactions had the same nonce as the incoming transaction.

This approach led to an issue where:
- If a transaction with a higher nonce than expected was accepted first, and later a transaction with the expected nonce arrived, the latter would be rejected.
- This happened due to the "at most one in-flight transaction" policy introduced in https://github.com/0xsoniclabs/sonic/pull/74.
- As a result, the first transaction (with the higher nonce) would be stuck.

*New Behavior*
The updated check ensures that delegated transactions are accepted in the correct execution order.
This prevents transactions from getting stuck while maintaining execution order integrity.

This PR addresses https://github.com/0xsoniclabs/sonic-admin/issues/137

### DIFF
--- a/evmcore/tx_pool.go
+++ b/evmcore/tx_pool.go
@@ -94,6 +94,10 @@ var (
 	// than some meaningful limit a user might use. This is not a consensus error
 	// making the transaction invalid, rather a DOS protection.
 	ErrOversizedData = errors.New("oversized data")
+
+	// ErrOutOfOrderTxFromDelegated is returned when the transaction with gapped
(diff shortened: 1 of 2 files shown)


LuisPH3 (approved):
The PR includes the changes merged in go-ethereum here https://github.com/ethereum/go-ethereum/pull/31430/files

PR #110, opened 2025-03-12
Add feature changes and test for EIP-7623

LuisPH3:
This PR introduces calldata cost changes https://eips.ethereum.org/EIPS/eip-7623.

The transaction pool validation is extended to reject transactions not having enough Gas according to new formulas.
The test is introduced to check calldata costs, and its interaction with the sonic 10% cost of the unused gas rule.

Additionally, Init code size limit check has been added to the transaction pool validation to reject transactions which would otherwise be accepted to be ignored in the processor.

PR #154, opened 2025-03-31
Free standing `validateTx` and unit test

facuMH:
This work started from the lack of unit testing of this function to check enforcement of new EIP features and reject transactions that do not comply with them (namely [7702](https://eips.ethereum.org/EIPS/eip-7702) intrinsic gas check and [7623](
https://eips.ethereum.org/EIPS/eip-7623) floor data cost). Also during the development of these feature it was noticed that a check was missing reported in [#110](https://github.com/0xsoniclabs/sonic/pull/110). That kind of checks and EIP changes should be added to `validateTx` and it's testing.  
To facilitate unit testing of `validateTx` without needing to instantiate a whole `TxPool`, the function is moved as an independent function (inspired by [geth's validation function](https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L61)) which takes a set of `validationOptions` and uses  those to judge whether a transaction is valid or not.

The objective of this PR is not to modify the behaviour or redefine any semantics, solely to make `validateTx` as independent from `TxPool` as possible and ensure the former rejects transactions under the specific conditions.


evmcore/validation.go
LuisPH3, line 1: Idea for refactor:
reorganize the validate function by cost of organizing validation.
This would check all static and policy checks first, leaving the last 3 check groups to be more expensive ones:
- check signature (retrieve sender)
- check local/non-local policies (need sender) 
- check sender balance and nonce (in any order) 
Wdyt?

evmcore/validation_test.go
```diff
@@ -0,0 +1,620 @@
...
+		eip7702:         true,
+		shanghai:        true,
+		currentMaxGas:   100_000,
+		currentGasPrice: big.NewInt(1),
+		currentBaseFee:  big.NewInt(1),
+		isLocal:         true,
+	}
+}
+
+func TestValidation_validateTx_Rejects(t *testing.T) {
```
LuisPH3, line 110: This test function does not contribute to the file:
its purpose is to make some variables common to the listed files.

It could be simplified by rewriting all `test_*` function to be proper `Test*(t testing.T)`, remove `t.Run` from them and initialize the variables needed inside.

evmcore/validation_test.go
```diff
@@ -0,0 +1,620 @@
...
+	for _, test := range tests {
+		for _, tx := range allTxTypes() {
+			test(tx)
+		}
+	}
+}
+
+// TestValidation_validateTx_RejectsOversizedData tests that validateTx
+// rejects transactions with oversized data fields.
+func test_OversizedData(t *testing.T, tx types.TxData, oversizedData []byte) {
```
LuisPH3, line 141: See previous comment, please rewrite as:
`TestValidateTx_TransactionIsRejectedWithOversizedData(t *testing.T)`

evmcore/validation_test.go
```diff
@@ -0,0 +1,620 @@
...
+			require.NoError(t, err)
+		})
+	}
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// Helper functions for testing.
+
+// allTxTypes returns a list of all transaction types for testing.
+func allTxTypes() []types.TxData {
```
LuisPH3, line 448:
```suggestion
func getTxsFromAllTypes() []types.TxData {
```

evmcore/validation_test.go
```diff
@@ -0,0 +1,620 @@
...
+		&types.AccessListTx{},
+		&types.DynamicFeeTx{},
+		makeBlobTx(nil, nil),
+		&types.SetCodeTx{AuthList: []types.SetCodeAuthorization{{}}},
+	}
+}
+
+// testSignTx generates a new key, signs the transaction with it, and returns
+// the signer, address, and signed transaction.
+func testSignTx(t *testing.T, tx types.TxData) (types.Signer, common.Address, *types.Transaction) {
```
LuisPH3, line 460:
```suggestion
func signTx(t *testing.T, tx types.TxData) (types.Signer, common.Address, *types.Transaction) {
```

The reason is that one may think that this function is testing signing

evmcore/validation_test.go
```diff
@@ -0,0 +1,620 @@
...
+	address := crypto.PubkeyToAddress(key.PublicKey)
+	require.NoError(t, err)
+	signer := types.NewPragueSigner(big.NewInt(1))
+	signedTx, err := types.SignTx(types.NewTx(tx), signer, key)
+	require.NoError(t, err)
+	return signer, address, signedTx
+}
+
+// txTypeName returns the name of the transaction type for logging purposes.
+func txTypeName(tx types.TxData) string {
```
LuisPH3, line 471:
```suggestion
func getTxTypeName(tx types.TxData) string {
```

evmcore/validation_test.go
```diff
@@ -0,0 +1,620 @@
...
+
+// Helper function to set the "Value" field of a transaction to a negative value.
+// for blob and setCode transactions, it sets the value to zero since they use uint256.
+func negativeValue(tt *testing.T, tx types.TxData) {
+	setTxField(tt, tx, "Value", big.NewInt(-1), big.NewInt(-1), big.NewInt(-1),
+		uint256.NewInt(0), uint256.NewInt(0))
+}
+
+// setTxField sets a field of a transaction to a specific value.
+func setTxField(tt *testing.T, tx types.TxData, field string, value ...any) {
```
LuisPH3, line 569: I would rather not have two concurrent mechanisms to do similar things: 
some field modifiers contain a switch, some call this routine,

This routine contains an obscure variadic argument that may not be clear to the user without reading the code. Explaining it may be long and tedious.

I see two options here:
- Beef up the `setTxField` function to be the one and only method to modify txs. (use reflection for type magic, return err on impossible modifications, have an easy-to-read interface without obscure variadic semantics)
- Remove it and rewrite modifiers with switches. 

Although the first choice seems more powerful, it hardly justifies development time.

evmcore/validation_test.go
```diff
@@ -0,0 +1,620 @@
...
+		}
+	default:
+		tt.Fatalf("unexpected transaction type: %T", tx)
+	}
+}
+
+// assignField assigns a value to a field of a transaction using reflection.
+// It checks if the field is valid and can be set before assigning the value.
+// If the field is not valid or cannot be set, it fails the test.
+func assignField(tt *testing.T, tx any, field string, value any) {
```
LuisPH3, line 593: Although this seems pretty cool. It leverages some basic functionality of the type system and converts symbols into runtime entities. Compile time errors into runtime errors. 
As a rule of thumb, compile time errors are always better.

LuisPH3 (changes requested):
The PR migrates transaction validation to be a freestanding function and thoroughly tests it. 
- Validation is 1:1 to the existing one
- Testing seems complete

Nevertheless, these changes can be improved for better maintainability. Please check comments:

evmcore/validation.go
facuMH: I agree that this should be done, but in a follow-up PR. https://github.com/0xsoniclabs/sonic-admin/issues/147

evmcore/validation_test.go
facuMH: what do you think about `signTxForTest` ? considering there is already a function called `signTx` and it is used already in this package, I would like to keep the names as different/explicit as possible to prevent confusion.

evmcore/validation.go
```diff
@@ -0,0 +1,187 @@
...
+		if !opt.eip4844 {
+			return ErrTxTypeNotSupported
+		}
+		// For now, Sonic only supports Blob transactions without blob data.
+		if len(tx.BlobHashes()) > 0 ||
+			(tx.BlobTxSidecar() != nil && len(tx.BlobTxSidecar().BlobHashes()) > 0) {
+			return ErrTxTypeNotSupported
+		}
+	}
+	if tx.Type() == types.SetCodeTxType {
```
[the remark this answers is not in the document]
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/dd18ca01dd60f96baca078b67b94229bab42ab80

evmcore/validation_test.go
facuMH: As requested, this Test was dropped and instead each the tx generation and loop is done in each test. https://github.com/0xsoniclabs/sonic/pull/154/commits/dd18ca01dd60f96baca078b67b94229bab42ab80

evmcore/validation_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/dd18ca01dd60f96baca078b67b94229bab42ab80

evmcore/validation_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/dd18ca01dd60f96baca078b67b94229bab42ab80

evmcore/validation_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/dd18ca01dd60f96baca078b67b94229bab42ab80

evmcore/validation_test.go
facuMH: after some offline discussion, I will remove the helper function and simply rewrite the modifiers using switches

evmcore/validation_test.go
facuMH: function removed.

facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/6fed21c6d712aa29ce22c17d6d2ad88d1efd1f6f

evmcore/validation_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/6fed21c6d712aa29ce22c17d6d2ad88d1efd1f6f

LuisPH3 (commented):
I am thinking: how can we read the test cases and be sure that we do not skip any important test? 

I would like that we put some time, figuring out a code structure, so that it is easier to identify each problem:
For example, one test per field in the transaction, then group all the possible issues this field could produce. That would be easy to read, don't you think?

evmcore/tx_pool.go
```diff
@@ -676,129 +673,25 @@ func (pool *TxPool) local() map[common.Address]types.Transactions {
...
+		shanghai:       pool.shanghai,
+		eip1559:        pool.eip1559,
+		eip2718:        pool.eip2718,
+		eip4844:        pool.eip4844,
+		eip7623:        pool.eip7623,
+		eip7702:        pool.eip7702,
+		currentState:   pool.currentState,
+		currentMaxGas:  pool.currentMaxGas,
+		currentBaseFee: pool.chain.GetCurrentBaseFee(),
+		minTip:         pool.gasPrice,
```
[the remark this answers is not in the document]
facuMH: https://github.com/0xsoniclabs/sonic-admin/issues/33

evmcore/validation_test.go
```diff
@@ -0,0 +1,634 @@
...
+	for name, tx := range getTxsFromAllTypes() {
+		t.Run(name, func(t *testing.T) {
+			if isBlobOrSetCode(tx) {
+				t.Skip("blob and setCode transactions cannot be used as create")
+			}
+			setData(t, types.TxData(tx), maxInitCode)
+			setToNil(t, tx)
+			err := validateTx(types.NewTx(tx), types.NewPragueSigner(big.NewInt(1)),
+				testTransactionsOption())
+			require.ErrorContains(t, err, ErrMaxInitCodeSizeExceeded.Error())
```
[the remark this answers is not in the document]
facuMH: changed to ErrorIs

evmcore/validation_test.go
```diff
@@ -0,0 +1,634 @@
...
+	case *types.SetCodeTx:
+		tx.To = common.Address{}
+	default:
+		t.Fatalf("unexpected transaction type: %T", tx)
+	}
+}
+
+// Helper function to set the "Value" field of a transaction to a negative value.
+// for blob and setCode transactions, it sets the value to zero since they use uint256.
+func negativeValue(t *testing.T, tx types.TxData) {
```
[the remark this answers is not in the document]
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/a9c352a9d0b5709f4552eed1be3b7ef636765de5

evmcore/validation_test.go
```diff
@@ -0,0 +1,634 @@
...
+
+func TestValidation_validateTx_RejectsTooLargeGasTip(t *testing.T) {
+	extremelyLargeN := new(big.Int).Lsh(big.NewInt(1), 256)
+	for name, tx := range getTxsFromAllTypes() {
+		t.Run(name, func(t *testing.T) {
+			if isBlobOrSetCode(tx) {
+				t.Skip("blob and setCode transactions cannot have gas larger than u256")
+			}
+			// set gas fee cap too large
+			setGasTipCap(t, tx, extremelyLargeN)
```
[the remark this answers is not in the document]
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/a9c352a9d0b5709f4552eed1be3b7ef636765de5

evmcore/validation_test.go
```diff
@@ -0,0 +1,634 @@
...
+				testTransactionsOption())
+
+			// Since for legacy and access list transactions, the gas price is used as the gas tip cap,
+			// it would be rejected as well in the gas fee cap check.
+			// For blob and setCode transactions, the gas tip cap is of type uint256
+			// so it can never have a bit length larger than 256.
+			if isLegacyOrAccessList(tx) {
+				require.ErrorIs(t, err, ErrFeeCapVeryHigh)
+			}
+			if _, ok := tx.(*types.DynamicFeeTx); ok {
```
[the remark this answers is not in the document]
facuMH: added an else fatal case, if it is not a dynamic fee tx, and as commented in the previous comment, all other know types are skipped.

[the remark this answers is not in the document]
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/a9c352a9d0b5709f4552eed1be3b7ef636765de5

evmcore/validation_test.go
```diff
@@ -0,0 +1,634 @@
...
+			setGas(t, tx, 1)
+
+			// --- needed for execution up to relevant check ---
+			// set tx for execution
+			setGasFeeCap(t, tx, opt.minTip)
+			// sign txs with sender
+			signer, address, signedTx := signTxForTest(t, tx)
+			// ---
+
+			// setup enough balance
```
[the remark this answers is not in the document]
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/a9c352a9d0b5709f4552eed1be3b7ef636765de5

evmcore/validation_test.go
```diff
@@ -0,0 +1,634 @@
...
+func setToNil(t *testing.T, tx types.TxData) {
+	switch tx := tx.(type) {
+	case *types.LegacyTx:
+		tx.To = nil
+	case *types.AccessListTx:
+		tx.To = nil
+	case *types.DynamicFeeTx:
+		tx.To = nil
+	case *types.BlobTx:
+		tx.To = common.Address{}
```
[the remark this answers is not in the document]
facuMH: marked as fatal https://github.com/0xsoniclabs/sonic/pull/154/commits/a9c352a9d0b5709f4552eed1be3b7ef636765de5

evmcore/validation_test.go
```diff
@@ -0,0 +1,634 @@
...
+	case *types.SetCodeTx:
+		tx.To = common.Address{}
+	default:
+		t.Fatalf("unexpected transaction type: %T", tx)
+	}
+}
+
+// Helper function to set the "Value" field of a transaction to a negative value.
```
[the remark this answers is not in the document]
facuMH: function name added at the beginning of the doc. https://github.com/0xsoniclabs/sonic/pull/154/commits/a9c352a9d0b5709f4552eed1be3b7ef636765de5

evmcore/validation_test.go
```diff
@@ -0,0 +1,634 @@
...
+	case *types.BlobTx:
+		tx.Gas = gas
+	case *types.SetCodeTx:
+		tx.Gas = gas
+	default:
+		t.Fatalf("unexpected transaction type: %T", tx)
+	}
+}
+
+// Helper function to add oversized data to a transaction.
```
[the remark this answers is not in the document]
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/a9c352a9d0b5709f4552eed1be3b7ef636765de5

evmcore/validation_test.go
```diff
@@ -0,0 +1,634 @@
...
+	case *types.BlobTx:
+		tx.Data = data
+	case *types.SetCodeTx:
+		tx.Data = data
+	default:
+		t.Fatalf("unexpected transaction type: %T", tx)
+	}
+}
+
+// Helper function to set the "To" field of a transaction to nil.
```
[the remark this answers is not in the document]
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/a9c352a9d0b5709f4552eed1be3b7ef636765de5

evmcore/validation_test.go
```diff
@@ -0,0 +1,634 @@
...
+	bigIntToU256 := func(bigInt *big.Int) *uint256.Int {
+		u256, overflow := uint256.FromBig(bigInt)
+		if overflow {
+			t.Fatalf("overflowed converting gasFeeCap to uint256")
+		}
+		return u256
+	}
+	switch tx := tx.(type) {
+	case *types.LegacyTx:
+		tx.GasPrice = gasTipCap
```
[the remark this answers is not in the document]
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/a9c352a9d0b5709f4552eed1be3b7ef636765de5 changed to fatal.

evmcore/validation_test.go
```diff
@@ -0,0 +1,634 @@
...
+			opt.currentState = testDb
+
+			// validate transaction
+			err := validateTx(signedTx, signer, opt)
+			require.ErrorIs(t, err, ErrIntrinsicGas)
+		})
+	}
+}
+
+func TestValidation_validateTx_RejectsCannotAffordFloorDataGas(t *testing.T) {
```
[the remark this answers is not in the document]
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/a9c352a9d0b5709f4552eed1be3b7ef636765de5 renamed

evmcore/validation_test.go
```diff
@@ -0,0 +1,634 @@
...
+			opt.currentState = testDb
+
+			// validate transaction
+			err := validateTx(signedTx, signer, opt)
+			require.ErrorIs(t, err, ErrInsufficientFunds)
+		})
+	}
+}
+
+func TestValidation_validateTx_RejectsCannotAffordIntrinsicGas(t *testing.T) {
```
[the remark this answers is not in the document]
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/a9c352a9d0b5709f4552eed1be3b7ef636765de5 renamed

evmcore/validation_test.go
```diff
@@ -0,0 +1,634 @@
...
+		tx.GasTipCap = bigIntToU256(gasTipCap)
+	default:
+		t.Fatalf("unexpected transaction type: %T", tx)
+	}
+}
+
+// setGasFeeCap sets the gas fee cap for a transaction. For legacy and access list
+// transactions, it sets the gas price.
+func setGasFeeCap(t *testing.T, tx types.TxData, gasFeeCap *big.Int) {
+	bigIntToU256 := func(bigInt *big.Int) *uint256.Int {
```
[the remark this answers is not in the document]
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/813385ca48fcdc74a8b76365f806d78cddc8b94b

evmcore/tx_pool.go
```diff
@@ -676,129 +673,24 @@ func (pool *TxPool) local() map[common.Address]types.Transactions {
...
-		pool.istanbul,  // is eip-2028 (transactional data gas cost reduction)
-		pool.shanghai,  // is eip-3860 (limit and meter init-code )
-	)
+	opts := validationOptions{
+		istanbul:       pool.istanbul,
+		shanghai:       pool.shanghai,
+		london:         pool.eip1559,
+		berlin:         pool.eip2718,
+		cancun:         pool.eip4844,
+		prague:         pool.eip7623,
```
LuisPH3, line 682: This could turn dangerous:
The pool contains flags with eip-granularity. the validation uses fork. 
Given 2 eips from the same fork: how would they be initialized? 

I like the simplification of this feature, but I am not sure if the validation can be correctly initialized. 
Shall we revert it and create an issue cleaning this in both the pool and the validation? wdyt?

facuMH: I rather keep using revision flags because we get these flags from the [chain config](https://github.com/ethereum/go-ethereum/blob/master/params/config.go#L380) which keeps block number of revisions and does not really have notion of EIPs (except for EIP150, EIP155 and EIP 158, which are anyway not used in the pool flags).  We inherit this configs in [`TxPool.reset`](https://github.com/0xsoniclabs/sonic/blob/main/evmcore/tx_pool.go#L1531) which is called during [`NewTxPool`](https://github.com/0xsoniclabs/sonic/blob/main/evmcore/tx_pool.go#L345).

In other words, our TxPool having an EIP granularity is an outdated design choice, which should probably be updated as well. 

As for your question `Given 2 eips from the same fork: how would they be initialized?`: Currently the 2 EIPs from Prague, use the same flag from the chain config. So it is equivalent to just keeping one revision flag.

LuisPH3: > In other words, our TxPool having an EIP granularity is an outdated design choice, which should probably be updated as well.
Yes, this is whay I mean, shall we do both changes together in a follow-up pr? Just to prevent having them in the two states simultaneusly. 

> As for your question Given 2 eips from the same fork: how would they be initialized?: given 2 EIPs from the same revision, they would either both be enabled or none, as far as I know revisions are not partially implemented.
As you say, this is implementation-dependent. The analysis to see if it is correct today is the same effort of updating the pool.

facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/737b7b5b5bd81510fb8b3bd4e5b15a57de05806b

LuisPH3: thanks

LuisPH3 (dismissed):

evmcore/validation_test.go
```diff
@@ -0,0 +1,775 @@
...
+	require.NoError(t, err)
+	signer := types.NewPragueSigner(big.NewInt(1))
+	signedTx, err := types.SignTx(types.NewTx(tx), signer, key)
+	require.NoError(t, err)
+	return signer, address, signedTx
+}
+
+// setNonce sets the nonce for a transaction.
+func setNonce(t *testing.T, tx types.TxData, nonce uint64) {
+	// setTxField(t, tx, "Nonce", nonce, nonce, nonce, nonce, nonce)
```
simonlechner, line 584: Outdated code?

simonlechner (commented):
Code wise this looks good. It is very hard to follow what checks are there and what checks have to be there in which order. I think some kind of documentation is required. This could also be used to then document what is tested in which test, the scope of those checks are quite extensive, it is very difficult to remember what is tested where.

evmcore/validation.go
```diff
@@ -0,0 +1,190 @@
+// Copyright 2014 The go-ethereum Authors
```
HerbertJordan, line 1: We should probably not give the go-ethereum Authors the copyright for this file ..

evmcore/validation.go
HerbertJordan, line 1: nit: rename the file to `tx_validation.go`?

evmcore/validation.go
```diff
@@ -0,0 +1,190 @@
...
+	// Make sure the transaction is signed properly.
+	from, err := types.Sender(signer, tx)
+	if err != nil {
+		return ErrInvalidSender
+	}
+
+	// Drop non-local transactions under our own minimal accepted gas price or tip
+	local := opt.isLocal || opt.locals.contains(from) // account may be local even if the transaction arrived from the network
+	if !local && tx.GasTipCapIntCmp(opt.minTip) < 0 {
+		log.Trace("Rejecting underpriced tx: pool.gasPrice", "pool.gasPrice",
```
HerbertJordan, line 136: should "pool.gasPrice" be renamed to "pool.minTip"?

evmcore/validation_test.go
```diff
@@ -0,0 +1,774 @@
...
+			require.NoError(t, err)
+		})
+	}
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// Helper functions for testing.
+
+// getTxsFromAllTypes returns a list of all transaction types for testing.
+func getTxsFromAllTypes() map[string]types.TxData {
```
HerbertJordan, line 560: nit: `getTxsOfAllTypes`?

evmcore/validation_test.go
```diff
@@ -0,0 +1,774 @@
...
+	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/crypto/kzg4844"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/holiman/uint256"
+	"github.com/stretchr/testify/require"
+)
+
+func TestValidateTx_RejectsNonLegacyTransactions_BeforeEip2718(t *testing.T) {
```
HerbertJordan, line 19: nit: first the condition then the effect
```suggestion
func TestValidateTx_BeforeEip2718_RejectsNonLegacyTransactions(t *testing.T) {
```

evmcore/validation.go
```diff
@@ -0,0 +1,190 @@
...
+	minTip         *big.Int      // Minimum gas tip to enforce for acceptance into the pool
+
+	locals  *accountSet // Set of local transaction to exempt from eviction rules
+	isLocal bool        // Whether the transaction came from a local source
+
+}
+
+// validateTx checks whether a transaction is valid according to the current
+// options and adheres to some heuristic limits of the local node (price and size).
+func validateTx(tx *types.Transaction, signer types.Signer, opt validationOptions) error {
```
HerbertJordan, line 55: nit: include the `signer` in the options?

evmcore/validation_test.go
```diff
@@ -0,0 +1,774 @@
...
+		}
+		t.Run(name, func(t *testing.T) {
+			err := validateTx(types.NewTx(tx), types.NewPragueSigner(big.NewInt(1)),
+				validationOptions{eip2718: false})
+			require.ErrorIs(t, ErrTxTypeNotSupported, err)
+		})
+	}
+}
+
+// testTransactionsOption is a set of options to adjust the validation of transactions
```
HerbertJordan, line 32:
```suggestion
// testTransactionsOption returns a set of options to adjust the validation of transactions
```

evmcore/validation_test.go
```diff
@@ -0,0 +1,774 @@
...
+// so that it would accept all types of transactions, considering them as local transactions
+// with a min tip of 1, current base fee of 1, and a current max gas of 100_000.
+func testTransactionsOption() validationOptions {
```
HerbertJordan, line 35: I'd say either
```suggestion
var testTransactionsOption = validationOptions {
```
or
```suggestion
func getTestTransactionsOption() validationOptions {
```
The function option has the advantage that the there can not be an accidental modification due to sharing the same instance among multiple tests.

evmcore/validation_test.go
```diff
@@ -0,0 +1,774 @@
...
+		currentBaseFee: big.NewInt(1),
+		minTip:         big.NewInt(1),
+		isLocal:        true,
+	}
+}
+
+func TestValidateTx_RejectsTxBasedOnTypeAndActiveRevision(t *testing.T) {
+	tests := map[string]struct {
+		tx   *types.Transaction
+		opts func(validationOptions) validationOptions
```
HerbertJordan, line 53: nit: `opts` => `modify` or `configure`

evmcore/validation_test.go
```diff
@@ -0,0 +1,774 @@
...
+			require.ErrorIs(t, err, ErrUnderpriced)
+		})
+	}
+
+	// GasTipCap test
+	for name, tx := range getTxsFromAllTypes() {
+		t.Run(fmt.Sprintf("gas tip longer than 256 bits/%s", name), func(t *testing.T) {
+			// Legacy and access list transactions do not conceptually have a tip.
+			if isLegacyOrAccessList(tx) {
+				t.Skip("legacy and access list transactions do not have gas tip")
```
HerbertJordan, line 176: While they do not have gas-tips, their gas-price fields should be treated as both, the gas-tip-cap and gas-fee-cap. So also for those, this test could be relevant.

evmcore/validation_test.go
```diff
@@ -0,0 +1,774 @@
...
+				t.Fatal("unknown transaction type")
+			}
+		})
+	}
+
+	for name, tx := range getTxsFromAllTypes() {
+		t.Run(fmt.Sprintf("gas tip lower than pool min tip/%v", name), func(t *testing.T) {
+			// legacy and access list transactions do not have tip cap
+			if isLegacyOrAccessList(tx) {
+				t.Skip("legacy and access list transactions do not have tip cap")
```
HerbertJordan, line 202: Please check whether the gas-price field serves the role of the gas-tip-cap for these types of transactions.

evmcore/validation_test.go
```diff
@@ -0,0 +1,774 @@
...
+				opt)
+			require.NoError(t, err)
+		})
+	}
+
+}
+
+func TestValidateTx_Signer_RejectsTxWith(t *testing.T) {
+	for name, tx := range getTxsFromAllTypes() {
+		t.Run(fmt.Sprintf("unsigned tx/%v", name), func(t *testing.T) {
```
HerbertJordan, line 409: "unsigned tx" => "invalid signature"?

evmcore/validation_test.go
```diff
@@ -0,0 +1,774 @@
...
+			setGasPriceOrFeeCap(t, tx, opt.currentBaseFee)
+			setGas(t, tx, opt.currentMaxGas)
+
+			// sign txs with sender
+			signer, address, signedTx := signTxForTest(t, tx)
+			// ---
+
+			// setup low balance
+			testDb := newTestTxPoolStateDb()
+			// balance = gas * fee cap + value
```
HerbertJordan, line 452: Please also include a positive `value` in the transaction to make sure it is considered in the validation.

evmcore/validation_test.go
```diff
@@ -0,0 +1,774 @@
...
+}
+
+// setGasTipCap sets the gas fee cap for a transaction. For legacy and access list
+// transactions, it sets the gas price.
+func setGasTipCap(t *testing.T, tx types.TxData, gasTipCap *big.Int) {
+	switch tx := tx.(type) {
+	case *types.LegacyTx:
+		t.Fatal("legacy transactions do not have gas tip cap")
+	case *types.AccessListTx:
+		t.Fatal("access list transactions do not  have gas tip cap")
```
HerbertJordan, line 607: For these, the `GasPrice` serves as the `GasFeeCap` and `GasTipCap` equally.

evmcore/validation_test.go
```diff
@@ -0,0 +1,774 @@
...
+		tx.Data = data
+	case *types.SetCodeTx:
+		tx.Data = data
+	default:
+		t.Fatalf("unexpected transaction type: %T", tx)
+	}
+}
+
+// setToNil is a helper function to set the "To" field of a transaction to nil.
+func setToNil(t *testing.T, tx types.TxData) {
```
HerbertJordan, line 678: `setToNil` => `setReceiverToNil` (since `setToToNil` sounds strange)

HerbertJordan (changes requested):
Looks good, but as other reviewers commented, it is very hard to determine the systematic testing of these properties. I assume the code coverage is 100%?

There are a range of minor issues I commented on, please have a look.

evmcore/validation.go
facuMH: https://github.com/0xsoniclabs/sonic-admin/issues/149

evmcore/validation_test.go
facuMH: I like the function option better. Will rename accordingly.

evmcore/validation_test.go
facuMH: the decision to skip them was the result of semantics discussion with @LuisPH3, which started from https://github.com/0xsoniclabs/sonic/pull/154#discussion_r2024696281.

evmcore/validation.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/9ea293d2a5d32b4d447845935fa88a7397446828

evmcore/validation_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/9ea293d2a5d32b4d447845935fa88a7397446828

evmcore/validation_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/9ea293d2a5d32b4d447845935fa88a7397446828

evmcore/validation.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/9ea293d2a5d32b4d447845935fa88a7397446828

evmcore/validation_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/4205286062d9a53c041a02d2ccb9f0719ab7ba0f

evmcore/validation_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/4205286062d9a53c041a02d2ccb9f0719ab7ba0f

evmcore/validation_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/4205286062d9a53c041a02d2ccb9f0719ab7ba0f

evmcore/validation_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/4205286062d9a53c041a02d2ccb9f0719ab7ba0f

evmcore/validation_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/4205286062d9a53c041a02d2ccb9f0719ab7ba0f

evmcore/validation_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/4205286062d9a53c041a02d2ccb9f0719ab7ba0f

evmcore/validation.go
LuisPH3: Once GPL, always GPL. I don't think you can remove any of these copyrights.
Nevertheless, this is a problem for a later time.

evmcore/validation.go
LuisPH3: we are planning a PR to correct this naming.

evmcore/validation_test.go
LuisPH3: I explicitly requested not to alias tip logic in the tests. The function is named set tip, and these transactions types cannot have tip. I would like to emphasize some basic hygene of concepts, so that no one gets confused. (i.e. the gas tip- gas fee- gas isse that we have to rename: https://github.com/0xsoniclabs/sonic-admin/issues/149 )

evmcore/validation.go
HerbertJordan: I though the header might be an accident. Was not aware that it was there before.

evmcore/validation_test.go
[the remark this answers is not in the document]
HerbertJordan: If I understand correctly, Legacy and Access List transactions are wrapped into transaction adapters in the TxPool that allows to extract gas fee and tip caps from them -- by using the underlying gas price for both values. 

Thus, while the input format does not support it, the transaction pool is still deriving those values from these types through an adapter. And this should be tested.

evmcore/validation_test.go
HerbertJordan: Even though `LegacyTx` and `AccessListTx` do not have explicit tips, in the implementation an effective tip is derived, just as for other transaction types. This should be tested.

Maybe the function `setGasTipCap` needs to be renamed. The way it is used in the test is not that the intention is to actually set the gas tip of a transaction to a given value but to make to change the transaction such that its effective gas-tip has a certain value -- since this is what is tested in the transaction verification code, right?

HerbertJordan (commented):
Changes look good, but there are still open discussion threads to be addressed.

evmcore/validation_test.go
HerbertJordan: I do not think this was resolved. I do not see how the `value` would not be zero, and the linked commit touches changes not related to this.

facuMH: https://github.com/0xsoniclabs/sonic/commit/4205286062d9a53c041a02d2ccb9f0719ab7ba0f#diff-c2357ac27ddbb583bf7649ef1daa11ce9cd1b47a6c1c0e9483621f931c3d8138R449
this is the specific line from the linked committee that adds the value into the formula. 
But as you mentioned, all txs have a `value` of 0, I will change this, to a non zero value.

evmcore/validation_test.go
facuMH: renamed `setGasTipCap` to `setEffectiveTip` and added the necessary considerations. https://github.com/0xsoniclabs/sonic/pull/154/commits/2bcc9f0c1d81257091eec396d0f790809ad9eae1

evmcore/validation_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/2bcc9f0c1d81257091eec396d0f790809ad9eae1
in this commit all transactions get assigned a non zero value.

evmcore/validation_test.go
facuMH: skip removed in https://github.com/0xsoniclabs/sonic/pull/154/commits/2bcc9f0c1d81257091eec396d0f790809ad9eae1

evmcore/tx_validation_test.go
```diff
@@ -0,0 +1,790 @@
...
+
+			err := validateTx(types.NewTx(tx), getTestTransactionsOption())
+			require.ErrorIs(t, err, ErrMaxInitCodeSizeExceeded)
+		})
+	}
+
+	for name, tx := range getTxsOfAllTypes() {
+		t.Run(fmt.Sprintf("init code size not checked before shanghai/%v", name), func(t *testing.T) {
+			if isBlobOrSetCode(tx) {
+				t.Skip("blob and setCode transactions cannot be used before eip4844 and eip7702")
```
HerbertJordan, line 381:
```suggestion
				t.Skip("blob and setCode transactions cannot be used to initialize a contract")
```

evmcore/tx_validation_test.go
```diff
@@ -0,0 +1,790 @@
...
+			opt.signer = types.HomesteadSigner{}
+			setSignatureValues(t, tx, big.NewInt(1), big.NewInt(2), big.NewInt(3))
+			err := validateTx(types.NewTx(tx), getTestTransactionsOption())
+			require.ErrorIs(t, err, ErrInvalidSender)
+		})
+	}
+
+	for name, tx := range getTxsOfAllTypes() {
+		t.Run(fmt.Sprintf("invalid signer/%v", name), func(t *testing.T) {
+			setSignatureValues(t, tx, big.NewInt(1), big.NewInt(2), big.NewInt(3))
```
HerbertJordan, line 420: This line seems to be unnecessary; removing it does not change the outcome of the test.

evmcore/tx_validation_test.go
```diff
@@ -0,0 +1,790 @@
...
+			opt := getTestTransactionsOption()
+			opt.signer = types.HomesteadSigner{}
+			setSignatureValues(t, tx, big.NewInt(1), big.NewInt(2), big.NewInt(3))
+			err := validateTx(types.NewTx(tx), getTestTransactionsOption())
+			require.ErrorIs(t, err, ErrInvalidSender)
+		})
+	}
+
+	for name, tx := range getTxsOfAllTypes() {
+		t.Run(fmt.Sprintf("invalid signer/%v", name), func(t *testing.T) {
```
HerbertJordan, line 419: please check: I think this test actually checks for an invalid chain-Id.
```suggestion
		t.Run(fmt.Sprintf("invalid chain-id/%v", name), func(t *testing.T) {
```

evmcore/tx_validation_test.go
```diff
@@ -0,0 +1,790 @@
...
+
+			// setup low balance
+			testDb := newTestTxPoolStateDb()
+			// balance = gas * fee cap + value
+			zero := uint256.NewInt(0)
+			gasCost := zero.Mul(
+				uint256.NewInt(opt.currentMaxGas),
+				uint256.MustFromBig(opt.currentBaseFee),
+			)
+			gasCost = zero.Add(gasCost, uint256.MustFromBig(signedTx.Value()))
```
HerbertJordan, line 459: nit:
```suggestion
			txCost := zero.Add(gasCost, uint256.MustFromBig(signedTx.Value()))
```

HerbertJordan (approved):
Looks good, only some minor remarks then it is good to be merged.

evmcore/tx_validation_test.go
```diff
@@ -0,0 +1,790 @@
...
+			err := validateTx(types.NewTx(tx), validationOptions{eip2718: false})
+			require.ErrorIs(t, ErrTxTypeNotSupported, err)
+		})
+	}
+}
+
+// getTestTransactionsOption returns a set of options to adjust the validation of transactions
+// so that it would accept all types of transactions, considering them as local transactions
+// with a min tip of 1, current base fee of 1, and a current max gas of 100_000.
+func getTestTransactionsOption() validationOptions {
```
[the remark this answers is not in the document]
facuMH: I prefer the current `getTestValidationOptions` because some options are numerical values, or complex structs. So naming it "enabled" would give the (wrong) impression that this is just a set of flags.

evmcore/tx_validation_test.go
```diff
@@ -0,0 +1,790 @@
...
+			err = validateTx(signedTx, opt)
+			require.ErrorIs(t, err, ErrFloorDataGas)
+		})
+	}
+
+	for name, tx := range getTxsOfAllTypes() {
+		t.Run(fmt.Sprintf("floor data gas not checked before eip7623/%v", name), func(t *testing.T) {
+			if _, ok := tx.(*types.SetCodeTx); ok {
+				t.Skip("setCode transactions cannot be used before eip7702")
+			}
```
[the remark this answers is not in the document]
facuMH: right. This check is a remnant of converting all eips flags to revisions, where 7702 and 7623 were both prague. shall be removed.

evmcore/tx_validation_test.go
facuMH: I know, I just added to make clear the intent of having "garbage" value in those fields, instead of leaving just default values.

evmcore/tx_validation_test.go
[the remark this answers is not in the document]
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/6208859ae9e8f578ac22b5dae776e9109e62bc7e

evmcore/tx_validation_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/6208859ae9e8f578ac22b5dae776e9109e62bc7e

evmcore/tx_validation_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/6208859ae9e8f578ac22b5dae776e9109e62bc7e

evmcore/tx_validation_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/154/commits/6208859ae9e8f578ac22b5dae776e9109e62bc7e

simonlechner (approved):

facuMH:
> ## [Codecov](https://app.codecov.io/gh/0xsoniclabs/sonic/pull/154?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=0xsoniclabs) Report
> Attention: Patch coverage is `94.69027%` with `6 lines` in your changes missing coverage. Please review.
> 
> [Files with missing lines](https://app.codecov.io/gh/0xsoniclabs/sonic/pull/154?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=0xsoniclabs)	Patch %	Lines
> [evmcore/tx_validation.go](https://app.codecov.io/gh/0xsoniclabs/sonic/pull/154?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=0xsoniclabs#diff-ZXZtY29yZS90eF92YWxpZGF0aW9uLmdv)	93.75%	[4 Missing and 2 partials ⚠️ ](https://app.codecov.io/gh/0xsoniclabs/sonic/pull/154?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=0xsoniclabs)
> 📢 Thoughts on this report? [Let us know!](https://github.com/codecov/feedback/issues/255)

missing lines are overflow error cases from geth/core functions which require over 400TB of allocated memory to trigger (for the [FloorDataCost](https://github.com/ethereum/go-ethereum/blob/master/core/state_transition.go#L120) or over 200TB for [IntrinsicGas]. With that in mind, I think this coverage is fine.

LuisPH3 (approved):

PR #155, opened 2025-03-31
Use single background worker for block notifications

HerbertJordan:
WIP


gossip/service.go
```diff
@@ -80,6 +88,64 @@ func (f *ServiceFeed) SubscribeNewLogs(ch chan<- []*types.Log) notify.Subscripti
...
+					select {
+					case <-stop:
+						return
+					case <-ticker.C:
+					}
+				}
+			}
+		}
+	}()
+}
```
HerbertJordan, line 127: Could be simplified to
```
func (f *ServiceFeed) Start(store *evmstore.Store) {
	incoming := make(chan feedUpdate, 1024)
	f.incomingUpdates = incoming
	stop := make(chan struct{})
	done := make(chan struct{})
	f.stopFeeder = stop
	f.feederDone = done
	go func() {
		defer close(done)
		ticker := time.NewTicker(10 * time.Millisecond)
		defer ticker.Stop()
		pending := []feedUpdate{}
		for {
			select {
			case <-stop:
				return
			case update := <-incoming:
				pending = append(pending, update) // assuming updates are in block order ...
			case <-ticker.C:
				height, empty, err := store.GetArchiveBlockHeight()
				if err != nil {
					log.Error("failed to get archive block height", "err", err)
					continue
				}
				if empty {
					continue
				}
				for i, update := range pending {
					if update.block.Number.Uint64() > height {
						pending = pending[i:]
						break
					}
					f.newBlock.Send(evmcore.ChainHeadNotify{Block: update.block})
					f.newLogs.Send(update.logs)
				}
			}
		}
	}()
}
```

HerbertJordan:
Integrated in #149

PR #157, opened 2025-04-01
Add execution spec tests

simonlechner:
The ethereum tests are no longer updated for all new eips, therefore the execution spec tests should be generated and ran against sonic to test its block processing and database. This PR extends the ethereum tests runner to also run the execution spec tests when filled.

### DIFF
--- a/tests/state_test.go
+++ b/tests/state_test.go
@@ -29,9 +29,11 @@ import (
 )
 
 var (
-	eipTests          = filepath.Join(".", "testdata", "EIPTests", "StateTests")
-	generalStateTests = filepath.Join(".", "testdata", "GeneralStateTests")
-	execSpecTests     = filepath.Join(".", "execution-spec-tests", "fixtures", "state_tests")
+	testPaths = []string{
+		filepath.Join(".", "testdata", "EIPTests", "StateTests"),
+		filepath.Join(".", "testdata", "GeneralStateTests"),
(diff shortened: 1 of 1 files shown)


LuisPH3 (dismissed):
Could you additionally write some lines on how to use it? 
So far, we know that we need to clone the eth tests. How does this PR change the workflow now?

tests/state_test.go
```diff
@@ -29,8 +29,9 @@ import (
 )
 
 var (
-	baseDir      = filepath.Join(".", "testdata")
-	stateTestDir = filepath.Join(baseDir, "GeneralStateTests")
+	eipTests          = filepath.Join(".", "testdata", "EIPTests", "StateTests")
+	generalStateTests = filepath.Join(".", "testdata", "GeneralStateTests")
+	execSpecTests     = filepath.Join(".", "execution-spec-tests", "fixtures", "state_tests")
```
HerbertJordan, line 34: nit: it would be better to define a slice here and use it below. So if in the future directories need to be modified, there is only a single place to change.

HerbertJordan (dismissed):
As @LuisPH3 mentioned, would be great to include some documentation in the code on how to fill the respective directories.

simonlechner:
I added documentation for both the ethereum/tests and ethereum/execution-spec-tests.

HerbertJordan (approved):

LuisPH3 (approved):

PR #167, opened 2025-04-09
Make ChainId field being returned from RPC calls for LegacyTxs

drwtsn95:
Before now, jsonRPC methods, such as eth_getTransactionByHash missed the chainId field in the output. It is not compatible with another popular EVM chains RPC model, such as Ethereum, Polygon, BSC. Only with chainId filled properly one can check the transactions signature because is able to calculate the correct transaction hash.

### DIFF
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -1383,6 +1383,11 @@ func newRPCTransaction(tx *types.Transaction, blockHash common.Hash, blockNumber
 	}
 
 	switch tx.Type() {
+	case types.LegacyTxType:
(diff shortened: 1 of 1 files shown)


HerbertJordan:
Hi @drwtsn95,

Unfortunately we do not have established all the legal preconditions for accepting third-party contributions yet. Work on this is in progress, but there is no clarity yet on the desired and/or required policies. Thus, I cannot accept this contribution right now. 

Nevertheless, I would like to thank you for notifying us on this issue. We have integrated it in our internal issue tracking and will include a corresponding fix in future releases of Sonic.

Thanks!

PR #169, opened 2025-04-10
Add fuzzer for scrambler 

LuisPH3:
This PR adds a fuzzing test for the public interface of the scrambler.

The fuzzer generates lists of transactions to be scrambled, shuffled and scrambled again to test reproducibility of the sorting algorithm. 

This PR moves the scrambler into its own package, this is done because fuzzing requires all tests from the package to pass before it can start.  This change reduces gossip package size, and it can be done because the scrambler does not depend on any other member of gossip.

### DIFF
--- a/gossip/scrambler/tx_scrambler.go
+++ b/gossip/scrambler/tx_scrambler.go
@@ -1,12 +1,13 @@
-package gossip
+package scrambler
 
 import (
 	"bytes"
 	"cmp"
-	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/core/types"
 	"math/big"
 	"slices"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
 )
 
 // ScramblerEntry stores meta information about transaction for sorting and filtering them.
@@ -43,11 +44,11 @@ func (tx *scramblerTransaction) Sender() common.Address {
 	return tx.sender
 }
 
-// getExecutionOrder returns correct order of the transactions.
+// GetExecutionOrder returns correct order of the transactions.
 // If Sonic is enabled, the tx scrambler is used, otherwise the
 // order stays unchanged. If signer is unable to derive sender for
 // a transaction, this transaction is not excluded from the final list.
-func getExecutionOrder(unorderedTxs types.Transactions, signer types.Signer, isSonic bool) types.Transactions {
+func GetExecutionOrder(unorderedTxs types.Transactions, signer types.Signer, isSonic bool) types.Transactions {
 	// Don't use scrambler if Sonic is not enabled
 	if !isSonic {
 		return unorderedTxs
@@ -130,8 +131,8 @@ func sortTransactionsWithSameSender(entries []ScramblerEntry) {
 func scrambleTransactions(list []ScramblerEntry, salt [32]byte) {
 	var aX, bX [32]byte
 	slices.SortFunc(list, func(a, b ScramblerEntry) int {
-		aX = xorBytes32(a.Hash(), salt)
-		bX = xorBytes32(b.Hash(), salt)
+		aX = XorBytes32(a.Hash(), salt)
+		bX = XorBytes32(b.Hash(), salt)
 		return bytes.Compare(aX[:], bX[:])
 	})
 }
@@ -157,7 +158,7 @@ func analyseEntryList(entries []ScramblerEntry) ([]ScramblerEntry, [32]byte, boo
 			hasDuplicateAddresses = true
 		}
 		seenAddresses[sender] = struct{}{}
-		salt = xorBytes32(salt, entry.Hash())
+		salt = XorBytes32(salt, entry.Hash())
 		uniqueList = append(uniqueList, entry)
 		seenHashes[entry.Hash()] = struct{}{}
 
@@ -166,7 +167,7 @@ func analyseEntryList(entries []ScramblerEntry) ([]ScramblerEntry, [32]byte, boo
 	return uniqueList, salt, hasDuplicateAddresses
 }
 
-func xorBytes32(a, b [32]byte) (dst [32]byte) {
+func XorBytes32(a, b [32]byte) (dst [32]byte) {
 	for i := 0; i < 32; i++ {
 		dst[i] = a[i] ^ b[i]
 	}
--- a/gossip/scrambler/tx_scrambler.go
+++ b/gossip/scrambler/tx_scrambler.go
@@ -131,8 +131,8 @@ func sortTransactionsWithSameSender(entries []ScramblerEntry) {
 func scrambleTransactions(list []ScramblerEntry, salt [32]byte) {
 	var aX, bX [32]byte
 	slices.SortFunc(list, func(a, b ScramblerEntry) int {
-		aX = XorBytes32(a.Hash(), salt)
-		bX = XorBytes32(b.Hash(), salt)
+		aX = xorBytes32(a.Hash(), salt)
+		bX = xorBytes32(b.Hash(), salt)
 		return bytes.Compare(aX[:], bX[:])
 	})
 }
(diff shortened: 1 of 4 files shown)


gossip/scrambler/tx_scrambler_fuzz_test.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+
+		txs := make([]*types.Transaction, 0, len(metaTxs))
+		for _, metaTx := range metaTxs {
+
+			key := accountKeys[metaTx.SenderAccount]
+			tx, err := types.SignTx(types.NewTx(&types.LegacyTx{
+				Nonce:    metaTx.Nonce,
+				GasPrice: big.NewInt(int64(metaTx.GasPrice)),
+			}), signer, key)
```
HerbertJordan, line 69: The overhead of signing transactions could be eliminated, and thereby the fuzzing likely significantly speed up, by targeting the function `filterAndOrderTransactions` instead of `GetExecutionOrder`;

gossip/scrambler/tx_scrambler_fuzz_test.go
```diff
@@ -0,0 +1,201 @@
...
+		GasPrice:      math.MaxUint64,
+	})
+	require.NoError(t, err)
+	return len(bytes)
+}
+
+// scrambledOrderIsReproducible is a naive implementation that checks if the scrambler order is
+// reproducible.
+// It is meant to be used by tests only.
+func scrambledOrderIsReproducible(t testing.TB, ordered types.Transactions, signer types.Signer) {
```
HerbertJordan, line 94: nit: this seems to be an integral part of the fuzzer test above and it should not be its own function; wdyt?

HerbertJordan (changes requested):
The test seems a bit too overloaded with content that is not really needed. I added a few suggestions how to slim it down without compromising its objective. Please have a look.

gossip/scrambler/tx_scrambler_fuzz_test.go
```diff
@@ -0,0 +1,201 @@
...
+			txs = append(txs, tx)
+		}
+
+		ordered := scrambler.GetExecutionOrder(txs, signer, true)
+
+		scrambledOrderIsReproducible(t, ordered, signer)
+	})
+}
+
+func sizeOfEncodedMetaTransaction(t testing.TB) int {
```
facuMH, line 81: small comment: this function is only used once, during setup. could be inlined.

gossip/scrambler/tx_scrambler_fuzz_test.go
```diff
@@ -0,0 +1,201 @@
...
+// reproducible.
+// It is meant to be used by tests only.
+func scrambledOrderIsReproducible(t testing.TB, ordered types.Transactions, signer types.Signer) {
+	t.Helper()
+
+	testList := slices.Clone(ordered)
+
+	// shuffle the list, but in a deterministic way
+	slices.SortFunc(testList, func(a, b *types.Transaction) int {
+		return bytes.Compare(a.Hash().Bytes(), b.Hash().Bytes())
```
facuMH, line 101: do you think it would add robustness to the test to check different initial orders ?

gossip/scrambler/tx_scrambler_fuzz_test.go
```diff
@@ -0,0 +1,201 @@
...
+		if len(encoded) > 10_000*maxMetaTransactionEncodedSize {
+			t.Skip("input too large")
+		}
+
+		stream := rlp.NewStream(bytes.NewReader(encoded), 0)
+
+		metaTxs := make([]metaTransaction, 0)
+		if err := stream.Decode(&metaTxs); err != nil {
+			t.Skip("invalid input", err)
+		}
```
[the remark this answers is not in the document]
facuMH: I had a similar thought, why not fuzz over the 3 parameters instead of the encoding? unless the intent is to also check the robustness of the encoding/decoding ?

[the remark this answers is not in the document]
LuisPH3: Here we are not fuzzing one metaTransaction, we are fuzzing a list of fuzzed meta transactions. How would we model this in fuzzer terms?  a list of bytes that can be splitted in 3-element tuples?  partial encodings would still be found in this format and this if would still need to exist. 

rlp is a convenience here, and it is actually very little and straight forward code.

gossip/scrambler/tx_scrambler.go
```diff
@@ -166,7 +167,7 @@ func analyseEntryList(entries []ScramblerEntry) ([]ScramblerEntry, [32]byte, boo
 	return uniqueList, salt, hasDuplicateAddresses
 }
 
-func xorBytes32(a, b [32]byte) (dst [32]byte) {
+func XorBytes32(a, b [32]byte) (dst [32]byte) {
```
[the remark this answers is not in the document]
LuisPH3: yes, because I want to target the scrambler public interface, I have two options, to make it public or to duplicate it. wdyt?

gossip/scrambler/tx_scrambler_fuzz_test.go
LuisPH3: I want to explicitly target the public interface of the scrambler. To prevent having to rewrite the fuzzer in case the scrambler is entirely rewritten. 
Because the signature is used to retrieve the sender, and the sender is relevant to the fuzzing. I would like to preserve the signature.

It is not that bad in execution time, after experimental runs, I expect this fuzzer to converge in a few hours.

gossip/scrambler/tx_scrambler_fuzz_test.go
LuisPH3: It just reads better as a function.

gossip/scrambler/tx_scrambler_fuzz_test.go
LuisPH3: If the fuzzer is rewritten, and a better mechanism-testing is found, this would be the one function to rewrite. 
We could even have different metrics. (are all tx from same sender consecutive? )

gossip/scrambler/tx_scrambler_fuzz_test.go
LuisPH3: I thought about it. I don't think so. My instinct is that sorting by hashes is the most entropic order I can come out with. 

I even tried `rand.shuffe` the distribution remains the same.

gossip/scrambler/tx_scrambler_test.go
```diff
@@ -141,6 +142,48 @@ func TestTxScrambler_ScrambleTransactions_ScrambleIsDeterministic(t *testing.T)
...
+				return a.Hash().Cmp(b.Hash())
+			}
+
+			aX := XorBytes32(a.Hash(), salt)
+			bX := XorBytes32(b.Hash(), salt)
+			return bytes.Compare(aX[:], bX[:])
+		}
+
+		// check that the order is correct
+		if !slices.IsSortedFunc(testInput, orderFunc) {
```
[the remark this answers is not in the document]
LuisPH3: you are right, this should not have been commited.

gossip/scrambler/tx_scrambler.go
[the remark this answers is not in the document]
LuisPH3: right

LuisPH3:
> The test seems a bit too overloaded with content that is not really needed. I added a few suggestions how to slim it down without compromising its objective. Please have a look.

about extras. 
- signature is required because this test targets the public interface. Performance is not an issue.
- rlp encoding is the off-the self solution which implements the technique described to encode the metaTransactions.

gossip/scrambler/tx_scrambler_fuzz_test.go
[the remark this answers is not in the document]
HerbertJordan: > How would we model this in fuzzer terms?

Just read the first n*17 bytes of a `[]byte` input, ignore the trailing remaining bytes. Every input is valid, nothing to be skipped.

> RLP is a core feature of the ethereum blockchain, I do not think we should treat it as an enemy if It can really solve problems for us.

I hope I do not make it look like an ideological war. That is not my intention. RLP has its merits and its justification. But it should used in contexts where it is reasonable, and not just because it is available. It has its implications, like the variable-length encoding of integers, that can make some inputs invalid.

HerbertJordan (approved):

gossip/scrambler/tx_scrambler_fuzz_test.go
LuisPH3: inlined

HerbertJordan (approved):

simonlechner (approved):
Looks good, thanks.

facuMH (approved):

PR #174, opened 2025-04-14
Add block proposal type for block-proposer upgrade

HerbertJordan:
This PR introduces the type used for proposing blocks in the new block-proposal protocol developed on top of lachesis.

The type is equipped with a hashing function and serialization support. The serialization implementation uses the [Protocol Buffer](https://protobuf.dev/) format to simplify future modifications and easier portability among different client implementations.


inter/proposal.go
```diff
@@ -0,0 +1,102 @@
...
+	ParentHash   common.Hash
+	Time         Timestamp
+	Randao       common.Hash
+	Transactions []*types.Transaction
+}
+
+// Hash computes a cryptographic hash of the proposal. The hash can be used to
+// sign and verify the proposal.
+func (p *Proposal) Hash() hash.Hash {
+	size := 8 + 4 + 32 + 8 + 32 + 32*len(p.Transactions)
```
LuisPH3, line 34: Premature optimization?
Decoupling members and sizes is an error-prone practice.

inter/proposal.go
```diff
@@ -0,0 +1,102 @@
...
+	data = binary.BigEndian.AppendUint64(data, uint64(p.Number))
+	data = binary.BigEndian.AppendUint32(data, uint32(p.Attempt))
+	data = append(data, p.ParentHash[:]...)
+	data = binary.BigEndian.AppendUint64(data, uint64(p.Time))
+	data = append(data, p.Randao[:]...)
+	for _, tx := range p.Transactions {
+		txHash := tx.Hash()
+		data = append(data, txHash[:]...)
+	}
+	return sha256.Sum256(data)
```
LuisPH3, line 45: question: why sha256?  don't we use Keccak256 ? is there some rule/literature where to choose what?

inter/proposal.go
```diff
@@ -0,0 +1,102 @@
...
+// A proposal is a candidate for inclusion in the blockchain and is
+// created by a proposer. It is signed by the proposer and sent to
+// validators for validation and inclusion in the blockchain.
+type Proposal struct {
+	Number       idx.Block
+	Attempt      uint32
+	ParentHash   common.Hash
+	Time         Timestamp
+	Randao       common.Hash
+	Transactions []*types.Transaction
```
LuisPH3, line 28: nit:
```suggestion
	Transactions types.Transactions
```

LuisPH3 (dismissed):

inter/proposal.go
HerbertJordan: Note that the code is just using the size as a hint. If the actual size would be different, the computation of the hash is not compromised. It does not allocate an array with this size and writes data into it, it allocates an empty slice with an underlying array of the given size and appends to the slice, expanding it if needed.

About the premature optimization - kinda borderline, as it is a clear improvement that does not significantly reduce readability. But I removed it.

LuisPH3 (dismissed):

simonlechner (dismissed):
Looks good to me.

inter/proposal_test.go
```diff
@@ -0,0 +1,193 @@
...
+			Time:       Timestamp(3 + i),
+			Randao:     [32]byte{0: 3, 1: byte(i), 21: 4},
+			Transactions: []*types.Transaction{
+				types.NewTx(&types.LegacyTx{Nonce: 1}),
+				types.NewTx(&types.LegacyTx{Nonce: 2}),
+				types.NewTx(&types.LegacyTx{Nonce: uint64(i)}),
+			},
+		}
+
+		hash := func(proposal *Proposal) hash.Hash {
```
facuMH, line 39: does this closure need to be inside the loop or could it be moved before it ? it would also prevent `proposal` shadowing.

inter/proposal.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// Hash computes a cryptographic hash of the proposal. The hash can be used to
+// sign and verify the proposal.
+func (p *Proposal) Hash() hash.Hash {
+	data := []byte{}
```
LuisPH3, line 33: One of the features that we get from using protobufers, is that the serialization is somehow resilient to version changes. 
This hashing implements an ad-hoc serialization that does not have that property.
Shall we hash a serialized stream instead?

LuisPH3 (approved):

PR #176, opened 2025-04-14
Fix flaky rejected-transactions integration test

HerbertJordan:
This PR fixes the expectation in the `TestRejectedTx_TransactionsAreRejectedBecauseOfAccountState` that caused it to be flaky.

Before this change it was assumed that once a receipt for a transaction was received, the transaction pool has already been updated and all processed transactions have been removed. However, this happens asynchronous, thus it is timing dependent whether a re-transmitted transaction is identified as "already known" (before the Tx-Pool update) or "nonce to low" (after the update).

This PR updates the test to accept "already known" responses, but expects the system to eventually transition to the "nonce to low" state.

### DIFF
--- a/tests/rejected_tx_test.go
+++ b/tests/rejected_tx_test.go
@@ -3,6 +3,7 @@ package tests
 import (
 	"context"
 	"math/big"
+	"strings"
 	"testing"
 
 	"github.com/0xsoniclabs/sonic/opera"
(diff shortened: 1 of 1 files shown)


LuisPH3 (approved):

LuisPH3 (changes requested):
the ci is still failing, i'll have a look

HerbertJordan:
Replaced by #177

PR #177, opened 2025-04-14
Refactor reject-transaction integration test to avoid flakiness

HerbertJordan:
Add-on to #176 

This PR adds a retry mechanism to wait for the tx pool to evict executed transactions once the state is updated.
It additionally refactors the generation of set code transactions with authorizations to avoid authorizations clash in the test.

### DIFF
--- a/tests/rejected_tx_test.go
+++ b/tests/rejected_tx_test.go
@@ -3,6 +3,7 @@ package tests
 import (
 	"context"
 	"math/big"
+	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


LuisPH3 (approved):

facuMH (approved):
This PR makes a refactor in the transaction generation for testing, to avoid account clashing on authorizations by generating them from different accounts and also introduces a loop to account for waiting of the re-organization loop.

PR #181, opened 2025-04-14
checks min base fee update within epoch

kjezek:
This PR tests if the `minBaseFee` network rule is applied only upon an epoch boundaries. 

At the moment this value is applied immediately.


tests/network_rules_update_test.go
```diff
@@ -0,0 +1,111 @@
...
+	var updatedRules rulesType
+	err = client.Client().Call(&updatedRules, "eth_getRules", "latest")
+	require.NoError(err)
+
+	require.Equal(originalRules.Economy.MinBaseFee, updatedRules.Economy.MinBaseFee,
+		"Network rules should not change - it must be an epoch bound")
+}
+
+// updateNetworkRules sends a transaction to update the network rules.
+func updateNetworkRules(t *testing.T, require *require.Assertions, net IntegrationTestNetSession, rulesChange any) {
```
LuisPH3, line 94: nit:
You do not need to pass the require object, It is ok to construct one in each function.
Or even do not construct it at all, and use the free-standing functions instead.

`require.True(t, expression)`  <- they implement this api as well.

opera/rules_test.go
```diff
@@ -0,0 +1,184 @@
+package opera
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestRules_Copy_DeepCopied(t *testing.T) {
```
LuisPH3, line 8: nit: we try to use full sentences in the test name, to describe the property tested:
ideas: 
```suggestion
func TestRules_Copy_CopiesAreDisjoint(t *testing.T) {
```
```suggestion
func TestRules_Copy_CopiesShareNoData(t *testing.T) {
```

```suggestion
func TestRules_Copy_IsImplementedAsDeepCopy(t *testing.T) {
```

tests/network_rules_update_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/0xsoniclabs/sonic/gossip/contract/driverauth100"
+	"github.com/0xsoniclabs/sonic/opera/contracts/driverauth"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/stretchr/testify/require"
+	"math/big"
+	"testing"
+)
+
+func TestNetworkRule_Update_Epoch_Boundaries(t *testing.T) {
```
LuisPH3, line 14:
```suggestion
func TestNetworkRule_Update_RulesChangeDuringEpochHasNoEffect(t *testing.T) {
```

LuisPH3 (commented):
The tests introduced are sufficient as regression tests. They target the pain-point.
IMO, for an integration test we may want to test the complete function:

- Issue change
- Check that change is not in effect
- Wait for end of epoch (somehow)
- Check that change was introduced

I have some minor style notes:

HerbertJordan (changes requested):
I agree with @LuisPH3 feedback that those tests should cover the effect of epochs.

There is a function in the driver called [advanceEpoch](https://github.com/0xsoniclabs/sonic-sfc/blob/2520ec9e055b6586b3b65832859810aa9a1845c8/contracts/sfc/NodeDriver.sol#L90). I have never tried this myself, but it looks to me like it is similar to the network rule changes, but it allows you to force a epoch progression. Maybe you can try this to extend the coverage beyond a single epoch?

Alternatively, we can reduce the epoch duration for tests to a few seconds to see multiple epochs in a short time interval.

tests/network_rules_update_test.go
```diff
@@ -0,0 +1,204 @@
...
+	// Network rule should not change - it must be an epoch bound
+	var updatedRules rulesType
+	err = client.Client().Call(&updatedRules, "eth_getRules", "latest")
+	require.NoError(err)
+
+	require.Equal(originalRules.Economy.MinBaseFee, updatedRules.Economy.MinBaseFee,
+		"Network rules should not change - it must be an epoch bound")
+
+	var blockBefore evmcore.EvmBlockJson
+	err = client.Client().Call(&blockBefore, "eth_getBlockByNumber", "latest", false)
```
HerbertJordan, line 52: You should send some transaction and wait for the receipt to make sure that you get a block after the change has been applied. Otherwise, this may get you just the block where the change was applied, where the base-fee is settled before the block execution.

A simple way to produce a new block is to run 
```
_, err := net.EndowAccount(common.Address{}, big.NewInt(1))
require.NoError(err)
```
which will send 1 wei to the 0 account and wait for the receipt to be ready.

tests/network_rules_update_test.go
```diff
@@ -0,0 +1,204 @@
...
+	require.NoError(err)
```
HerbertJordan, line 53: For most common operations, the client library has a few ready-to-use features such that you do not have to know the explicit RPC methods. For this one, for instance, you can use
```suggestion
	blockBefore, err := client.BlockByNumber(t.Context(), nil)
	require.NoError(err)
```

tests/network_rules_update_test.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/stretchr/testify/require"
+	"math/big"
+	"testing"
+)
+
+func TestNetworkRule_Update_RulesChangeDuringEpochHasNoEffect(t *testing.T) {
+	require := require.New(t)
+	net := StartIntegrationTestNetWithFakeGenesis(t)
+	defer net.Stop()
```
HerbertJordan, line 19: nit: this is no longer needed. The function above is registering a shut-down handler in `t` automatically
```suggestion
```

tests/network_rules_update_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+	newMinBaseFee := 1e3 * originalRules.Economy.MinBaseFee.Int64()
+	updateRequest := rulesType{}
+	updateRequest.Economy.MinBaseFee = new(big.Int).SetInt64(newMinBaseFee)
+
+	// Update network rules
+	updateNetworkRules(t, net, updateRequest)
+
+	// Restart the network, since the rules happened withing a current epoch
+	// it should not be applied and persisted.
```
HerbertJordan, line 102:
```suggestion
	// it should not be applied immediately but persisted to be applied at the end of the epoch.
```

tests/network_rules_update_test.go
```diff
@@ -0,0 +1,204 @@
...
+	// wait the next two blocks as the fee is applied to the next block after
+	//the epoch change becomes effective
+	for {
+		var newBlock evmcore.EvmBlockJson
+		err = client.Client().Call(&newBlock, "eth_getBlockByNumber", "latest", false)
+		require.NoError(err)
+		if newBlock.Number.ToInt().Int64() > currentBlock.Number.ToInt().Int64()+1 {
+			break
+		}
+	}
```
HerbertJordan, line 203: nit: this is use case specific and should not be part of the `advanceEpoch` function. Maybe you can wrap the `advanceEpoch` in a `advanceEpochAndWaitForBlocks`?

HerbertJordan (approved):
Nice new feature in the integration tests! Thanks for building this!

tests/network_rules_update_test.go
LuisPH3: I was going to ask the same: why do we need to advance a block after an epoch change?

tests/network_rules_update_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+	receipt, err := net.Apply(func(ops *bind.TransactOpts) (*types.Transaction, error) {
+		return contract.UpdateNetworkRules(ops, b)
+	})
+
+	require.NoError(err)
+	require.Equal(receipt.Status, types.ReceiptStatusSuccessful)
+}
+
+func advanceEpoch(t *testing.T, net IntegrationTestNetSession) {
```
LuisPH3, line 158: This was surprisingly straight forward :tada:

tests/network_rules_update_test.go
```diff
@@ -0,0 +1,204 @@
...
+	err = client2.Client().Call(&updatedRules, "eth_getRules", "latest")
+	require.NoError(err)
+
+	require.Equal(originalRules.Economy.MinBaseFee, updatedRules.Economy.MinBaseFee,
+		"Network rules should not change - it must be an epoch bound")
+
+	// apply epoch change
+	advanceEpoch(t, net)
+
+	// rule should be effective
```
LuisPH3, line 121:
```suggestion
	// rule change should be effective
```

LuisPH3 (approved):
:+1:

tests/network_rules_update_test.go
kjezek: This is because some network rules are applied to a next block after the network rule change takes place. In this case, min base fee is applied to the next block after the block the network rule was applied

tests/network_rules_update_test.go
```diff
@@ -0,0 +1,218 @@
...
+	defer client2.Close()
+
+	// Network rule should not change - it must be an epoch bound
+	var updatedRules rulesType
+	err = client2.Client().Call(&updatedRules, "eth_getRules", "latest")
+	require.NoError(err)
+
+	require.Equal(originalRules.Economy.MinBaseFee, updatedRules.Economy.MinBaseFee,
+		"Network rules should not change - it must be an epoch bound")
+
```
HerbertJordan, line 117: Can you produce a block here, after the restart, and check that the old fee is still in use before the epoch boundary?

HerbertJordan (dismissed):
Looks good to me, thanks for extending it to test epoch boundaries as well.

HerbertJordan (dismissed):

opera/rules.go
```diff
@@ -420,6 +420,7 @@ func DefaultGasPowerRules() GasPowerRules {
 func (r Rules) Copy() Rules {
 	cp := r
 	cp.Economy.MinGasPrice = new(big.Int).Set(r.Economy.MinGasPrice)
+	cp.Economy.MinBaseFee = new(big.Int).Set(r.Economy.MinBaseFee)
```
LuisPH3, line 423: the changes introduced in https://github.com/0xsoniclabs/sonic/pull/198/commits/2ee6497e42d873f5ea517d79a4ed67c483addbbc still prevail, right?

tests/network_rules_update_test.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/stretchr/testify/require"
+	"math/big"
+	"testing"
+)
+
+func TestNetworkRule_Update_RulesChangeDuringEpochHasNoEffect(t *testing.T) {
```
LuisPH3, line 16: nit:
```suggestion
func TestNetworkRule_Update_RulesChangeIsDelayedUntilNextEpochStart(t *testing.T) {
```
Because the change has an effect, eventually.

tests/network_rules_update_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+	blockAfter, err := client.BlockByNumber(t.Context(), nil)
+	require.NoError(err)
+
+	require.GreaterOrEqual(blockAfter.BaseFee().Int64(), newMinBaseFee, "BaseFee should reflect new MinBaseFee")
+}
+
+func TestNetworkRule_Update_Restart_Recovers_Original_Value(t *testing.T) {
+	require := require.New(t)
+	net := StartIntegrationTestNetWithFakeGenesis(t)
```
LuisPH3, line 77: I presume that, once https://github.com/0xsoniclabs/sonic/pull/198/commits/2ee6497e42d873f5ea517d79a4ed67c483addbbc is merged. this test will have different behavioir before and after allegro. 
Shall we wait for the other changes to be merged?

opera/rules.go
kjezek: right, it will get merged after this PR

tests/network_rules_update_test.go
kjezek: This PR has to go first, the other PR was forked from this one.

HerbertJordan: Do you want to merge https://github.com/0xsoniclabs/sonic/commit/2ee6497e42d873f5ea517d79a4ed67c483addbbc into this PR and then merge them together as one into main?

HerbertJordan (approved):

tests/network_rules_update_test.go
```diff
@@ -0,0 +1,222 @@
...
+	require.NoError(err)
+
+	client2, err := net.GetClient()
+	require.NoError(err)
+	defer client2.Close()
+
+	// produce a block to make sure the rule is not applied
+	_, err = net.EndowAccount(common.Address{}, big.NewInt(1))
+	require.NoError(err)
+
```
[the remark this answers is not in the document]
kjezek: done

tests/network_rules_update_test.go
kjezek: will merge after

PR #184, opened 2025-04-16
Add configurable variable for RPC simulation with EVM

jenikd:
This PR adds an configurable flag to select, if EVM should be used for transaction simulation with `eth_call` RPC function

### DIFF
--- a/config/flags/flags.go
+++ b/config/flags/flags.go
@@ -330,6 +330,10 @@ var (
 		Name:  "exitwhensynced.epoch",
 		Usage: "Exits after synchronisation reaches the required epoch",
 	}
+	RPCEVMSimulationFlag = cli.BoolFlag{
+		Name:  "rpc.evmsimulation",
+		Usage: "EVM is used for RPC transaction simulation",
+	}
 
 	// Validator
 	ValidatorIDFlag = cli.UintFlag{
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -1014,7 +1014,7 @@ func (diff *StateOverride) Apply(state state.StateDB) error {
 	return nil
 }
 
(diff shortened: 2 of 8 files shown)


ethapi/api.go
```diff
@@ -1043,6 +1043,11 @@ func DoCall(ctx context.Context, b Backend, args TransactionArgs, blockNrOrHash
 		return nil, err
 	}
 	vmConfig := opera.DefaultVMConfig
+	// use EVM for transaction simulation
```
HerbertJordan, line 1046: Feature question: should we play it safe and only switch to geth if there is a code that is longer than 64KB? This way most requests would still be processed with the LFVM and only the special cases switch to geth.

In this case we would not even add a flag. It will just be the default, without alternative. Wdyt?

config/flags/flags.go
```diff
@@ -330,6 +330,10 @@ var (
 		Name:  "exitwhensynced.epoch",
 		Usage: "Exits after synchronisation reaches the required epoch",
 	}
+	RPCEVMSimulationFlag = cli.BoolFlag{
+		Name:  "rpc.evmsimulation",
+		Usage: "EVM is used for RPC transaction simulation",
```
HerbertJordan, line 335: We should be precise in the terminology here: an EVM is an abstract specification of Ethereum's Virtual Machine which is implemented by `geth` and Tosca through its `lfvm` implementation. 

The description of this flag should thus be something like "if enabled, the geth implementation of the EVM is used in RPC transaction simulation calls";

HerbertJordan (changes requested):
Looks good but I am wondering whether we should even make this feature optional. We could just use the LFVM in all cases except if there is a code override exceeding 64KB -- in which case we fall back to geth. Do you see a use case where somebody would not want that? Wdyt?

If we stick to the flag, I would like to ask to fix the naming. It seems that the term "EVM" is used at many places when actually referring to "geth". This terminology issue should be fixed.

jenikd:
Implemented different solution with PR https://github.com/0xsoniclabs/sonic/pull/189

PR #187, opened 2025-04-16
Add algorithm to select proposer for each turn

HerbertJordan:
This PR introduces the algorithm used to select validators allowed to propose blocks from a list of validators.

The selection algorithm is selecting validators proportionally to their stake. Thus, a valdiator A with twice the stake of validator B will get selected twice as often for being a proposer for turns.

### DIFF
--- a/inter/proposer_selection.go
+++ b/inter/proposer_selection.go
@@ -1,0 +1,65 @@
+package inter
+
+import (
+	"crypto/sha256"
+	"encoding/binary"
+	"fmt"
+	"math/big"
+
+	"github.com/Fantom-foundation/lachesis-base/inter/idx"
+	"github.com/Fantom-foundation/lachesis-base/inter/pos"
+)
+
+// Turn is the turn number of a proposal. Turns are used to orchestrate the
+// sequence of block proposals in the consensus protocol. Turns are processed
+// in order. A turn ends with a proposer making a proposal or a timeout.
+type Turn uint32
+
+// GetProposer returns the designated proposer for a given turn.
+// The proposer is determined through deterministic sampling of validators
+// proportional to the validator's stake.
+func GetProposer(
+	validators *pos.Validators,
+	turn Turn,
+) (idx.ValidatorID, error) {
+
+	// The selection of the proposer for a given round is conducted as follows:
+	//  1. f := sha256(turn) / 2^256
+	//  2. limit := f * total_weight
+	//  3. from the list of validators sorted by their stake, find the first
+	//     validator whose cumulative weight is greater than or equal to limit.
+
+	// -- Preconditions --
+	ids := validators.SortedIDs()
+	if len(ids) == 0 {
+		return 0, fmt.Errorf("no validators")
+	}
+
+	// Note that we use big.Rat to preserve precision in the division.
+	// limit := (sha256(turn) * total_weight) / 2^256
+	data := make([]byte, 0, 4)
+	data = binary.BigEndian.AppendUint32(data, uint32(turn))
+	hash := sha256.Sum256(data)
+	limit := new(big.Rat).Quo(
+		new(big.Rat).SetInt(
+			new(big.Int).Mul(
+				new(big.Int).SetBytes(hash[:]),
+				big.NewInt(int64(validators.TotalWeight())),
+			),
+		),
+		new(big.Rat).SetInt(new(big.Int).Lsh(big.NewInt(1), 256)),
+	)
+
+	// Find the first validator whose cumulative weight is greater than or equal
+	// to the limit.
+	res := ids[0]
+	cumulated := big.NewRat(0, 1)
+	for i, weight := range validators.SortedWeights() {
+		cumulated.Num().Add(cumulated.Num(), big.NewInt(int64(weight)))
+		if cumulated.Cmp(limit) >= 0 {
+			res = ids[i]
+			break
+		}
+	}
+	return res, nil
+}
--- a/inter/proposer_selection.go
+++ b/inter/proposer_selection.go
@@ -50,8 +50,8 @@ func GetProposer(
 		new(big.Rat).SetInt(new(big.Int).Lsh(big.NewInt(1), 256)),
 	)
 
-	// Find the first validator whose cumulative weight is greater than or equal
-	// to the limit.
+	// Walk through the validators sorted by their stake (and ID as a tiebreaker)
+	// and accumulate their weights until we reach the limit calculated above.
 	res := ids[0]
 	cumulated := big.NewRat(0, 1)
 	for i, weight := range validators.SortedWeights() {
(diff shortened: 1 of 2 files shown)


inter/proposer_selection_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetProposer_IsDeterministic(t *testing.T) {
+	require := require.New(t)
+
+	builder := pos.ValidatorsBuilder{}
+	builder.Set(1, 10)
+	builder.Set(2, 20)
+	builder.Set(3, 30)
```
LuisPH3, line 17: Because the stake is computed and may be opaque to this algorithm. Would it make sense to add a test with a member with 0 stake?

LuisPH3 (approved):
The proposed algorithm chooses the first validator from a list of validators shorted by stake, which stake is greater than a modulo operation on the turn.

inter/proposer_selection_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"testing"
+
+	"github.com/Fantom-foundation/lachesis-base/inter/idx"
+	"github.com/Fantom-foundation/lachesis-base/inter/pos"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetProposer_IsDeterministic(t *testing.T) {
```
LuisPH3, line 11: Missing test: because the algorithm chooses the first validator in the sorted list, what happens when two validators have the exact same stake?

inter/proposer_selection_test.go
LuisPH3: I pressure a test is needed, wich checks that both validators are chosen with a 50% chance.

LuisPH3 (changes requested):
I believe there is a missing test.

inter/proposer_selection_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	const Samples = 100000
+	counters := map[idx.ValidatorID]int{}
+	for turn := range Turn(Samples) {
+		proposer, err := GetProposer(validators, turn)
+		require.NoError(err)
+		counters[proposer]++
+	}
+
+	tolerance := float64(Samples / 100) // 1% tolerance
```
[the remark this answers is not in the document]
HerbertJordan: The absolute value is only beg because of the 100k samples. But you are right, the true derivation from the expectation is somewhere at 0.22%, so I set the limit to 0.25%.

I also expanded the test to check the error for various numbers of samples to add more evidence that the error compared to the ideal distribution is indeed converging towards zero.

inter/proposer_selection_test.go
HerbertJordan: The validator builder is sorting validators by their stake and uses the validator IDs as a tie breaker. This defines a deterministic order of the validators in the input set.

The builder is also eliminating validators with zero stack.

inter/proposer_selection_test.go
```diff
@@ -0,0 +1,65 @@
...
+		a, err := GetProposer(validators, turn)
+		require.NoError(err)
+		b, err := GetProposer(validators, turn)
+		require.NoError(err)
+		require.Equal(a, b, "proposer selection is not deterministic")
+	}
+}
+
+func TestGetProposer_EmptyValidatorSet_Fails(t *testing.T) {
+	require := require.New(t)
```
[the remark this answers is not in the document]
HerbertJordan: Yes, I am aware of the free-standing functions, and I am using them from time to time, but in general I consider the version without the extra `t` parameter more readable. Hope that is Ok.

inter/proposer_selection_test.go
```diff
@@ -46,20 +93,38 @@ func TestGetProposer_ProposersAreSelectedProportionalToStake(t *testing.T) {
...
-	for id, idx := range validators.Idxs() {
-		expected := int(Samples * validators.GetWeightByIdx(idx) / validators.TotalWeight())
-		require.InDelta(
-			counters[id], expected, tolerance,
-			"validator %d is not selected proportional to stake", id,
-		)
+	for _, test := range tests {
+		t.Run(fmt.Sprintf("samples=%v", test.samples), func(t *testing.T) {
+			require := require.New(t)
+			t.Parallel()
```
LuisPH3, line 112: nit: t. Parallel linter requires parallel subtest to be children of parallel test. This reduces the amount of stop-resume needed to execute these tests.
(we do not follow this in the integration tests, because our top level needs to create the integration test net, but this test can do it)

github.com/moricho/tparallel

inter/proposer_selection_test.go
```diff
@@ -0,0 +1,130 @@
...
+	for turn := range Turn(5) {
+		a, err := GetProposer(validators, turn)
+		require.NoError(err)
+		b, err := GetProposer(validators, turn)
+		require.NoError(err)
+		require.Equal(a, b, "proposer selection is not deterministic")
+	}
+}
+
+func TestGetProposer_EqualStakes_SelectionIsDeterministic(t *testing.T) {
```
LuisPH3, line 30: I do not see how this test checks that both validators are given equal chances.
What I see is that the 10 first proposers are the same in both executions, but they could all id 1 or id 2. 
can you explain this to me?

inter/proposer_selection_test.go
LuisPH3: See comment in new test, I may still be missing something.

inter/proposer_selection_test.go
HerbertJordan: Right, it does not. It only checks that there is no in-determinism due to the equality of the weights.

The fair-share allocation was implicitly covered by `TestGetProposer_ProposersAreSelectedProportionalToStake`, assuming that if it is working for a non-equal stake it also works for the special case of a uniform stake. This should be done explicitly. I extended the corresponding test case.

LuisPH3 (approved):
Now I get it, cumulative weight. this is how equal stake validators end up being 50:50

inter/proposer_selection.go
```diff
@@ -0,0 +1,65 @@
...
+		new(big.Rat).SetInt(
+			new(big.Int).Mul(
+				new(big.Int).SetBytes(hash[:]),
+				big.NewInt(int64(validators.TotalWeight())),
+			),
+		),
+		new(big.Rat).SetInt(new(big.Int).Lsh(big.NewInt(1), 256)),
+	)
+
+	// Find the first validator whose cumulative weight is greater than or equal
```
simonlechner, line 53: It took me a little bit to be sure that the order of validators with the same stake does not impact how often each of them is chosen. So I am happy to see a test case got added.
Although in this context, there is only one cumulative weight, it would be easier to read if it was clear that a validators cumulative weight is the sum of the weight of all its predecessors and itself.

simonlechner (dismissed):
Looks good.

inter/proposer_selection.go
HerbertJordan: I updated the comment, please have a look.

LuisPH3 (dismissed):

PR #189, opened 2025-04-17
Use GETH for eth_call simulation with large code in state override

jenikd:
This PR modifies `eth_call` RPC function when there is contract code in account state override largen then the maximum allowed code for `LFVM` interpreter. In this case it uses `GETH` interpreter to simulate `eth_call`.

### DIFF
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -1020,7 +1020,7 @@ func (diff *StateOverride) CodeTooBig() bool {
 	}
 	for _, account := range *diff {
 		// Check account(contract) code length.
-		if account.Code != nil && len(*account.Code) > math.MaxUint16 {
+		if account.Code != nil && len(*account.Code) > params.MaxCodeSize {
 			return true
 		}
 	}
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -1014,7 +1014,7 @@ func (diff *StateOverride) Apply(state state.StateDB) error {
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


ethapi/api.go
```diff
@@ -1014,6 +1014,19 @@ func (diff *StateOverride) Apply(state state.StateDB) error {
 	return nil
 }
 
+func (diff *StateOverride) CodeTooBig() bool {
```
HerbertJordan, line 1017: nit: `CodeTooBig` => `HasCodesExceedingOnChainLimit` ?

HerbertJordan (approved):
Looks good to me :)

tests/eth_call_test.go
```diff
@@ -1,26 +1,22 @@
...
 )
 
-func TestEthCall_CodeLargerThanMaxInitCodeSizeIsNotAccepted(t *testing.T) {
+func TestEthCall_CodeLargerThanMaxInitCodeSizeIsAccepted(t *testing.T) {
 	tests := map[string]struct {
 		codeSize int
-		err      error
 	}{
-		"max code size": {
+		"max code size lfvm": {
```
HerbertJordan, line 15: Can you also add two more test cases for a code that has `params.MaxCodeSize` and `params.MaxCodeSize+1`?

The way it is now, both cases are covered by geth.

HerbertJordan (dismissed):

HerbertJordan (approved):

PR #192, opened 2025-04-17
Add `make integration-cover-all` target

facuMH:
This PR adds a new target to our Makefile, `integration-cover-all`. 

This new target will:
- create a new folder (`./coverage/{current-date}`)
- run the integrations tests (`go test ./tests/`) and produce a coverage report on all packages in `github.com/0xsoniclabs/sonic/`.
- the raw report will be in `./coverage/{current-date}/cover.out` and a friendlier human readable html report will be in  `./coverage/{current-date}/coverage.html`

The folder where all reports will be generated is included in `.gitignore` as to prevent uploading them by mistake.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 */*.test
 */*.out
 fuzzing/
+coverage/
 
 # Build output
 build
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,6 @@
 */*.test
 */*.out
 fuzzing/
-coverage/
 
 # Build output
 build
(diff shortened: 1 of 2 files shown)


HerbertJordan (approved):
Thanks for adding this!

LuisPH3:
Could you please generate build byproducts inside the ignored build folder? or ignore the generated paths?  (I am fond of the first one, as other build systems/languages do) 
This would prevent polluting the repository with by-products that need to be manually ignored.

facuMH:
> Could you please generate build byproducts inside the ignored build folder? or ignore the generated paths? (I am fond of the first one, as other build systems/languages do) This would prevent polluting the repository with by-products that need to be manually ignored.

since it is not really a "build", I thought it would be best to have its own folder and add it to the gitignore. I can alternatively not add it to the git ignore and instead drop the reports in `build/coverage/`. @HerbertJordan @wsodsong  let me know if you prefer the coverage reports to be inside build or an independent folder. I have no strong feelings towards one or the other option

LuisPH3 (approved):
thanks!

PR #193, opened 2025-04-17
Add validation of proposer turn progression

HerbertJordan:
This PR implements a function verifying valid turn progression steps. The corresponding documentation can be found in the code.

In short: after every proposal for turn `X`, any subsequent proposal `Y` has a time window of `(d*C,C]` frames to send in a proposal. The first received proposal is accepted, and the timer is restarted from the receiving frame for subsequent proposals.


inter/turns.go
```diff
@@ -0,0 +1,60 @@
...
+	// Turns and frames must strictly increase in each progression step.
+	if last.Turn >= next.Turn || last.Frame >= next.Frame {
+		return false
+	}
+
+	// Every attempt has a window of frames (d*C,(d+1)*C] where d is the number
+	// of turns between the last and the next turn. Thus, the immediate
+	// successor of a turn can be processed in the frames (0,C] = [1,C] after
+	// the last turn. If the successor turn is not materializing, the next turn
+	// can be processed in the frames (C,2*C] = [C+1,2*C] after the last turn.
```
simonlechner, line 43: Not just `d` but also `C` should be given a name and described.

inter/turns_test.go
```diff
@@ -0,0 +1,264 @@
...
+		},
+		"skipped turn long after its own window": {
+			last: S{Turn: 1, Frame: 1},
+			next: S{Turn: 3, Frame: 1 + 2*C + 100},
+			want: false,
+		},
+
+		// -- double-skipped turn proposals --
+		// In these scenarios, turns 2 and 3 have not been consumed, enabling
+		// turn 4 to be used between (2*C, 3*C] frames after the last proposal.
```
simonlechner, line 134: What do you think about a case: "double skipped turn far before being enabled"?

simonlechner (commented):
The testing looks good, some comments regarding the documentation.
As this function only consists of a handful of operations, why is the production version chosen over the test version?

inter/turns.go
LuisPH3: I have this same problem in the attached documentation, 
I cannot read the interval (d*C,(d+1)*C]

inter/turns_test.go
```diff
@@ -0,0 +1,264 @@
...
+			want: true,
+		},
+		"skipped turn just before end of time window": {
+			last: S{Turn: 1, Frame: 1},
+			next: S{Turn: 3, Frame: 1 + 2*C},
+			want: true,
+		},
+		"skipped turn just after end of time window": {
+			last: S{Turn: 1, Frame: 1},
+			next: S{Turn: 3, Frame: 1 + 2*C + 1},
```
LuisPH3, line 123: There  is something that makes formulas very difficult to read,
Usually polimia are ordered by exponents, and incognitas, leaving the constants as the last component. 
I think the intention here was to "illustrate" components, but it just turns difficult to read.

inter/turns_test.go
```diff
@@ -0,0 +1,264 @@
...
+	if last.Turn >= next.Turn {
+		return false
+	}
+	delta := uint64(next.Turn - last.Turn - 1)
+	min := uint64(last.Frame) + delta*TurnTimeoutInFrames
+	max := min + TurnTimeoutInFrames
+	return min < uint64(next.Frame) && uint64(next.Frame) <= max
+}
+
+func FuzzIsValidTurnProgression(f *testing.F) {
```
LuisPH3, line 221: This test is superfluous, given the existence of TestIsValidTurnProgression_ExhaustiveTests

inter/turns_test.go
```diff
@@ -0,0 +1,264 @@
...
+	for range b.N {
+		last = IsValidTurnProgression(
+			ProposalSummary{Turn: 1, Frame: 1},
+			ProposalSummary{Turn: validTurnBenchInput, Frame: 2},
+		)
+	}
+	validTurnBenchResult = last // needed to avoid compiler optimization
+}
+
+func BenchmarkIsValidTurnProgression_Tests(b *testing.B) {
```
LuisPH3, line 255: superfluous.

inter/turns_test.go
```diff
@@ -0,0 +1,264 @@
...
+	"github.com/Fantom-foundation/lachesis-base/inter/idx"
+)
+
+func TestIsValidTurnProgression_ExampleCasesAndResults(t *testing.T) {
+	type S = ProposalSummary
+	const C = TurnTimeoutInFrames
+	tests := map[string]struct {
+		last ProposalSummary
+		next ProposalSummary
+		want bool
```
LuisPH3, line 15:
```suggestion
		isValid bool
```

inter/turns_test.go
```diff
@@ -0,0 +1,264 @@
...
+				}
+			}
+		}
+	}
+}
+
+// isValidTurnProgressionForTests is an alternative implementation of the function to
+// compare the results with the production version. It is intended to be a
+// reference implementations in tests.
+func isValidTurnProgressionForTests(
```
LuisPH3, line 206: I have to say that first i found this method "funny". 
I would have preferred to have a list of declarative properties of what makes a turn "valid".
If declared valid: 
- last.Turn < next.Turn
- last.Frame < next.Frame
- and the difference between turns matches the expected count.
If not, or the opposite.
Once factoring everything it in, it turns to be the code used for the original implementation.

LuisPH3 (approved):
Some minor cosemetic comments:

inter/turns.go
HerbertJordan: Should I explain the interval notation or rewrite it to not use the interval notation?

HerbertJordan: I rewrote it without intervals.

inter/turns_test.go
HerbertJordan: renamed to "valid"

inter/turns_test.go
HerbertJordan: The format intended here is
`<the frame of the last turn> + <offset written as a polynome>`

The first "1+" is always just the "last frame" prefix. I added parentheses around the offset polynome where it is composed of more than one term.

inter/turns_test.go
HerbertJordan: The exhaustive test is badly named. It is not really exhaustive, as there are 2^128 combinations to be tested. I renamed the test to an "_EnumerationTest".

The fuzzer bases test is still justified for detecting overflow errors. In an earlier version I had an overflow such that a turn 0 could be followed by a turn 1 or a much larger turn number that was treated like 1 due to overflows.

inter/turns_test.go
HerbertJordan: Actually, this benchmark lead me to switch the implementations of the productive version and the test version as the original test version was faster than the original productive version.

inter/turns_test.go
HerbertJordan: Added.

inter/turns_test.go
HerbertJordan: So ... I guess this is settled?

LuisPH3 (dismissed):

PR #194, opened 2025-04-17
Introduce new event version using protobuffer payload

HerbertJordan:
This PR integrates the new `Payload` type into a new version for the Lachesis `Event` type.

This modification adds support for proposals to be annotated to events, thereby enabling the forwarding of proposals through the network as well as retaining them in the event database.

### DIFF
--- a/inter/event.go
+++ b/inter/event.go
@@ -58,6 +58,7 @@ type EventPayloadI interface {
 	EpochVote() LlrEpochVote
 	BlockVotes() LlrBlockVotes
 	MisbehaviourProofs() []MisbehaviourProof
+	Payload() *Payload
 }
 
 var emptyPayloadHash1 = CalcPayloadHash(&MutableEventPayload{extEventData: extEventData{version: 1}})
@@ -65,6 +66,8 @@ var emptyPayloadHash1 = CalcPayloadHash(&MutableEventPayload{extEventData: extEv
 func EmptyPayloadHash(version uint8) hash.Hash {
 	if version == 1 {
 		return emptyPayloadHash1
+	} else if version == 3 {
+		return (&Payload{}).Hash()
 	} else {
 		return hash.Hash(types.EmptyRootHash)
 	}
@@ -105,6 +108,8 @@ type payloadData struct {
 
 	epochVote  LlrEpochVote
 	blockVotes LlrBlockVotes
+
+	payload Payload
 }
 
 type Event struct {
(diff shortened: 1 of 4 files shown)


inter/event.go
```diff
@@ -58,13 +58,16 @@ type EventPayloadI interface {
...
+	Payload() *Payload
 }
 
 var emptyPayloadHash1 = CalcPayloadHash(&MutableEventPayload{extEventData: extEventData{version: 1}})
 
 func EmptyPayloadHash(version uint8) hash.Hash {
 	if version == 1 {
 		return emptyPayloadHash1
+	} else if version == 3 {
+		return (&Payload{}).Hash()
```
LuisPH3, line 70: For coherence, use `CalcPayloadHash` as well. 
either inline `emptyPayloadHash1` or create `emptyPayloadHash2`

inter/event.go
```diff
@@ -209,9 +218,11 @@ func CalcMisbehaviourProofsHash(mps []MisbehaviourProof) hash.Hash {
 func CalcPayloadHash(e EventPayloadI) hash.Hash {
 	if e.Version() == 1 {
 		return hash.Of(hash.Of(CalcTxHash(e.Txs()).Bytes(), CalcMisbehaviourProofsHash(e.MisbehaviourProofs()).Bytes()).Bytes(), hash.Of(e.EpochVote().Hash().Bytes(), e.BlockVotes().Hash().Bytes()).Bytes())
-	} else {
-		return CalcTxHash(e.Txs())
 	}
+	if e.Version() == 3 {
+		return e.Payload().Hash()
```
LuisPH3, line 223: Missing hash of `MisbehaviourProofs` `EpochVote` `BlockVotes` members of the event struct. 
`EventPayloadI` is a superset that includes more data than the new `Payload` object.

LuisPH3 (commented):
These changes are not easy to follow, mixing the old infrastructure with the new one is awkward. 
Some comments about the content of hashes.

inter/event.go
HerbertJordan: Nice, done!

inter/event.go
HerbertJordan: MisbehaviourProofs, EpochVotes, and BlockVotes are not used in Sonic (these are part of the LLR feature), and the plan is to add an event evaluation step that requires that all events with version 3 have these fields (and the Tx field) empty.

These constraints seem to be enforced right now in the serializer code [here](https://github.com/0xsoniclabs/sonic/blob/64bfcf99a2070d55638506b1a32270ce88d7ddc0/inter/event_serializer.go#L242-L258). I do not think this is the best place to check it, but I guess we should follow the precedence. So I added the check there as well.

LuisPH3:
> Any alternative you can think of?

just ranting about the protocol buffer change, I am conflicted, but it should make our life way easier down the line.

PR #195, opened 2025-04-18
Introduce interface to run transactions in blocks incrementally

HerbertJordan:
This PR extends the `StateProcessor` implementation to support the incremental execution of transactions. This is required by the emitter to build up proposals incrementally. In particular, the Emitter needs to know for each transaction individually whether it can be processed or not, and its cause (skipped or error).

Before this change, the `StateProcessor` had a single `Process` function, accepting a list of transactions and producing receipts, logs, a list of skipped transactions, and potentially an error. Internally, it was running some setup code (e.g. creation of  an EVM environment, updating the parent-block-hash) followed by running the list of transactions in a loop.

The two steps of the `Process` function have been split by this PR. Now there is a `BeginBlock` function handling the setup operations and returning a `TransactionProcessor` which is able to run individual transactions incrementally. 

Due to the criticality of the `Process` function for the network consensus, the previous implementation has been preserved. To verify the alignment of the `BeginBlock` and `TransactionProcessor` with the pre-existing `Process` function, a `process_incremental` method is implemented in tests based on the `BeginBlock` and `TransactionProcessor` infrastructure to facilitate a side-by-side comparison.


evmcore/state_processor.go
```diff
@@ -103,6 +99,113 @@ func (p *StateProcessor) Process(
 	return
 }
 
+// Process processes the state changes according to the Ethereum rules by running
```
LuisPH3, line 102: This docstring, both paragraphs were verbatim copied from Ethereum https://github.com/ethereum/go-ethereum/blob/846c14e21f5f7ceffe47898a34a520895b1b0f11/core/state_processor.go#L50
And it does not reflect the changes done for the fantom/sonic use.
namely rewards for coinbase and uncles. and lack of description about skipped txs.

evmcore/state_processor.go
```diff
@@ -103,6 +99,113 @@ func (p *StateProcessor) Process(
...
+// processing.
+func (tp *TransactionProcessor) Run(i int, tx *types.Transaction) (
+	receipt *types.Receipt,
+	skipped bool,
+	err error,
+) {
+	msg, err := TxAsMessage(tx, tp.signer, tp.header.BaseFee)
+	if err != nil {
+		return nil, false, fmt.Errorf(
+			"could not apply tx %d [%v]: %w", i, tx.Hash().Hex(), err,
```
LuisPH3, line 198: nit: for coherence with function name: "failed to run tx ...."
(Potentially: "failed to run tx ...., conversion to message failed")

evmcore/state_processor.go
```diff
@@ -225,8 +328,10 @@ func applyTransaction(
 	// contract listener, only the sender, topics, and the data are relevant.
 	// The block hash is not used.
 	logs := statedb.GetLogs(tx.Hash(), common.Hash{})
-	for _, l := range logs {
```
LuisPH3, line 228: How would this not fail before? 
the more I read and compare with geth, the less I like this onNewLog callback. Maybe we should consider cleaning it up.

evmcore/state_processor.go
LuisPH3: becase this error gets wrapped into ""could not apply tx %d [%v]: %w" on the caller side, I would just write here "failed to convert tx to message"

evmcore/state_processor.go
```diff
@@ -103,6 +99,113 @@ func (p *StateProcessor) Process(
...
+	for i, tx := range block.Transactions {
+		receipt, skip, err := txProcessor.Run(i, tx)
+		if skip {
+			skipped = append(skipped, uint32(i))
+			receipts[i] = nil
+			err = nil
+			continue
+		}
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("could not apply tx %d [%v]: %w", i, tx.Hash().Hex(), err)
```
LuisPH3, line 129:
```suggestion
			return nil, nil, nil, fmt.Errorf("failed to process transaction tx %d [%v]: %w", i, tx.Hash().Hex(), err)
```

evmcore/state_processor.go
```diff
@@ -103,6 +99,113 @@ func (p *StateProcessor) Process(
...
+	stateDb       state.StateDB
+	usedGas       uint64
+	vmEnvironment *vm.EVM
+}
+
+// Run processes a single transaction in the block, where i is the index of
+// the transaction in the block. It returns the receipt of the transaction,
+// whether the transaction was skipped, and any error that occurred during
+// processing.
+func (tp *TransactionProcessor) Run(i int, tx *types.Transaction) (
```
LuisPH3, line 190: I think this function could be replaced with calls to TxAsMessage and ApplyTransaction in the parent context.
The reasons are:
- run does not add any semantic value to the algorithm, wich already has process, apply (and maybe execute) worlds.
- The error reporting, as described in other comments, is cumbersome. Apply gets nested twice.
- It is not quite clear to me why a tx needs to be converted into a message before execution. 
- the algorithm would read better if txs are converted in the loop before executed.

evmcore/state_processor_test.go
```diff
@@ -49,3 +257,65 @@ func TestApplyTransaction_InternalTransactionsSkipBaseFeeCharges(t *testing.T) {
...
+	state.EXPECT().Prepare(any, any, any, any, any, any).AnyTimes()
+	state.EXPECT().GetNonce(any).AnyTimes()
+	state.EXPECT().SetNonce(any, any, any).AnyTimes()
+	state.EXPECT().GetCodeHash(any).Return(types.EmptyCodeHash).AnyTimes()
+	state.EXPECT().GetCode(any).AnyTimes()
+	state.EXPECT().GetStorageRoot(any).Return(types.EmptyRootHash).AnyTimes()
+	state.EXPECT().Snapshot().AnyTimes()
+	state.EXPECT().Exist(any).Return(true).AnyTimes()
+	state.EXPECT().GetRefund().AnyTimes()
+	state.EXPECT().EndTransaction().AnyTimes()
```
LuisPH3, line 319: My gut feeling is that a test testing the block processor should know the minimum number of calls to these callbacks are invoked and with which arguments. For some of them an exact value.
i.e: get nonce and set nonce should only happen once per tx (legacy)  for the sender address. 
prepare, snapshot, end transaction, finalize.... should only be called once each.

LuisPH3 (commented):
The proposed changes make a semantic partition of the steps required to execute a block.
The code generated is clear, but there are some issues with the error propagation. 

Regarding correctness, the algorithm seems correct, unit testing seems insufficient, maybe a complete list of skipped tx reasons should be needed.
In overall, the changes will be tested with a simple network sync. Under the assumption that  every skip reason has happened already in the chain history.

evmcore/state_processor.go
HerbertJordan: > How would this not fail before?

There was never a caller with a `nil` before. I basically added this to make testing simpler, which, as a side effect, makes the code also more robust.

> The more I read and compare with geth, the less I like this onNewLog callback. Maybe we should consider cleaning it up.

The problem I see is the danger of accidentally introducing something that breaks the full backward compatibility. This code is part of the replicated processing required to be identical among all nodes on the network. The PR as it is attempts to minimize the changes to be bear minimium.

LuisPH3 (approved):
The changes introduce a new block processor loop with incremental features. 
This code complements the existing "bulk" processor, and testing is provided to add guarantees on the compatibility with each other.

PR #197, opened 2025-04-21
Add `ValidateTxWithNetworkRules`

facuMH:
This PR does the first of a series of refactors in `validateTx` from the `TxPool`. 

In order to isolate and tests the verifications related to the network rules, the previous `validateOptions` is split into `validateOptions` and `networkOptions`. This way it can be individually tested:
- Revision and tx types compatibility (including revision dependent checks such as max init code size, empty blobs and non-empty auth list)
- GasPrice (GasFee) over network's BaseFee
- Gas less or equal than current max gas in the block

Note to reviewer:  changes to `tx_pool_test.go` and `tests/blob_tx_test.go` are only necessary because now the check for tx size is done before checking for tx type.

### DIFF
--- a/evmcore/tx_validation.go
+++ b/evmcore/tx_validation.go
@@ -28,9 +28,9 @@ import (
 	"github.com/ethereum/go-ethereum/params"
 )
 
-// validationOptions is a set of options to adjust the validation of transactions
-// according to the current state of the transaction pool.
-type validationOptions struct {
+// networkOptions defines the set of active network rules and parameters
+// used to validate transactions based on the current fork state.
+type networkOptions struct {
 	istanbul bool // Fork indicator whether we are in the istanbul revision.
 	shanghai bool // Fork indicator whether we are in the shanghai revision.
 
@@ -40,10 +40,16 @@ type validationOptions struct {
 	eip7623 bool // Fork indicator whether we are using EIP-7623 floor gas validation.
 	eip7702 bool // Fork indicator whether we are using EIP-7702 set code transactions.
 
-	currentState   TxPoolStateDB // Current state in the blockchain head
-	currentMaxGas  uint64        // Current gas limit for transaction caps
-	currentBaseFee *big.Int      // Current base fee for transaction caps
-	minTip         *big.Int      // Minimum gas tip to enforce for acceptance into the pool
+	currentBaseFee *big.Int // Current base fee for transaction caps
+	currentMaxGas  uint64   // Current gas limit for transaction caps
+}
+
+// validationOptions is a set of options to adjust the validation of transactions
+// according to the current state of the transaction pool.
+type validationOptions struct {
+	currentState TxPoolStateDB // Current state in the blockchain head
+
+	minTip *big.Int // Minimum gas tip to enforce for acceptance into the pool
 
 	locals  *accountSet // Set of local transaction to exempt from eviction rules
 	isLocal bool        // Whether the transaction came from a local source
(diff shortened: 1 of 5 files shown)


evmcore/tx_validation.go
```diff
@@ -28,9 +28,9 @@ import (
 	"github.com/ethereum/go-ethereum/params"
 )
 
-// validationOptions is a set of options to adjust the validation of transactions
-// according to the current state of the transaction pool.
-type validationOptions struct {
+// networkOptions defines the set of active network rules and parameters
+// used to validate transactions based on the current fork state.
+type networkOptions struct {
```
HerbertJordan, line 33: nit: these are not options for any network ... so the name is not very helpful. Consider including the fact that these are still options for the transaction validation into the type name.

evmcore/tx_validation.go
```diff
@@ -53,7 +59,80 @@ type validationOptions struct {
...
+//   - if setCode transactions are supported, they have a non-empty authorization list
+//   - if the transaction is a contract creation, it doesn't exceed the maximum
+//     init code size
+//   - the transaction's gas limit is not greater than the current block gas limit
+//   - the transaction's gas fee cap is not less than the minimum base fee
+//   - the transaction's gas limit is not less than the intrinsic gas
+//   - the transaction's gas limit is not less than the floor data gas
+//
+// The function returns an error if any of the checks fail and nil otherwise.
+func ValidateTxWithNetworkRules(tx *types.Transaction, opt networkOptions) error {
```
HerbertJordan, line 135: rename: "ValidateTxWithNetworkRules" => "ValidateTxForNetworkRules"?

tests/blob_tx_test.go
```diff
@@ -49,11 +49,11 @@ func testBlobTx_WithBlobsIsRejected(t *testing.T, ctxt *testContext) {
 
 	// attempt to run tx
 	_, err = ctxt.net.Run(tx)
-	require.ErrorContains(err, "transaction type not supported")
+	require.ErrorContains(err, "oversized data")
```
HerbertJordan, line 52: This change looks like it is breaking the previous intention of checking that the transaction type is rejected. The text indicates that now a different reason is causing the rejection, shadowing the originally check tested for.

It would be good to figure out why no suddenly the data is considered to big.

evmcore/tx_validation_test.go
```diff
@@ -16,68 +16,71 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// getTestTransactionsOption returns a set of options to adjust the validation of transactions
+// so that it would accept all types of transactions, considering them as local transactions
+// with a min tip of 1, current base fee of 1, and a current max gas of 100_000.
+func getTestTransactionsOption() (validationOptions, networkOptions) {
```
HerbertJordan, line 22: "getTestTransactionsOption" => "getTestValidationOption"

evmcore/tx_validation_test.go
```diff
@@ -16,68 +16,71 @@ import (
...
 		if _, ok := tx.(*types.LegacyTx); ok {
 			continue // Skip legacy transactions
 		}
 		t.Run(name, func(t *testing.T) {
-			err := validateTx(types.NewTx(tx), validationOptions{eip2718: false})
-			require.ErrorIs(t, ErrTxTypeNotSupported, err)
+			opt, netOpts := getTestTransactionsOption()
+			netOpts.eip2718 = false
+			err := validateTx(types.NewTx(tx), opt, netOpts)
+			require.ErrorContains(t, err, "network validation failed")
```
HerbertJordan, line 48: This seems like now the test is accepting any reason for failing the validation while before the test was specifically targeting the "ErrTxTypeNotSupported" issue.

evmcore/tx_validation_test.go
```diff
@@ -86,20 +89,23 @@ func TestValidateTx_RejectsTxBasedOnTypeAndActiveRevision(t *testing.T) {
 	}
 	for name, test := range tests {
 		t.Run(name, func(t *testing.T) {
-			err := validateTx(test.tx, test.configure(getTestTransactionsOption()))
-			require.Equal(t, ErrTxTypeNotSupported, err)
+			validateOpts, netOpts := getTestTransactionsOption()
+			err := validateTx(test.tx, validateOpts, test.configure(netOpts))
+			require.ErrorContains(t, err, "network validation failed")
```
HerbertJordan, line 94: Are you sure the test condition for what has caused the test to fail should be so relaxed? It seems to defeat the specificity of the tests.

evmcore/tx_validation_test.go
```diff
@@ -248,26 +259,26 @@ func TestValidateTx_GasPriceAndTip_RejectsTxWith(t *testing.T) {
...
-			opt.currentMaxGas = 1
+			opt, netOpts := getTestTransactionsOption()
+			netOpts.currentMaxGas = 1
 
 			setGas(t, tx, 2)
 
-			err := validateTx(types.NewTx(tx), opt)
-			require.ErrorIs(t, err, ErrGasLimit)
+			err := validateTx(types.NewTx(tx), opt, netOpts)
+			require.ErrorContains(t, err, "network validation failed")
```
HerbertJordan, line 268: Please revert all changes where you have been transitioning from a specific error to the generic "network validation failed" error, as this significantly weakens the whole test setup.

HerbertJordan (changes requested):

facuMH:
this PR will be reworked to introduce 3 helper functions testing different aspects of the tx.

PR #198, opened 2025-04-23
base fee deep copy allegro

kjezek:
This PR updates copy of `MinBaseFee` to be deep copied for `Allegro` update only.  This prevent situation when a network runs pre-Allegro updated, but already running old and new version of the client. 

To prevent network issues, this enables deep copy of this property only when Allegro is enabled, which is possible only when majority of nodes is updated as well.


LuisPH3 (approved):
The changes proposed modifies the copy of the `Rules.Economy.MinBaseFee` field and makes it conditional to the activation of the allegro fork. 
This change prevents changing the network rules semantics before the allegro fork.

HerbertJordan (dismissed):
Thanks for identifying and fixing this!

LuisPH3 (approved):
conflicts fixed

PR #200, opened 2025-04-24
Split `validateTx` in `TxPool` 

facuMH:
This PR refactors the `validateTx` function by breaking it into smaller functions, each grouped according to the type of validation and its dependencies. The goal is to make the validation logic more modular and reusable across packages—even in scenarios where some context (e.g., state, network rules) might be unavailable or irrelevant.

Network validations are performed first, because it is important to report unsupported transaction type as soon as possible. 
After network, validations are ordered to prioritize performance: cheaper checks are performed first, while more costly operations (like state access or complex computations) are deferred until later, and only executed if necessary.

New Validation Functions
These are the four introduced functions, listed in the order they are called:
- `ValidateTxForNetworkRules`: Validates the transaction against the current network rules, including fork-specific behaviors, gas limits, and fee caps.
- `ValidateTxStatic`: Performs checks that rely solely on data within the transaction itself (e.g., size, basic field validity).
- `ValidateTxForBlock`: Verifies that the transaction has `Gas` and `GasFee` compliant to the current block's `MaxGas` and `BaseFee` respectively.
- `validateTxForPool`:  Applies transaction pool-specific policies, such as minimum tip requirements for non-local senders.
- `ValidateTxForState`: Verifies that the sender has a sufficient balance to cover the cost and that the transaction nonce is acceptable based on the current state.

Notes for Reviewers: Changes to `tx_pool_test.go` and `blob_tx_test.go` were required because the transaction size check is now performed earlier—before the transaction type is validated.

### DIFF
--- a/evmcore/tx_pool.go
+++ b/evmcore/tx_pool.go
@@ -673,10 +673,11 @@ func (pool *TxPool) local() map[common.Address]types.Transactions {
 // validateTx checks whether a transaction is valid according to the consensus
 // rules and adheres to some heuristic limits of the local node (price and size).
 func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
-	opts := validationOptions{currentState: pool.currentState,
-		minTip:  pool.gasPrice,
-		locals:  pool.locals,
-		isLocal: local,
+	opts := validationOptions{
+		currentState: pool.currentState,
+		minTip:       pool.gasPrice,
+		locals:       pool.locals,
+		isLocal:      local,
 	}
 	netRules := NetworkRulesForValidateTx{
 		istanbul:       pool.istanbul,
--- a/evmcore/tx_pool.go
+++ b/evmcore/tx_pool.go
@@ -673,7 +673,12 @@ func (pool *TxPool) local() map[common.Address]types.Transactions {
 // validateTx checks whether a transaction is valid according to the consensus
 // rules and adheres to some heuristic limits of the local node (price and size).
 func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
-	opts := validationOptions{
+	opts := validationOptions{currentState: pool.currentState,
+		minTip:  pool.gasPrice,
+		locals:  pool.locals,
+		isLocal: local,
+	}
+	netRules := NetworkRulesForValidateTx{
 		istanbul:       pool.istanbul,
 		shanghai:       pool.shanghai,
 		eip1559:        pool.eip1559,
@@ -681,15 +686,11 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
 		eip4844:        pool.eip4844,
 		eip7623:        pool.eip7623,
 		eip7702:        pool.eip7702,
-		currentState:   pool.currentState,
 		currentMaxGas:  pool.currentMaxGas,
 		currentBaseFee: pool.chain.GetCurrentBaseFee(),
-		minTip:         pool.gasPrice,
-		locals:         pool.locals,
-		isLocal:        local,
 		signer:         pool.signer,
 	}
-	err := validateTx(tx, opts)
+	err := validateTx(tx, opts, netRules)
 	if err != nil {
 		return err
 	}
--- a/evmcore/tx_pool.go
+++ b/evmcore/tx_pool.go
@@ -673,7 +673,7 @@ func (pool *TxPool) local() map[common.Address]types.Transactions {
 // validateTx checks whether a transaction is valid according to the consensus
 // rules and adheres to some heuristic limits of the local node (price and size).
 func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
-	opts := validationOptions{
+	opts := poolOptions{
 		currentState: pool.currentState,
 		minTip:       pool.gasPrice,
 		locals:       pool.locals,
@@ -683,7 +683,7 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
 		maxGas:  pool.currentMaxGas,
 		baseFee: pool.chain.GetCurrentBaseFee(),
 	}
-	netRules := ActiveEips{
+	netRules := NetworkRules{
 		istanbul: pool.istanbul,
 		shanghai: pool.shanghai,
 		eip1559:  pool.eip1559,
@@ -691,6 +691,7 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
 		eip4844:  pool.eip4844,
 		eip7623:  pool.eip7623,
 		eip7702:  pool.eip7702,
+		signer:   pool.signer,
 	}
 	err := validateTx(tx, opts, blockState, netRules)
 	if err != nil {
--- a/evmcore/tx_pool.go
+++ b/evmcore/tx_pool.go
@@ -679,19 +679,20 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
 		locals:       pool.locals,
 		isLocal:      local,
 	}
-	netRules := NetworkRulesForValidateTx{
-		istanbul:       pool.istanbul,
-		shanghai:       pool.shanghai,
-		eip1559:        pool.eip1559,
-		eip2718:        pool.eip2718,
-		eip4844:        pool.eip4844,
-		eip7623:        pool.eip7623,
-		eip7702:        pool.eip7702,
-		currentMaxGas:  pool.currentMaxGas,
-		currentBaseFee: pool.chain.GetCurrentBaseFee(),
-		signer:         pool.signer,
+	blockState := blockState{
+		maxGas:  pool.currentMaxGas,
+		baseFee: pool.chain.GetCurrentBaseFee(),
 	}
-	err := validateTx(tx, opts, netRules)
+	netRules := ActiveEips{
+		istanbul: pool.istanbul,
+		shanghai: pool.shanghai,
+		eip1559:  pool.eip1559,
+		eip2718:  pool.eip2718,
+		eip4844:  pool.eip4844,
+		eip7623:  pool.eip7623,
+		eip7702:  pool.eip7702,
+	}
+	err := validateTx(tx, opts, blockState, netRules)
 	if err != nil {
 		return err
 	}
--- a/evmcore/tx_pool_test.go
+++ b/evmcore/tx_pool_test.go
@@ -497,9 +497,9 @@ func TestEIP4844Transactions(t *testing.T) {
 		err    error
 	}{
 		{"empty blob tx before cancun", nil, false, ErrTxTypeNotSupported},
-		{"blob tx before cancun", common.Address{1}.Bytes(), false, ErrTxTypeNotSupported},
+		{"blob tx before cancun", common.Address{1}.Bytes(), false, ErrOversizedData},
 		{"empty blob tx", nil, true, nil},
-		{"blob tx with data", common.Address{1}.Bytes(), true, ErrTxTypeNotSupported},
+		{"blob tx with data", common.Address{1}.Bytes(), true, ErrOversizedData},
 	}
 
 	for _, test := range tests {
@@ -872,20 +872,19 @@ func TestInvalidTransactions(t *testing.T) {
 	tx := transaction(0, 100, key)
 	from, _ := deriveSender(tx)
 
-	testAddBalance(pool, from, big.NewInt(1))
-	if err := pool.AddRemote(tx); !errors.Is(err, ErrInsufficientFunds) {
-		t.Error("expected", ErrInsufficientFunds)
-	}
-
-	balance := new(big.Int).Add(tx.Value(), new(big.Int).Mul(new(big.Int).SetUint64(tx.Gas()), tx.GasPrice()))
-	testAddBalance(pool, from, balance)
 	if err := pool.AddRemote(tx); !errors.Is(err, ErrIntrinsicGas) {
 		t.Error("expected", ErrIntrinsicGas, "got", err)
 	}
 
+	tx = transaction(0, 100000, key)
+	testAddBalance(pool, from, big.NewInt(1))
+	if err := pool.AddRemote(tx); !errors.Is(err, ErrInsufficientFunds) {
+		t.Errorf("expected %v, but got: %v", ErrInsufficientFunds, err)
+	}
+
 	testSetNonce(pool, from, 1)
 	testAddBalance(pool, from, big.NewInt(0xffffffffffffff))
-	tx = transaction(0, 100000, key)
+
 	if err := pool.AddRemote(tx); !errors.Is(err, ErrNonceTooLow) {
 		t.Error("expected", ErrNonceTooLow)
 	}
--- a/evmcore/tx_pool_test.go
+++ b/evmcore/tx_pool_test.go
@@ -549,7 +549,7 @@ func TestEIP7702Transactions_InvalidTransactionsReturnAnError(t *testing.T) {
 		expectedErr    error
 	}{
 		"set code tx before prague": {
-			expectedErr: ErrTxTypeNotSupported,
+			expectedErr: ErrEmptyAuthorizations,
 		},
 		"set code tx with nil authorizations": {
 			pragueTime:  new(uint64),
--- a/evmcore/tx_pool_test.go
+++ b/evmcore/tx_pool_test.go
@@ -497,9 +497,9 @@ func TestEIP4844Transactions(t *testing.T) {
 		err    error
 	}{
 		{"empty blob tx before cancun", nil, false, ErrTxTypeNotSupported},
-		{"blob tx before cancun", common.Address{1}.Bytes(), false, ErrOversizedData},
+		{"blob tx before cancun", common.Address{1}.Bytes(), false, ErrNonEmptyBlobTx},
 		{"empty blob tx", nil, true, nil},
-		{"blob tx with data", common.Address{1}.Bytes(), true, ErrOversizedData},
+		{"blob tx with data", common.Address{1}.Bytes(), true, ErrNonEmptyBlobTx},
 	}
 
 	for _, test := range tests {
--- a/evmcore/tx_pool_test.go
+++ b/evmcore/tx_pool_test.go
@@ -497,7 +497,7 @@ func TestEIP4844Transactions(t *testing.T) {
 		err    error
 	}{
 		{"empty blob tx before cancun", nil, false, ErrTxTypeNotSupported},
-		{"blob tx before cancun", common.Address{1}.Bytes(), false, ErrNonEmptyBlobTx},
+		{"blob tx before cancun", common.Address{1}.Bytes(), false, ErrTxTypeNotSupported},
 		{"empty blob tx", nil, true, nil},
 		{"blob tx with data", common.Address{1}.Bytes(), true, ErrNonEmptyBlobTx},
 	}
@@ -549,7 +549,7 @@ func TestEIP7702Transactions_InvalidTransactionsReturnAnError(t *testing.T) {
 		expectedErr    error
 	}{
 		"set code tx before prague": {
-			expectedErr: ErrEmptyAuthorizations,
+			expectedErr: ErrTxTypeNotSupported,
 		},
 		"set code tx with nil authorizations": {
 			pragueTime:  new(uint64),
@@ -972,7 +972,7 @@ func TestTransactionNegativeValue(t *testing.T) {
 	pool, key := setupTxPool()
 	defer pool.Stop()
 
-	tx, _ := types.SignTx(types.NewTransaction(0, common.Address{}, big.NewInt(-1), 100, big.NewInt(1), nil), types.HomesteadSigner{}, key)
+	tx, _ := types.SignTx(types.NewTransaction(0, common.Address{}, big.NewInt(-1), 200_000, big.NewInt(1), nil), types.HomesteadSigner{}, key)
 	from, _ := deriveSender(tx)
 	testAddBalance(pool, from, big.NewInt(1))
 	if err := pool.AddRemote(tx); err != ErrNegativeValue {
@@ -986,7 +986,7 @@ func TestTransactionTipAboveFeeCap(t *testing.T) {
 	pool, key := setupTxPoolWithConfig(eip1559Config)
 	defer pool.Stop()
 
-	tx := dynamicFeeTx(0, 100, big.NewInt(1), big.NewInt(2), key)
+	tx := dynamicFeeTx(0, 200_000, big.NewInt(1), big.NewInt(2), key)
 
 	if err := pool.AddRemote(tx); err != ErrTipAboveFeeCap {
 		t.Error("expected", ErrTipAboveFeeCap, "got", err)
@@ -1002,12 +1002,12 @@ func TestTransactionVeryHighValues(t *testing.T) {
 	veryBigNumber := big.NewInt(1)
 	veryBigNumber.Lsh(veryBigNumber, 300)
 
-	tx := dynamicFeeTx(0, 100, big.NewInt(1), veryBigNumber, key)
+	tx := dynamicFeeTx(0, 200_000, big.NewInt(1), veryBigNumber, key)
 	if err := pool.AddRemote(tx); err != ErrTipVeryHigh {
 		t.Error("expected", ErrTipVeryHigh, "got", err)
 	}
 
-	tx2 := dynamicFeeTx(0, 100, veryBigNumber, big.NewInt(1), key)
+	tx2 := dynamicFeeTx(0, 200_000, veryBigNumber, big.NewInt(1), key)
 	if err := pool.AddRemote(tx2); err != ErrFeeCapVeryHigh {
 		t.Error("expected", ErrFeeCapVeryHigh, "got", err)
 	}
--- a/evmcore/tx_validation.go
+++ b/evmcore/tx_validation.go
@@ -62,12 +62,15 @@ func validateTx(
 	opt validationOptions,
 	netRules NetworkRulesForValidateTx) error {
 
-	err := ValidateTxForNetworkRules(tx, netRules)
-	if err != nil {
+	if err := ValidateTxType(tx, netRules); err != nil {
 		return err
 	}
 
 	if err := ValidateTxStatic(tx); err != nil {
+		return err
+	}
+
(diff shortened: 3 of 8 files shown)


LuisPH3 (commented):
I think that the first check should always be if the transaction type is accepted at this block height. The reason for it is that it will always be needed and could save a lot of arithmetic operations if not in range. 
Additionally, this would allow moving Static checks from specific tx types into the static section.

I would like to read the document describing the validation before continuing reviewing this PR.

evmcore/tx_pool.go
```diff
@@ -674,22 +674,24 @@ func (pool *TxPool) local() map[common.Address]types.Transactions {
 // rules and adheres to some heuristic limits of the local node (price and size).
 func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
 	opts := validationOptions{
+		currentState: pool.currentState,
+		minTip:       pool.gasPrice,
+		locals:       pool.locals,
+		isLocal:      local,
+	}
+	netRules := NetworkRulesForValidateTx{
```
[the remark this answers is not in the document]
facuMH: The definition of "which block number activates a given revision" is part of the chain configuration—this is why I categorized these checks as `netRules`. While the last block is needed to determine the current revision, it's only used to compare against predefined values from the config. Similarly, the `BaseFee` is derived from the latest block, but it's provided via the chain state reader, this is why I included it in this set of checks.

evmcore/tx_validation.go
```diff
@@ -189,3 +228,41 @@ func validateTx(tx *types.Transaction, opt validationOptions) error {
...
+// and that the transaction's nonce is not lower than the sender's nonce in the state.
+// Returns an error if any of these conditions are not met.
+func ValidateTxForState(tx *types.Transaction, state TxPoolStateDB, from common.Address) error {
+	// Ensure the transaction adheres to nonce ordering
+	if state.GetNonce(from) > tx.Nonce() {
+		return ErrNonceTooLow
+	}
+
+	// Transactor should have enough funds to cover the costs
+	// cost == V + GP * GL
```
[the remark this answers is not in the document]
facuMH: in some contexts the transaction's `Gas` field is referred to as `GasLimit` because it is compared against the `MaxGas`, which is the gas limit. I believe this is reference to that, but I will change it to just "G" for better clarity.

facuMH:
> I think that the first check should always be if the transaction type is accepted at this block height. The reason for it is that it will always be needed and could save a lot of arithmetic operations if not in range. 

The idea for the order proposed was to make the simpler verifications first. In this regard, running `ValidateTxStatic` is cheaper than `ValidateTxForNetworkRules`.  If we had some statistics of the most common motives for rejecting a transaction, we could adjust to that. But so far I have no knowledge of such statistics. 

> Additionally, this would allow moving Static checks from specific tx types into the static section.

As for moving the static checks of specific types, this can and will be done, independently of the order of the validation functions.

evmcore/tx_validation.go
```diff
@@ -40,20 +50,91 @@ type validationOptions struct {
...
+// network rules. It checks the transaction's:
+// - type (legacy, access list, dynamic fee, blob, set code) is supported in the current network
+// - gas is within the max accepted gas per transaction
+// - sender is valid
+// - gas fee cap is greater than the minimum base fee
+// - gas is greater than the intrinsic gas
+// - gas is greater than the floor data cost (EIP-7623)
+//
+// It returns an error if any of the checks fail.
+func ValidateTxForNetworkRules(tx *types.Transaction, opt NetworkRulesForValidateTx) error {
```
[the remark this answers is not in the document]
facuMH: https://github.com/0xsoniclabs/sonic/pull/200/commits/95eea0367c59be9d79475922a3c6a0326df39ba8

evmcore/tx_validation.go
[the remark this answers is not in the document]
facuMH: corrected in https://github.com/0xsoniclabs/sonic/pull/200/commits/95eea0367c59be9d79475922a3c6a0326df39ba8 line 234

evmcore/tx_validation.go
```diff
@@ -189,3 +228,41 @@ func validateTx(tx *types.Transaction, opt validationOptions) error {
...
+}
+
+// validateTxForPool checks whether a transaction is valid according to the
+// current minimum tip for the pool. It returns an error if the transaction's
+// tip is lower than the minimum tip.
+func validateTxForPool(tx *types.Transaction, opt validationOptions,
+	from common.Address) error {
+
+	// Drop non-local transactions under our own minimal accepted gas price or tip
+	local := opt.isLocal || opt.locals.contains(from) // account may be local even if the transaction arrived from the network
```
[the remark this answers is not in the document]
facuMH: changed in https://github.com/0xsoniclabs/sonic/pull/200/commits/95eea0367c59be9d79475922a3c6a0326df39ba8 line 247

evmcore/tx_validation.go
```diff
@@ -40,78 +50,61 @@ type validationOptions struct {
...
-		// Check minimum revision
-		if !opt.eip7702 {
-			return ErrTxTypeNotSupported
-		}
 
-		// Check non-empty authorization list
-		if len(tx.SetCodeAuthorizations()) == 0 {
-			return ErrEmptyAuthorizations
-		}
+	// leave state accesses for last check
```
HerbertJordan, line 79: nit: not a very useful comment ..

evmcore/tx_validation.go
```diff
@@ -125,39 +118,76 @@ func validateTx(tx *types.Transaction, opt validationOptions) error {
 		return ErrTipAboveFeeCap
 	}
 
+	// For now, Sonic only supports Blob transactions without blob data.
+	if tx.BlobHashes() != nil && (len(tx.BlobHashes()) > 0 ||
+		(tx.BlobTxSidecar() != nil && len(tx.BlobTxSidecar().BlobHashes()) > 0)) {
+		return ErrTxTypeNotSupported
```
HerbertJordan, line 124: Given the granularity of other error types, this situation should get its own error definition instead of re-using the "ErrTxTypeNotSupported".

evmcore/tx_validation.go
```diff
@@ -40,78 +50,61 @@ type validationOptions struct {
...
-	if !opt.eip2718 && tx.Type() != types.LegacyTxType {
-		return ErrTxTypeNotSupported
+	if err := ValidateTxStatic(tx); err != nil {
+		return err
 	}
-	// Reject dynamic fee transactions until EIP-1559 activates.
-	if !opt.eip1559 && tx.Type() == types.DynamicFeeTxType {
-		return ErrTxTypeNotSupported
+
+	from, err := ValidateTxForNetworkRules(tx, netRules)
```
HerbertJordan, line 70: It seems like a premature optimization and an unnecessary complication to return the "from" address here and pass it to the subsequent validation steps. Logically, this does not make a lot of sense and performance wise the impact is likely very minor, since resolved sender addresses are cached.

evmcore/tx_validation.go
```diff
@@ -125,39 +118,76 @@ func validateTx(tx *types.Transaction, opt validationOptions) error {
...
+		return common.Address{}, ErrTxTypeNotSupported
+	}
+	// Reject dynamic fee transactions until EIP-1559 activates.
+	if !opt.eip1559 && tx.Type() == types.DynamicFeeTxType {
+		return common.Address{}, ErrTxTypeNotSupported
+	}
+	// Reject blob transactions until EIP-4844 activates or if is already EIP-4844 and they are not empty
+	if tx.Type() == types.BlobTxType && !opt.eip4844 {
+		return common.Address{}, ErrTxTypeNotSupported
+	}
```
HerbertJordan, line 161: If I recall correctly, we do not actually support a version of Sonic where these EIPs are disabled. Does it make sense to have these checks at this fine-granular level or should it be simplified?

@LuisPH3 wdyt?

evmcore/tx_validation.go
```diff
@@ -125,39 +118,76 @@ func validateTx(tx *types.Transaction, opt validationOptions) error {
...
+	// validate EIP-7702 transactions, part of prague revision
+	if tx.Type() == types.SetCodeTxType && !opt.eip7702 {
+		return common.Address{}, ErrTxTypeNotSupported
+	}
+	// Check whether the init code size has been exceeded, introduced in EIP-3860
+	if opt.shanghai && tx.To() == nil && len(tx.Data()) > params.MaxInitCodeSize {
```
HerbertJordan, line 167: Same here, I think "Shanghai" is always enabled in our case.

evmcore/tx_validation.go
```diff
@@ -169,23 +199,62 @@ func validateTx(tx *types.Transaction, opt validationOptions) error {
...
+// and that the transaction's nonce is not lower than the sender's nonce in the state.
+// Returns an error if any of these conditions are not met.
+func ValidateTxForState(tx *types.Transaction, state TxPoolStateDB, from common.Address) error {
+	// Ensure the transaction adheres to nonce ordering
+	if state.GetNonce(from) > tx.Nonce() {
+		return ErrNonceTooLow
+	}
+
+	// Transactor should have enough funds to cover the costs
+	// cost == V + GP * G
```
HerbertJordan, line 234: nit: It is not clear what "V", "GP", and "G" is in this context ...

evmcore/tx_pool_test.go
```diff
@@ -497,9 +497,9 @@ func TestEIP4844Transactions(t *testing.T) {
 		err    error
 	}{
 		{"empty blob tx before cancun", nil, false, ErrTxTypeNotSupported},
-		{"blob tx before cancun", common.Address{1}.Bytes(), false, ErrTxTypeNotSupported},
+		{"blob tx before cancun", common.Address{1}.Bytes(), false, ErrOversizedData},
```
HerbertJordan, line 500: The fact that the produced error changed may undermine the intention of this test. Can you update the transaction that is tested here such that it passes the `OversizedData` check and reaches the error that it previously was supposed to produce?

tests/set_code_tx_test.go
```diff
@@ -990,5 +990,5 @@ func TestSetCodeTransaction_IsRejectBeforeAllegro(t *testing.T) {
 	tx := signTransaction(t, chainId, &types.SetCodeTx{}, net.GetSessionSponsor())
 
 	err = client.SendTransaction(context.Background(), tx)
-	require.ErrorContains(t, err, "transaction type not supported")
+	require.ErrorContains(t, err, "empty authorization")
```
HerbertJordan, line 993: This defeats the original purpose of this test. You should add an authorization to the transaction to make it pass this static check to actually get to the point where the transaction type is checked.

evmcore/tx_validation_test.go
```diff
@@ -245,47 +132,156 @@ func TestValidateTx_GasPriceAndTip_RejectsTxWith(t *testing.T) {
...
+
+func TestValidateTxStatic_AuthorizationList_RejectsTxWith(t *testing.T) {
+	t.Run("setCode tx with empty authorization list", func(t *testing.T) {
+		tx := types.NewTx(&types.SetCodeTx{})
+		err := ValidateTxStatic(tx)
+		require.ErrorIs(t, err, ErrEmptyAuthorizations)
+	})
+}
+
+func TestValidateTxStatic_ReturnsNilToAcceptableTx(t *testing.T) {
```
HerbertJordan, line 160: nit: less technical, more semantic
```suggestion
func TestValidateTxStatic_AcceptsValidTransactions(t *testing.T) {
```

evmcore/tx_validation_test.go
```diff
@@ -421,26 +389,62 @@ func TestValidateTx_Signer_RejectsTxWith(t *testing.T) {
...
+
+////////////////////////////////////////////////////////////////////////////////
+// State Validation
+
+func TestValidateTxForState_Nonce_RejectsTxWith(t *testing.T) {
+	for name, tx := range getTxsOfAllTypes() {
+		t.Run(fmt.Sprintf("older nonce/%v", name), func(t *testing.T) {
+
+			// set nonce lower than the current account nonce
+			currentNonce := uint64(2)
```
HerbertJordan, line 423: nit: should be taken from the `tx` instead of hard-coding it

HerbertJordan (changes requested):
I did not go through all the tests in detail, but I assume they are equivalent to the preceding tests.

There are are few minor things that should be looked at, please see comments.

evmcore/tx_validation.go
facuMH: This is something that occurred to me as well, since the earlier revision sonic supports in Cancun, everything before will always be enabled, so this checks could be joined. That said I wanted to keep this PR focus on only splitting the validations and reordering them, as that is a mess of its own, and then add follow-up tasks with refactors.

evmcore/tx_pool_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/200/commits/2724ab4e13337c83cbe9efe010de5df4c21bf2ba

evmcore/tx_validation.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/200/commits/2724ab4e13337c83cbe9efe010de5df4c21bf2ba

evmcore/tx_validation.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/200/commits/32055b804e8a5906eb9786572ca27dc9a38ca183

evmcore/tx_validation_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/200/commits/730cc402e58a98f26bdde28704fbde8a849ccae9

tests/set_code_tx_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/200/commits/730cc402e58a98f26bdde28704fbde8a849ccae9

evmcore/tx_pool_test.go
```diff
@@ -497,9 +497,9 @@ func TestEIP4844Transactions(t *testing.T) {
 		err    error
 	}{
 		{"empty blob tx before cancun", nil, false, ErrTxTypeNotSupported},
-		{"blob tx before cancun", common.Address{1}.Bytes(), false, ErrTxTypeNotSupported},
+		{"blob tx before cancun", common.Address{1}.Bytes(), false, ErrNonEmptyBlobTx},
```
LuisPH3, line 500: I am concerned about this. The test is called `"blob tx before cancun`. The previous error was correct.
This comes together with the order of checks.
I really think we should check first the tx type. before anything else.

evmcore/tx_pool.go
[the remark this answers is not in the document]
LuisPH3: I understand the semantic grouping, but I still think the type should be tested first.
Lets clear the documentation before we merge this PR

LuisPH3 (changes requested):
There are severe inconsistencies between this code and the proposed changes in documentation. 
We need to close changes in the documentation first.

evmcore/tx_pool_test.go
facuMH: order of verifications has been changed so that tx type checked first and this tests checks again for `ErrTxTypeNotSupported`

HerbertJordan (commented):
I am not sure about the state of this PR. Is it work in progress or ready for review?

Since it is quite extensive, and rebases have deleted past review history for me, I will delay the review until all open comments are addressed and resolved. Please re-request a review at that point.

LuisPH3 (approved):
looking good.
checks have been separated: types are the first test executed for user-friendliness, the rest progress with the complexity of the tests executed.
Signature verification is embedded in the tests requiring the sender address.

Testing seems sufficient, some day we will CT the pool and the validation as well.

evmcore/tx_validation.go
facuMH: signature check has been moved inside functions that need the sender address

tests/set_code_tx_test.go
```diff
@@ -993,7 +993,8 @@ func TestSetCodeTransaction_IsRejectBeforeAllegro(t *testing.T) {
 	chainId, err := client.ChainID(context.Background())
 	require.NoError(t, err, "failed to get chain ID")
 
-	tx := signTransaction(t, chainId, &types.SetCodeTx{}, net.GetSessionSponsor())
+	setCodeTx := &types.SetCodeTx{AuthList: []types.SetCodeAuthorization{{}}}
+	tx := signTransaction(t, chainId, setCodeTx, net.GetSessionSponsor())
```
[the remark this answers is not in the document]
facuMH: https://github.com/0xsoniclabs/sonic/pull/200/commits/0bfef4eec902803082e6785ef49853e1694b3bf5

LuisPH3 (approved):
The PR refactors the transaction validation routine for modularity, using semantic blocks of validation checks,

evmcore/tx_validation.go
```diff
@@ -125,19 +116,73 @@ func validateTx(tx *types.Transaction, opt validationOptions) error {
...
-		return ErrUnderpriced
+	// Check non-empty authorization list
+	if tx.Type() == types.SetCodeTxType && len(tx.SetCodeAuthorizations()) == 0 {
+		return ErrEmptyAuthorizations
+	}
+
+	// Reject transactions over defined size to prevent DoS attacks
+	if uint64(tx.Size()) > txMaxSize {
+		return ErrOversizedData
+	}
```
HerbertJordan, line 133: This property is a limit imposed by the transaction pool, not by the specification of transactions. This makes it not independent of any context. This check should probably be part of `validateTxForPool`.

evmcore/tx_validation.go
```diff
@@ -125,19 +116,73 @@ func validateTx(tx *types.Transaction, opt validationOptions) error {
...
 	}
 
-	// Make sure the transaction is signed properly.
-	from, err := types.Sender(opt.signer, tx)
-	if err != nil {
-		return ErrInvalidSender
+	// Sonic only supports Blob transactions without blob data.
+	if tx.Type() == types.BlobTxType &&
+		(len(tx.BlobHashes()) > 0 || (tx.BlobTxSidecar() != nil && len(tx.BlobTxSidecar().BlobHashes()) > 0)) {
+		return ErrNonEmptyBlobTx
```
HerbertJordan, line 122: This is not an immutable property defined by the specification of transactions and may actually change, depending on the context. 

For instance, if we would, one day, support blobs, this becomes a check to be moved to the `ValidateTxForNetwork` function. If at that point the static checks are part of the event validation steps, we need to add a network rule parameter to this function to manage the transition.

evmcore/tx_validation.go
```diff
@@ -125,19 +116,73 @@ func validateTx(tx *types.Transaction, opt validationOptions) error {
...
+// ValidateTxForNetwork checks if a transaction is valid based on the current network rules.
+// Specifically, it ensures the transaction adheres to the following conditions:
+// - init code size (if contract creation) is within the defined limit
+// - gas fee cap is greater than the minimum base fee
+// - gas is within the max accepted gas per transaction
+// - gas is greater than the intrinsic gas
+// - gas is greater than the floor data cost (EIP-7623)
+//
+// Returns an error if any of these conditions are not met.
+func ValidateTxForNetwork(tx *types.Transaction, opt NetworkRulesForValidateTx) error {
```
HerbertJordan, line 178: The functions `ValidateTxType` and `ValidateTxForNetwork` seem to aim for the same objective -- make sure that a given transaction is valid under a given set of network rules. Why are these separated?

evmcore/tx_validation.go
```diff
@@ -40,78 +49,60 @@ type validationOptions struct {
 	eip7623 bool // Fork indicator whether we are using EIP-7623 floor gas validation.
 	eip7702 bool // Fork indicator whether we are using EIP-7702 set code transactions.
 
-	currentState   TxPoolStateDB // Current state in the blockchain head
-	currentMaxGas  uint64        // Current gas limit for transaction caps
-	currentBaseFee *big.Int      // Current base fee for transaction caps
-	minTip         *big.Int      // Minimum gas tip to enforce for acceptance into the pool
+	currentMaxGas  uint64   // Current gas limit for transaction caps
+	currentBaseFee *big.Int // Current base fee for transaction caps
```
HerbertJordan, line 53: The `currentBaseFee` is a per-block property, not a network rule property. It does not look like it should be in this list.

evmcore/tx_validation.go
```diff
@@ -189,3 +228,56 @@ func validateTx(tx *types.Transaction, opt validationOptions) error {
...
+func validateTxForPool(tx *types.Transaction, opt validationOptions,
+	signer types.Signer) error {
+
+	// Make sure the transaction is signed properly.
+	from, err := types.Sender(signer, tx)
+	if err != nil {
+		return ErrInvalidSender
+	}
+
+	// tx is local if received from the local RPC or if the sender belongs to the local accounts set
```
HerbertJordan, line 270:
```suggestion
	// A transaction is local if received from the local RPC or if the sender belongs to the local accounts set.
	// For local transactions, minimum gas tips are not enforced.
```

evmcore/tx_validation.go
```diff
@@ -189,3 +228,56 @@ func validateTx(tx *types.Transaction, opt validationOptions) error {
...
+	local := opt.isLocal || opt.locals.contains(from)
+	if local {
+		return nil
+	}
+
+	// Drop non-local transactions under our own minimal accepted gas price or tip.
+	if tx.GasTipCapIntCmp(opt.minTip) < 0 {
+		log.Trace("Rejecting underpriced tx: pool.gasPrice", "pool.gasPrice",
```
HerbertJordan, line 278:
```suggestion
		log.Trace("Rejecting underpriced tx: pool.gasPrice", "pool.minTip",
```

HerbertJordan (changes requested):

evmcore/tx_validation.go
LuisPH3: It is important, for user feedback, that the transaction type is the first check executed. 
After this one, checks proceed by cost

evmcore/tx_validation.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/200/commits/bc4a8cd1eadd2d6e1c75954cb1852d9da27c655e

evmcore/tx_validation.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/200/commits/bc4a8cd1eadd2d6e1c75954cb1852d9da27c655e

evmcore/tx_validation.go
facuMH: `ValidateTxType` and `ValidateTxForNetwork` have been re-merged. As Luis says, we want to report tx type errors first, so `ValidateTxForNetwork` is called first and after that validate functions are called by compute cost.

evmcore/tx_validation_test.go
```diff
@@ -245,117 +128,156 @@ func TestValidateTx_GasPriceAndTip_RejectsTxWith(t *testing.T) {
 	}
 }
 
-func TestValidateTx_Gas_RejectsTxWith(t *testing.T) {
-	for name, tx := range getTxsOfAllTypes() {
-		t.Run(fmt.Sprintf("current max gas lower than tx gas/%v", name), func(t *testing.T) {
-			opt := getTestTransactionsOption()
-			opt.currentMaxGas = 1
+func TestValidateTxStatic_AuthorizationList_RejectsTxWith(t *testing.T) {
```
LuisPH3, line 131: nit
```suggestion
func TestValidateTxStatic_AuthorizationList_RejectsTxWithEmptyAuthorization(t *testing.T) {
```

remove superfluous t..Run

LuisPH3 (approved):
I missed the decision of merging netrules and type.
This has a performance tradeoff, but it can be changed anytime without hardfork. 
On the other hand, the classification seems clean and easy to use.

evmcore/tx_validation.go
```diff
@@ -40,28 +49,70 @@ type validationOptions struct {
...
-func validateTx(tx *types.Transaction, opt validationOptions) error {
+func validateTx(
+	tx *types.Transaction,
+	opt poolOptions,
+	currentBlockState blockState,
+	netRules NetworkRules) error {
+
+	// The transaction is checked against network rules to detect unsupported
+	// transaction types first, ensuring protocol compliance before performing
+	// further validation.
```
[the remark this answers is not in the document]
facuMH: The motivation for this comment was to make obvious the reasoning behind checking validating against network rules first, and ordering by complexity afterwards. Since the rest are ordered by complexity I did not think they needed a comment.

evmcore/tx_validation_test.go
```diff
@@ -245,117 +128,156 @@ func TestValidateTx_GasPriceAndTip_RejectsTxWith(t *testing.T) {
...
-			setGasPriceOrFeeCap(t, tx, opt.minTip)
-			// sign txs with sender
-			address, signedTx := signTxForTest(t, tx)
-			// setup enough balance
-			testDb := newTestTxPoolStateDb()
-			testDb.balances[address] = uint256.NewInt(math.MaxUint64)
-			opt.currentState = testDb
-			// ---
+func TestValidateTxForNetwork_Blobs_RejectsTxWith(t *testing.T) {
+	// blob txs are not supported in sonic, so they must have empty hash list and sidecar
```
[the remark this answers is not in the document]
facuMH: I will apply the comment suggestion, but given that there are two different sub tests, I rather keep the test name as they are.

evmcore/tx_validation_test.go
```diff
@@ -16,154 +16,72 @@ import (
...
-		shanghai:       true,
-		eip4844:        true,
-		eip7623:        true,
-		eip7702:        true,
-		currentMaxGas:  100_000,
-		currentBaseFee: big.NewInt(1),
-		minTip:         big.NewInt(1),
-		isLocal:        true,
-		signer:         types.NewPragueSigner(big.NewInt(1)),
+// getTestBlockState returns a test block state database that can be used for
```
HerbertJordan, line 30: "database"?

facuMH: no database, just max gas and basefee. comment rephrased.

LuisPH3 (approved):

evmcore/tx_validation_test.go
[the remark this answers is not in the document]
facuMH: an internal document showing this mapping has been shared among reviewers.

PR #209, opened 2025-04-30
Synchronize access to cached hash in Block

HerbertJordan:
This PR adds synchronization to the cached hash of the internal block structure.

The included unit tests revealed issues consistent with the problems reported by https://github.com/0xsoniclabs/sonic-admin/issues/177. While the effect was not observed end-to-end, as its reconstruction would be very timing sensitive, this is a plausible cause for the effect reported in the issue.

### DIFF
--- a/inter/block.go
+++ b/inter/block.go
@@ -3,14 +3,17 @@ package inter
 import (
 	"encoding/binary"
 	"errors"
+	"io"
 	"math/big"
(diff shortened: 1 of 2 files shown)


jenikd (approved):
Looks good

LuisPH3 (approved):
:+1:

PR #213, opened 2025-05-02
Add scheduler for block proposals

HerbertJordan:
This PR introduces a simple, greedy scheduler for transactions in proposed blocks.

The introduced scheduler accepts transactions in a predefined order (to be defined by the emitter's `transactionsByPriceAndNonce` type) and attempts to include transactions from this order one-by-one into the final block until the gas limit for the block is reached or its scheduling context is signaled to stop. Transactions that can not be processed for whatever reason are skipped.


gossip/emitter/scheduler/scheduler_test.go
LuisPH3, line 1: missing test?

TestScheduler_Schedule_PreservesPositioningRespectToOriginalTxsOrder

gossip/emitter/scheduler/scheduler.go
```diff
@@ -0,0 +1,163 @@
...
+
+	return res
+}
+
+// PrioritizedTransactions is an interface for a collection of transactions to
+// be scheduled in a block. The scheduler consumes the collection's elements
+// one by one, signalling whether each of them was accepted or rejected. This
+// allows the collection to effectively prune candidates that are not going to
+// be accepted due to failed preconditions.
+type PrioritizedTransactions interface {
```
LuisPH3, line 97: nit: I keep on thinking whenever this interface should be called "cursor".
(I am looking for some name that reflects the abstraction of a collection where only one element a the time can be accessed)

LuisPH3 (approved):
The changes propose an eager algorithm with linear complexity that evaluates all candidate transaction for their inclusion within a block scoped by total gas cost. 

The execution is cancelable to allow for timeouts.
The execution preserves the position of transactions in respect to the original order.

gossip/emitter/scheduler/scheduler.go
HerbertJordan: > I am looking for some name that reflects the abstraction of a collection where only one element a the time can be accessed

That would be an iterator, or a `iter.Seq` in Go terms. 

In the previous scheduler implementation I actually used a `iter.Seq`, ignoring the feature of signaling back to the collection that some transactions have been skipped, and thus a range of additional transactions can be pruned.

In this iteration, I wanted to preserve this feature for efficiency reasons. Consequently, a `iter.Seq` or an `Iterator` does not fit the description for me any more.

simonlechner (approved):
This "simple" implementation removes a lot of complexity compared to the old one. The implementation is quite straight forward and tests cover it nicely.

gossip/emitter/scheduler/scheduler_test.go
HerbertJordan: Added by 3909afb71ddb797544a78ed6492dd14731907fd5

gossip/emitter/scheduler/scheduler.go
```diff
@@ -0,0 +1,162 @@
...
+
+// Schedule determines an executable sequence of transactions from the given
+// candidates collection. It does so by trail-running the transaction in order,
+// accepting all transactions that can be successfully processed within the
+// given gas limit.
+//
+// The provided collection of candidates is expected to only enumerate
+// transactions with unique sender/nonce pairs. If there are duplicates, the
+// scheduler will only accept the first that can be successfully processed,
+// ignoring the rest.
```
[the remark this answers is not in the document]
HerbertJordan: This is not a problem to be solved by the scheduler. Replacements are to be handled by the transaction pool. The collection of candidates is delivered by the transaction pool after resolving those.

gossip/emitter/scheduler/scheduler.go
```diff
@@ -0,0 +1,162 @@
...
+
+	// Accept signals that current transaction was accepted and should be
+	// removed from the collection, leading to the next transaction being
+	// returned by the Current method.
+	Accept()
+
+	// Skip signals that current transaction was rejected and should be removed
+	// from the collection. Furthermore, any transactions that depend on the
+	// rejected transaction should also be removed from the collection.
+	Skip()
```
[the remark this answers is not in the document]
HerbertJordan: "Accepted" means that a transaction was included in the resulting schedule, as all its pre-conditions have been satisfied. Accepted transactions may have succeed or failed, but in either case gas fees are charged and a receipt ends up in the block.

"Skipped" transactions are transactions without satisfied pre-conditions (not enough balance, wrong nonce, etc) that can not be included in a block, and are thus skipped during the scheduling.

facuMH (approved):

LuisPH3 (approved):

PR #214, opened 2025-05-05
Check NoArchiveError when sending block notifications

jenikd:
This PR fixes situation, when there is no archive database for checking current block height when sending subscribers block notification
Related to https://github.com/0xsoniclabs/sonic-admin/issues/175

### DIFF
--- a/gossip/service.go
+++ b/gossip/service.go
@@ -130,7 +130,7 @@ func (f *ServiceFeed) Start(store ArchiveBlockHeightSource) {
 			if err != nil {
 				// If there is no archive, set height to the last block
 				// and send all notifications
-				if err == evmstore.NoArchiveError {
+				if errors.Is(err, evmstore.NoArchiveError) {
 					height = pending[len(pending)-1].block.Number.Uint64()
 				} else {
 					log.Error("failed to get archive block height", "err", err)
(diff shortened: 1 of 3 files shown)


gossip/service.go
```diff
@@ -127,12 +128,20 @@ func (f *ServiceFeed) Start(store ArchiveBlockHeightSource) {
...
 			height, empty, err := store.GetArchiveBlockHeight()
 			if err != nil {
-				log.Error("failed to get archive block height", "err", err)
-				continue
-			}
-			if empty {
-				continue
+				// If there is no archive, set height to the last block
+				// and send all notifications
+				if err == evmstore.NoArchiveError {
```
HerbertJordan, line 133: nit: To be resilient towards future changes, this should be
```suggestion
				if errors.Is(err, evmstore.NoArchiveError) {
```

gossip/service_test.go
```diff
@@ -94,3 +96,83 @@ func TestServiceFeed_BlocksInOrder(t *testing.T) {
...
+				EvmHeader: evmcore.EvmHeader{
+					Number: big.NewInt(int64(test.blockHeight)),
+				},
+			}, nil)
+
+			// The notification should be delivered.
+			select {
+			case notification := <-consumer:
+				if test.expectedNotification == nil {
+					t.Fatal("expected notification to be sent")
```
HerbertJordan, line 162:
```suggestion
					t.Fatal("expected no notification to be sent")
```

HerbertJordan (approved):
Looks good, thanks!

PR #219, opened 2025-05-07
Add utilities for building block proposals

HerbertJordan:
This PR introduces a `createPayload` utility function (and needed helper infrastructure) to build payload data for events to be emitted.

When creating payload for an event, the following steps need to be conducted:
- compute the current proposal-sync-state by aggregating the information attached to the parent events
- decide whether a new proposal may and can be made
- if a proposal can be made, a proposal for a new block is created

This PR does not include the integration in the `Emitter` code base. This will follow as part of a follow-up PR.


gossip/emitter/proposals.go
```diff
@@ -0,0 +1,229 @@
...
+	rules opera.Rules,
+	incomingSyncState inter.ProposalSyncState,
+	latestBlock *inter.Block,
+	newBlockTime inter.Timestamp,
+	currentFrame idx.Frame,
+	transactionScheduler txScheduler,
+	candidates scheduler.PrioritizedTransactions,
+	durationMetric timerMetric,
+	timeoutMetric counterMetric,
+) (inter.Payload, error) {
```
LuisPH3, line 95: I think it is a little misleading to have two functions returning `(inter.Payload, error)` but with completely different semantic names. 
IMO, the first function `createPayload` should handle detection of turns and construct the `Paiload` type.
The second function should handle the block proposal exclusively, It would read better if it returns a `Proposal` type instead of an already aggregated payload.

gossip/emitter/proposals.go
```diff
@@ -0,0 +1,229 @@
...
+type timerMetric interface {
+	Update(time.Duration)
+}
+
+// counterMetric is an abstraction for monitoring metrics to facilitate testing.
+type counterMetric interface {
+	Inc(int64)
+}
+
+// We put a strict cap of 2 second on the accumulated gas. Thus, if the delay
```
LuisPH3, line 173:
```suggestion
// We put a strict cap of 2 second on the accumulated gas calculation. Thus, if the delay
```

gossip/emitter/proposals.go
```diff
@@ -0,0 +1,229 @@
...
+	lastBlockTime := latestBlock.Time
+	if lastBlockTime >= newBlockTime {
+		// no time has passed, so no new proposal can be made
+		return inter.Payload{
+			ProposalSyncState: incomingSyncState,
+		}, nil
+	}
+	effectiveGasLimit := getEffectiveGasLimit(
+		newBlockTime.Time().Sub(lastBlockTime.Time()),
+		rules.Economy.ShortGasPower.AllocPerSec, // TODO: consider using a new rule set parameter
```
LuisPH3, line 107: question, should the maximum block time cap be configurable as well?

LuisPH3 (approved):
some style notes:

gossip/emitter/proposals.go
```diff
@@ -0,0 +1,229 @@
...
+type counterMetric interface {
+	Inc(int64)
+}
+
+// We put a strict cap of 2 second on the accumulated gas. Thus, if the delay
+// between two blocks is less than 2 seconds, gas is accumulated linearly.
+// If the delay is longer than 2 seconds, we cap the gas to the maximum
+// accumulation time. This is to limit the maximum block size to at most
+// 2 seconds worth of gas.
+const maxAccumulationTime = 2 * time.Second
```
simonlechner, line 178: Would it make sense to have this max time configurable?

simonlechner (approved):
Looks good, I could not find anything problems with the code and testing.

gossip/emitter/proposals.go
HerbertJordan: The maximum block time cap needs to be aligned among all nodes, since it is consensus critical. Thus, the only way to make this configurable would be through network rules.

However, it turns out that extending the network rules is difficult due to the internal use of RLP encoding, which is neither forward nor backward compatible. When I was writing this, I thought it would not be worth the effort, and used a constant instead.

Do you see a real use case for making this configurable? As we might modify the network rules for other reasons, we could consider including this.

gossip/emitter/proposals.go
```diff
@@ -0,0 +1,229 @@
...
+	inter.EventReader
+	GetLatestBlock() *inter.Block
+	GetRules() opera.Rules
+}
+
+// --- proposal creation ---
+
+// createProposal creates a new block proposal based on the given context
+// information. The resulting payload contains the new block proposal.
+func createProposal(
```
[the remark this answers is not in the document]
HerbertJordan: I updated the return type of the function to be an actual proposal. Regarding the name, I renamed it to `makeProposal`.

gossip/emitter/proposals.go
```diff
@@ -0,0 +1,229 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/txpool"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/holiman/uint256"
+)
+
+//go:generate mockgen -source=proposals.go -destination=proposals_mock.go -package=emitter
+
+// createPayload is a helper function handling the actual creation of the
+// payload. It is detached from the Emitter to facilitate testing.
```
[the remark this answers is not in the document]
HerbertJordan: Updated

gossip/emitter/proposals.go
HerbertJordan: Fixed.

gossip/emitter/proposals.go
HerbertJordan: I rephrased this documentation, leading to additional changes.

gossip/emitter/proposals.go
```diff
@@ -0,0 +1,229 @@
...
+func createPayload(
```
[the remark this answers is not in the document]
HerbertJordan: Yes, the documentation needed improvement. See https://github.com/0xsoniclabs/sonic/pull/219#discussion_r2099879039

gossip/emitter/proposals.go
simonlechner: If we figure out we have overseen something related to security it will be easier to fix if the time is configurable. But I guess we have thought about all possible cases a few times by now.

LuisPH3 (approved):

PR #220, opened 2025-05-07
Move mocks for emitter package into expected location

HerbertJordan:
The mock for the interfaces defined for the `emitter` package have so fare been placed in `gossip/emitter/mock/world.go`. By convention, these definitions should be placed in `gossip/emitter/world_mock.go`.

The non-conforming location of the mock definitions adds unnecessary complexity to the code structure. Furthermore, it leads to problems with the code coverage reporting, as only `XXX_mock.go` files are identified as generated mocks and excluded from the code coverage report. At its current location, a lack of utilizing all features in the generated code leads to low code coverage results.

This change updates the code location and fixes dependencies.


facuMH (approved):
This PR moves the mock of interfaces defined in `gossip/emitter/world.go` from `gossip/emitter/mock/world.go` to `gossip/emitter/world_mock.go` as is conforming with good practices and standards.

PR #221, opened 2025-05-07
Add ChainId to Legacy RPC transactions

facuMH:
This PR build on https://github.com/0xsoniclabs/sonic/pull/167 by expanding a unit test.

As mentioned in the original PR, adding the `ChainId` to Legacy transactions for compatibility with other chains.


ethapi/api.go
```diff
@@ -1401,6 +1401,8 @@ func newRPCTransaction(tx *types.Transaction, blockHash common.Hash, blockNumber
 	}
 
 	switch tx.Type() {
+	case types.LegacyTxType:
+		result.ChainID = (*hexutil.Big)(tx.ChainId())
```
HerbertJordan, line 1405: Can you explain why adding the ChainID is only necessary for the Legacy transaction type?

HerbertJordan (changes requested):
The original PR #167 introducing this change is motivating the change by the inability to verify transaction hashes and/or signatures for legacy transactions when being retrieved through the RPC interface. Can you reproduce this case in a test and verify that this change fixes this?

Also, is this problem limited to legacy transactions?

ethapi/api.go
facuMH: on one hand, chainID is assigned in `copyAccessList`, which is called for all types except Legacy. I believe the reason for this could have been that semantically, legacy Tx do not have a chain id.

LuisPH3: Legacy transactions existed before transaction envelope, the eip which introduces different types of txs (https://eips.ethereum.org/EIPS/eip-2718), all new types have chain id embedded. 

BUT! Legacy transactions existed before replay protection was introduced: https://eips.ethereum.org/EIPS/eip-155
Technically, old enough signers can sign legacy txs without involving any chain id. (i think is homestead, but not this needs a check)

Any TX in Sonic should contain chain id.  either the network id (testing/production) or the 0 chain id.

ethapi/api_test.go
```diff
@@ -595,6 +604,75 @@ func TestTransactionJSONSerialization(t *testing.T) {
...
+			signed := signTransaction(t, ChainId, tx, key)
+
+			// some block data for the test
+			blockHash := common.Hash{1, 2, 3, 4}
+			blockNumber := uint64(4321)
+			index := uint64(42)
+			baseFee := big.NewInt(1234)
+
+			rpcTx := newRPCTransaction(signed, blockHash, blockNumber, index, baseFee)
+			require.Equal(t, signed.Hash(), rpcTx.Hash)
```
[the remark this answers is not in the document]
facuMH: Agreed. Convertion from `rpcTx` back to `types.Transaction` and comparison of hashes added to both tests, alltxs and legacy with chain id 0.

ethapi/api.go
facuMH: assignment now done in struct initialization.

facuMH:
> can we check that all chain id, hash, and signature are correct?

These checks are done [line 670](https://github.com/0xsoniclabs/sonic/pull/221/files#diff-70dfcd2cb385d58859af9d28d9857410844f0d1df288fcbdfa12cd7f2fc636afR670) and [line 706](https://github.com/0xsoniclabs/sonic/pull/221/files#diff-70dfcd2cb385d58859af9d28d9857410844f0d1df288fcbdfa12cd7f2fc636afR706) respectively.

HerbertJordan (approved):
Looks good, thanks.

ethapi/api_test.go
```diff
@@ -660,23 +660,61 @@ func TestNewRPCTransaction_AllTxSignatureAndHashCanBeVerified(t *testing.T) {
...
+
+			// verify sender can be retrieved
+			sender, err := types.Sender(types.LatestSignerForChainID(rpcTx.ChainID.ToInt()), decoded)
+			require.NoError(t, err)
+			require.Equal(t, crypto.PubkeyToAddress(key.PublicKey), sender)
 		})
 	}
 }
 
+func TestNewRPCTransaction_LegacyTxSignedWithHomesteadMaintainsChainID(t *testing.T) {
```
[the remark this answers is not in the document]
facuMH: rename to: `TestNewRPCTransaction_LegacyTxSignedWithHomesteadCanBeReproducedAndVerified`

PR #224, opened 2025-05-08
Add proposal synchronization utilities

HerbertJordan:
This PR introduces the `ProposalSyncState` type and associated utility functions.

The `ProposalSyncState` is the state propagated along with each event to track the latest seen proposal information. This information is required to determine whether a validator is allowed to include a proposal in an emitted event.

The provided utilities are required to be consistent at two places:
- the event emitter process, to decide whether a proposal may be made
- the event validation process, to decide whether a received proposal is valid

Thus, a shared implementation for both cases is provided in the `inter` package.


inter/payload.go
```diff
@@ -20,10 +20,8 @@ const (
 // defining new RLP encoded content, this payload uses protobuf encoding to
 // standardize the serialization of the content and simplify portability.
 type Payload struct {
-	LastSeenProposalTurn  Turn
-	LastSeenProposedBlock idx.Block
-	LastSeenProposalFrame idx.Frame
-	Proposal              *Proposal
+	ProposalSyncState
+	Proposal *Proposal
```
LuisPH3, line 24: I think this could be a good place to state that `ProposalSyncState` is mandatory on every event, while only proposing events should have a proposal in payload. Therefore making then invalid when not in turn.

inter/proposal_sync_state.go
```diff
@@ -0,0 +1,104 @@
...
+	return ProposalSyncState{
+		LastSeenProposalTurn:  max(a.LastSeenProposalTurn, b.LastSeenProposalTurn),
+		LastSeenProposedBlock: max(a.LastSeenProposedBlock, b.LastSeenProposedBlock),
+		LastSeenProposalFrame: max(a.LastSeenProposalFrame, b.LastSeenProposalFrame),
+	}
+}
+
+// GetIncomingProposalSyncState aggregates the last seen proposal information
+// from the event's parents.
+func GetIncomingProposalSyncState(
```
LuisPH3, line 33: This is the constructor of the `ProposalSyncState`. This constructor happens when creating a new event.
Would a "constructing" verb fit better than "get" ?
- `GenerateIncomingProposalSyncState`
- or even just: `NewProposalSyncState`  for simplicity

inter/proposal_sync_state.go
```diff
@@ -0,0 +1,104 @@
...
+	parents := event.Parents()
+
+	// For genesis events without errors, there is no last seen proposal.
+	// However, we need to set the last seen proposed block to the start block
+	// of the epoch to retain progress.
+	if len(parents) == 0 {
+		res.LastSeenProposedBlock = reader.GetEpochStartBlock(event.Epoch())
+		return res
+	}
+
```
LuisPH3, line 47: question?  Is there any scenario with != 2 parents?  other than 0 parents genesis.

inter/proposal_sync_state.go
```diff
@@ -0,0 +1,104 @@
...
+	// For all other events, the last seen proposal information of the parents
+	// needs to be aggregated.
+	for _, parent := range parents {
+		current := reader.GetEventPayload(parent).ProposalSyncState
+		res = JoinProposalSyncStates(res, current)
+	}
+	return res
+}
+
+type EventReader interface {
```
LuisPH3, line 57: I am missing some context:
is this the complete history of events?  or is it some sort of subset?

inter/proposal_sync_state.go
```diff
@@ -0,0 +1,104 @@
...
+// --- determination of the proposal turn ---
+
+// IsAllowedToPropose checks whether the current validator is allowed to
+// propose a new block.
+func IsAllowedToPropose(
+	proposalState ProposalSyncState,
+	currentFrame idx.Frame,
+	validators *pos.Validators,
+	thisValidator idx.ValidatorID,
+	blockToPropose idx.Block,
```
LuisPH3, line 71: nit:

"is" (third person)  and this validator (almost first person)  reads a little weird. I propose:
```go
func IsAllowedToPropose(
    validator idx.ValidatorID,
	proposalState ProposalSyncState,
	currentFrame idx.Frame,
	validators *pos.Validators,
	blockToPropose idx.Block,
	) (bool, error) {
```
- rename
- first argument to read along the function name

inter/proposal_sync_state.go
```diff
@@ -0,0 +1,104 @@
...
+		ProposalSummary{
+			Turn:  proposalState.LastSeenProposalTurn,
+			Frame: proposalState.LastSeenProposalFrame,
+		},
+		ProposalSummary{
+			Turn:  nextTurn,
+			Frame: currentFrame,
+		},
+	)
+	if !valid {
```
LuisPH3, line 98: nit:
`return valid, nil`

inter/proposal_sync_state_test.go
```diff
@@ -0,0 +1,226 @@
...
+
+	"github.com/Fantom-foundation/lachesis-base/hash"
+	"github.com/Fantom-foundation/lachesis-base/inter/dag"
+	"github.com/Fantom-foundation/lachesis-base/inter/idx"
+	"github.com/Fantom-foundation/lachesis-base/inter/pos"
+	"github.com/stretchr/testify/require"
+	gomock "go.uber.org/mock/gomock"
+)
+
+func TestProposalSyncState_Join(t *testing.T) {
```
LuisPH3, line 14: nit: missing property tested name

LuisPH3 (commented):
- event payload Turn, Frame, Block# are migrated to a class ProposalSyncState

LuisPH3 (dismissed):
-  event payload Turn, Frame, Block# are migrated to a class ProposalSyncState
- ProposalSyncState provides tooling to generate the next event ProposalSyncState value given the state of the parent events.
- A function IsAllowedToPropose is provided to check both proposing time and received proposal turn validation

inter/proposal_sync_state.go
HerbertJordan: This is not really the constructor of the sync state of an event. This function is an aggregator of the sync state of all parents of an event. This is intended to be used in two different contexts:
- as the fist step when creating an event, this state must be computed by aggregating the state of all parents events; this then serves as the input to decide if a new proposal can be made. If not, the incoming state is the same as the state that ends up in the event. However, if a proposal can be made, the sync state on the event must be updated to consider the new proposal;
- as a step during event validation; we need to re-compute the sync state for event during validation to make sure the sync-state information given by the event (and a potential proposal) is valid.

As such, it is more an analysis than a "generator".

inter/proposal_sync_state.go
HerbertJordan: The parent selection process is [here](https://github.com/0xsoniclabs/sonic/blob/b7051ed82af0d601bd5c0be5763ff7548926fbb0/gossip/emitter/parents.go#L40). There is no limit to only 2 parents.

inter/proposal_sync_state.go
HerbertJordan: I renamed it to `CalculateIncomingProposalSyncState`.

inter/proposal_sync_state.go
HerbertJordan: I updated the documentation to document the requirements.

inter/proposal_sync_state.go
HerbertJordan: Right, I updated it to use the validator ID as the first parameter, and then sort the rest of the parameter in decreasing scope.

simonlechner (approved):

inter/payload.go
```diff
@@ -20,10 +20,14 @@ const (
 // defining new RLP encoded content, this payload uses protobuf encoding to
 // standardize the serialization of the content and simplify portability.
 type Payload struct {
-	LastSeenProposalTurn  Turn
-	LastSeenProposedBlock idx.Block
-	LastSeenProposalFrame idx.Frame
-	Proposal              *Proposal
+	// ProposalSyncState keeps track of the turn-taking of proposers and enables
+	// the decision of who's turn it is to propose a block. This information is
```
facuMH, line 24:
```suggestion
	// the decision of whose turn it is to propose a block. This information is
```

inter/payload.go
```diff
@@ -20,10 +20,14 @@ const (
...
+	// present in all events.
+	ProposalSyncState
+	// Proposal is an optional proposal for a new block that can be included in
+	// the payload of an event by a producer who is allowed to do so based on
```
facuMH, line 28:
```suggestion
	// the payload of an event only by a producer who is allowed to do so based on
```

PR #211, opened 2025-05-02
Change linters to golangci-lint

wsodsong:
This PR introduces golangci-lint which is a go linters runner. Linters which are enabled includes  `errcheck`, `govet`, `ineffassign`, `staticcheck` and `unused`.

The linters currently used in Sonic are `errcheck`, `govet`, `staticcheck` and `deadcode`. All except `deadcode` are supported by golangci-lint. The runner supports `unused` tool which serves as unused code detection which has similar features as `deadcode`.

TODO
- [ ] update Makefile

PR #225, opened 2025-05-15
Formating changes

facuMH:
This PR changes the formatting in several files to be compliant with `gofmt`, as this linter will be included as part of `golangci-lint` in https://github.com/0xsoniclabs/sonic/pull/211.


cmd/sonicd/metrics/disksize_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/ethereum/go-ethereum/metrics"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMeasureDbDir(t *testing.T) {
```
LuisPH3, line 13:
```suggestion
func TestMeasureDbDir_LogsDBDirSizeEveryMinute(t *testing.T) {
```

cmd/sonicd/metrics/disksize_test.go
```diff
@@ -0,0 +1,38 @@
...
+	require.NoError(t, err)
+
+	gaugeName := "test_db_size"
+
+	// Run measureDbDir in a goroutine and stop it after a short duration
+	go func() {
+		measureDbDir(gaugeName, tempDir)
+	}()
+	// disk gets measured once per minute, so we have to wait for that
+	time.Sleep(1*time.Second + time.Minute)
```
LuisPH3, line 32: I am afraid that we should not merge this change, having a unit test that takes 1 minute to complete is very taxing, and it does only check that the function records filesize once. it should check that it is a periodic event.

I suggest that we extract this issue from this pr, and we try to test it using this: https://tip.golang.org/doc/go1.24#testing-synctest

LuisPH3 (changes requested):
Format seems correct, unfortunately I think that we should have another iteration on the introduced unit test.

simonlechner (commented):
The formatting changes consist of white space addition/removal and the removal of types in the construction of container types. These changes look good, please consider @LuisPH3' suggestions regarding the unit test.

cmd/sonicd/metrics/disksize_test.go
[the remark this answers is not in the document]
HerbertJordan: I do not think that we should mix up such general style fixes with fixing missing code coverage.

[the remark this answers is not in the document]
facuMH: unit test shall be moved to its own PR

cmd/sonicd/metrics/disksize_test.go
facuMH: this is a great suggestion. I will look into this experimental package

HerbertJordan (approved):

simonlechner (approved):

PR #226, opened 2025-05-19
Integrate new randao types into block proposal. 

LuisPH3:
This PR integrates the use of `randao.RandaoReveal` into block proposing events payload.


inter/pb/proposal.proto
```diff
@@ -15,13 +15,12 @@ message Proposal {
     // The time stamp to be used in the block, expressed in Unix nano-seconds.
     uint64 timestamp = 2;
     
-    // The hash of the parent block, to 
+    // The hash of the parent block, to synchronize on the correct processing of the preceding block
```
HerbertJordan, line 18: Sorry, second thoughts:
```suggestion
    // The hash of the parent block, to establish the basis this proposal is build on.
```

inter/proposal.go
```diff
@@ -84,7 +85,7 @@ func (p *Proposal) fromProto(pb *pb.Proposal) error {
 	p.Number = idx.Block(pb.Number)
 	copy(p.ParentHash[:], pb.ParentHash)
 	p.Time = Timestamp(pb.Timestamp)
-	copy(p.Randao[:], pb.Randao)
+	copy(p.RandaoReveal[:], pb.RandaoReveal)
```
HerbertJordan, line 88: The previous `p.RandaoReveal` may not have the same size as the incoming value, as now the former array type `common.Hash` was replaced by a slice-based `randao.RandaoReveal`. I guess this line should use something like [`bytes.Clone`](https://pkg.go.dev/bytes#Clone).

Alternatively, I think the randao-reveals should actually be fixed-sized, as signatures have a fixed length. How about updating the `RandaoReveal` to be an array, not a slice?

inter/event_serializer_test.go
```diff
@@ -637,7 +637,7 @@ func FakeEvent(version uint8, txsNum, mpsNum, bvsNum int, ersNum bool) *EventPay
 				Number:       idx.Block(rand.IntN(10_000_000)),
 				ParentHash:   common.Hash(randHash(r)),
 				Time:         Timestamp(rand.Uint64()),
-				Randao:       common.Hash(randHash(r)),
+				RandaoReveal: []byte{13},
```
HerbertJordan, line 640:
```suggestion
				RandaoReveal: randBytes(r,48),
```

HerbertJordan (changes requested):

inter/proposal.go
LuisPH3: now that the randaoreveal is fully specified, it makes sense to make it an array, yes

inter/pb/proposal.proto
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/226/commits/fabaacd1900ffddd91217ca46143bf61a598e95d

inter/event_serializer_test.go
LuisPH3: 64. yes

inter/proposal.go
LuisPH3: it is now an array.

HerbertJordan (approved):
Looks good, thanks!

PR #227, opened 2025-05-19
Rename FeatureSet to HardFork

HerbertJordan:
This PR renames the former `FeatureSet` infrastructure into `HardFork`s, as these feature sets are indeed describing irreversible protocol changes.

As we see the increasing need for the ability to integrate features into the client below the scope of a full-blown hard-fork, the name "feature" for such options is colliding with the existing "FeatureSet" used for hard forks. To make space for the new feature flags, this PR is renaming the existing feature sets into hard forks, which is arguable a more accurate designation for what these options are representing.

A follow-up PR will introduce new infrastructure for true feature flags, enabling the forward-compatible controlling of sub-hard-fork scale changes on the network.

**Note for the reviewers:** these changes should only comprise the renaming of types, fields, variables, and functions as well as adjustments of the corresponding documentation strings.


LuisPH3 (commented):
The changes are meaningful. Unfortunately, these operations expose untested code. Please let me know if I can help to add tests here

LuisPH3:
I have been thinking about this. Correct me if I am wrong.
The hard fork needs to be scheduled in time, to be active from an agreed-upon block height. This structure cannot affect hard forks because such need to be hardcoded with the chain config. 
I think the current structure name may be better reflecting that this is a feature flag, and not a hard fork.

HerbertJordan:
> The hard fork needs to be scheduled in time, to be active from an agreed-upon block height. This structure cannot affect hard forks because such need to be hardcoded with the chain config.

Unlike Ethereum, Sonic is not using hard-coded block heights for hard-forks. Sonic uses network rule changes to signal hard forks. These rules contain Upgrade flags ([here](https://github.com/0xsoniclabs/sonic/blob/dbbff57c132e3ccd6662c085a73cf02bbfea3f1e/opera/rules.go#L211-L220)) that get converted to Ethereum's ChainConfig by the [EvmChainConfig](https://github.com/0xsoniclabs/sonic/blob/dbbff57c132e3ccd6662c085a73cf02bbfea3f1e/opera/rules.go#L256) function as needed.

Thus, these flags in the Update struct of the Rules are the flags used for signaling hard-forks. And the former `FeatureSet` was setting those flags.

LuisPH3:
this PR conflicts with https://github.com/0xsoniclabs/sonic/pull/238. where FeatureSet is removed.

HerbertJordan:
Closed in favor of #238

PR #229, opened 2025-05-21
Refactor and test `measureDbDir`

facuMH:
This change was originally part of [#225](https://github.com/0xsoniclabs/sonic/pull/225), where a minor formatting fix to `measureDbDir` (to comply with `gofmt`) revealed that the function had no existing test coverage.

This PR introduces a `Context` to `measureDbDir`, allowing the internal measurement loop to be cleanly stopped. It also parametrizes how long should the delay between measurements, adding a last parameter of type `time.Duration`. This last change makes testing of this function a lot easier. 

NOTE: since the signature of `measureDbDir` is changed, a test is also required for `SetDataDir`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,6 @@ sonic-image:
 test:
 	go test -cover ./...
 
-
 .PHONY: coverage
 coverage:
 	go test -coverprofile=cover.prof $$(go list ./... | grep -v '/gossip/contract/' | grep -v '/gossip/emitter/mock' | xargs)
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,8 @@ sonic-image:
 
 .PHONY: test
 test:
-	go test -cover ./...
+	GOEXPERIMENT=synctest go test -cover -coverprofile=coverage.txt -timeout=30m ./...
+
 
 .PHONY: coverage
 coverage:
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ sonic-image:
 
 .PHONY: test
 test:
-	GOEXPERIMENT=synctest go test -cover -coverprofile=coverage.txt -timeout=30m ./...
+	go test -cover ./...
 
 
 .PHONY: coverage
--- a/cmd/sonicd/metrics/disksize.go
+++ b/cmd/sonicd/metrics/disksize.go
@@ -13,10 +13,10 @@ import (
 
 var once sync.Once
 
-func SetDataDir(datadir string) {
+func SetDataDir(ctx context.Context, datadir string) {
 	once.Do(func() {
-		go measureDbDir(context.Background(), "db_size", datadir)
-		go measureDbDir(context.Background(), "statedb/disksize", filepath.Join(datadir, "carmen"))
+		go measureDbDir(ctx, "db_size", datadir)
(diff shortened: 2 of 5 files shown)


cmd/sonicd/metrics/disksize.go
```diff
@@ -14,17 +15,20 @@ var once sync.Once
...
+func measureDbDir(ctx context.Context, name, datadir string) {
 	var (
-		gauge = metrics.GetOrRegisterGauge(name, nil)
+		gauge  = metrics.GetOrRegisterGauge(name, nil)
 		rescan = len(datadir) > 0 && datadir != "inmemory"
 	)
 	for rescan {
+		if ctx.Err() != nil {
+			return
+		}
```
HerbertJordan, line 31: If the goal is to support the cancellation of this function by canceling the context, the sleep should be replaced by a select like this (not tested, just a rough outline):
```
select {
  case <-ctx.Done(): return
  case <-time.After(time.Minute):
}
```

cmd/sonicd/metrics/disksize.go
```diff
@@ -14,17 +15,20 @@ var once sync.Once
...
+		go measureDbDir(context.Background(), "db_size", datadir)
+		go measureDbDir(context.Background(), "statedb/disksize", filepath.Join(datadir, "carmen"))
 	})
 }
 
-func measureDbDir(name, datadir string) {
+func measureDbDir(ctx context.Context, name, datadir string) {
 	var (
-		gauge = metrics.GetOrRegisterGauge(name, nil)
+		gauge  = metrics.GetOrRegisterGauge(name, nil)
```
HerbertJordan, line 25: nit: probably a formatting issue, like the one this started with.

cmd/sonicd/metrics/disksize.go
HerbertJordan: Also, unrelated to your changes: Since this is a monitoring utility that wants to sample the disk usage periodically, a timer should be used instead of a sleep; this way, you can avoid the drifting of the sampling due to the delay introduced by measuring the disk size;

cmd/sonicd/metrics/disksize.go
facuMH: this is the one that started this.

Makefile
```diff
@@ -34,7 +34,8 @@ sonic-image:
 
 .PHONY: test
 test:
-	go test -cover ./...
+	GOEXPERIMENT=synctest go test -cover -coverprofile=coverage.txt -timeout=30m ./...
```
LuisPH3, line 37: this only sets the flag for this one target. there are other test-running targets.

LuisPH3: additionally, this is quite blind, not explanation whatsoever why we need this flag or where does it come from.

LuisPH3 (changes requested):

Makefile
facuMH: experimental package is no longer needed. so this change is removed.

HerbertJordan (changes requested):
Those tests seem to be at risk of being flaky. Did you verify this?

LuisPH3 (commented):
This PR incorporates too many changes,
once interval is introduced, there is no need to add context and test it.

facuMH:
This PR will be resumed in the future.

PR #234, opened 2025-05-23
Relax constraints on which blocks are allowed to be proposed

HerbertJordan:
This addresses [#182](https://github.com/0xsoniclabs/sonic-admin/issues/182) by relaxing the constraints under which new blocks can be proposed.

Before this change, a proposal for block `N` could only be followed by a proposal for block `N+1`. While formally valid, this could be very brittle. Any issue that causes the dropping of a proposal or the introduction of a block by the block processor would break this strict succession rule.

The new rules introduced by this PR only impose the following restriction:
- a proposal for block 'N' must not be followed by another proposal for block 'N' for a 'TurnTimoutFrame' frames

This should avoid different validators to make competing proposals for the same block. At the same time, it allows validators to propose subsequent blocks any time. Furthermore, it allows honest validators to correct faulty proposals of byzantine validators.


inter/proposal_sync_state.go
```diff
@@ -77,15 +64,18 @@ func IsAllowedToPropose(
...
-	if proposalState.LastSeenProposedBlock+1 != blockToPropose {
-		return false, nil
+	// Check that the block about to be proposed is not a replacement of the
+	// last seen proposed block, which might not have been confirmed yet.
+	// If a proposal was not confirmed within the timeout period, a replacement
+	// can be proposed.
+	if currentFrame < proposalState.LastSeenProposalFrame+TurnTimeoutInFrames {
+		if blockToPropose == proposalState.LastSeenProposedBlock {
+			return false, nil
+		}
```
LuisPH3, line 74: could this ever be `blockToPropose < proposalState.LastSeenProposedBlock`

inter/proposal_sync_state_test.go
```diff
@@ -40,7 +40,7 @@ func TestProposalSyncState_Join_ComputesTheMaximumForIndividualStateProperties(t
 	}
 }
 
-func TestCalculateIncomingProposalSyncState_ProducesEpochStartStateForGenesisEvent(t *testing.T) {
+func TestCalculateIncomingProposalSyncState_ProducesDefaultStateForGenesisEvent(t *testing.T) {
```
LuisPH3, line 43: is this final?  or still to be decided? 
I dont get to imagine what turn we need after a genesis event.

inter/proposal_sync_state.go
HerbertJordan: Yes, this can happen. For instance:
- node A and node M are both at block height 7
- node M, being malicious, proposes a new block for height 10; it is allowed to do so with this rule changes
- node A sees the proposal for block 10 and ignores it, since it is waiting for block 8
- now it is node A's turn, and the next block to be proposed is block 8 < 10 = last seen proposed block

inter/proposal_sync_state_test.go
HerbertJordan: This is final, based on the current development state. However, we might find issues that require us to apply further udpates.

PR #239, opened 2025-05-26
Include epoch as seed in proposer selection

HerbertJordan:
This PR adds the epoch number to the seed for the verifiable random selection of proposers.

Before this change, the order of proposers in each block is equivalent as long as the stake distribution does not change. With this change, the epoch number adds additional randomness to the selection of proposers for individual turns.


LuisPH3 (approved):
I am not sure about the robustness of this operation.
This obfuscates the future turns, but they are still predictable. 
Would it be possible to change the turn assignment in the future?

PR #240, opened 2025-05-27
Sort indexed logs

jenikd:
This PR adds sorting of indexed logs. Indexed logs are searched in parallel and because of that they were not sorted. Sorting is done by `block number` and then by `log index` as log index is unique for a block.

Refers to [issue 129](https://github.com/0xsoniclabs/sonic-admin/issues/129)

### DIFF
--- a/tests/indexed_logs_test.go
+++ b/tests/indexed_logs_test.go
@@ -39,6 +39,8 @@ func TestClient_IndexedLogsAreInOrder(t *testing.T) {
 		Topics:    [][]common.Hash{{logs[0].Topics[0], logs[1].Topics[0], logs[2].Topics[0]}},
 	})
 	require.NoError(t, err)
+	// EmitEvents is called twice and contract produces 3 logs 5 times
+	require.Len(t, blockLogs, 2*5*3)
 
 	for i := 0; i < len(blockLogs)-1; i++ {
 		current := blockLogs[i]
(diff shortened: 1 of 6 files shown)


tests/indexed_logs_test.go
```diff
@@ -0,0 +1,57 @@
...
+		current := blockLogs[i]
+		next := blockLogs[i+1]
+
+		// Check if BlockNumber is non-decreasing
+		if current.BlockNumber > next.BlockNumber {
+			t.Errorf("BlockNumber out of order: %d > %d", current.BlockNumber, next.BlockNumber)
+		}
+
+		// If BlockNumbers are equal, check if Index is non-decreasing
+		if current.BlockNumber == next.BlockNumber && current.Index > next.Index {
```
HerbertJordan, line 53: nit: also check that there are no duplicates
```suggestion
		if current.BlockNumber == next.BlockNumber && current.Index >= next.Index {
```

tests/indexed_logs_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// Search in indexed logs in parallel
+	blockLogs, err := client.FilterLogs(t.Context(), ethereum.FilterQuery{
+		FromBlock: txReceiptBlock1.BlockNumber,
+		ToBlock:   txReceiptBlock2.BlockNumber,
+		Addresses: []common.Address{contractAddress},
+		Topics:    [][]common.Hash{{logs[0].Topics[0], logs[1].Topics[0], logs[2].Topics[0]}},
+	})
+	require.NoError(t, err)
+
```
HerbertJordan, line 42: Maybe here the number of reported log messages can be checked as well.
```
         require.Len(t, blockLogs, 2*5*3)
```

HerbertJordan (approved):
Great, thanks!

tests/indexed_logs_test.go
jenikd: Done

tests/indexed_logs_test.go
jenikd: Added

PR #241, opened 2025-05-27
Simplify makefile targets.

LuisPH3:
This PR is required because the changes committed this morning would only test integration test runs. This was reporting misleading coverage for the PRs during the day.

To avoid future confusion. This script is simplified to the minimum set of features: 
- test target runs all tests without coverage
- coverage target runs all tests with coverage
These targets are meant for integration with the CI system. No need to keep historical data or other features. 
Developers have access to better tooling in the IDE.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -33,19 +33,9 @@ pipeline {
             }
         }
 
-        stage('Run unit tests') {
+        stage('Run tests') {
             steps {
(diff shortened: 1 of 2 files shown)


LuisPH3:
FYI @wsodsong

facuMH (approved):
This PRs combines the testing makefile targets of unit testing and jenkins targets.

PR #242, opened 2025-05-28
Replace on-DAG block proposal tracking with local tracking

HerbertJordan:
This PR replaces the tracking of the progression of block proposals retained as part of the event payloads with a process local tracking of pending proposals.

**Avoiding Proposals for the same Block**
Without tracking pending proposals the following sequence of operations would occur on the chain:
- initially, validator A and B have completed the processing of block 10
- validator A is allowed to make a proposal for a new block, and does so for block 11
- with the proposal of A for block 11, a new proposer turn starts, allowing validator B to make a proposal
- since the proposal 11 is not yet confirmed and processed, validator B's local state is at block 10; thus, validator B would decide to make another proposal for block 11, thus producing a competing proposal to the one made by validator A, adding extra load on the network

The aim of the proposal tracking is to avoid such competing proposals, essentially by blocking the proposal of block numbers that have been seen on the DAG, but not yet been confirmed.

**The Existing Mechanism**
Before this change, the events on the DAG kept track of the last proposed block number and validators are not allowed to re-propose the last seen block for a certain time-out window.

However, this mechanism has the following vulnerability:
- initially all validators have completed block 6 locally
- an honest validator A proposed a (valid) block 7
- a malicious validator B uses the next turn to propose a block 100, expected to be ignored
- an honest validator C sees the last proposed block 100, identifies it as invalid, checks its local state, and creates a proposal for block 7; this is competing with A's proposal, leading to inefficiencies

By this mechanism, a malicious validator can double its impact. The cause for this problem the lack of history when just tracking the last seen proposal.

**The New Mechanism**
The mechanism introduced in this PR removes the on-DAG tracking of seen proposals. Instead, every emitter locally tracks seen proposals, invalidating them as they time out. This information on pending proposals is used by emitters when deciding for which block a proposal is to be made.

**For reviewers**
The PR comprises three parts:
- the introduction of the `ProposalTracker` in the `inter` package
- the integration of the tracker into the `Emitter` in the `emitter` package
- the elimination of the ´LastSeenProposedBlock` field of the sync state (all the rest)

It is suggested to review these changes in the given order.

### DIFF
--- a/gossip/emitter/emitter.go
+++ b/gossip/emitter/emitter.go
@@ -109,6 +109,8 @@ type Emitter struct {
 	baseFeeSource BaseFeeSource
 
 	lastTimeAnEventWasConfirmed atomic.Pointer[time.Time]
+
+	proposalTracker inter.ProposalTracker
 }
 
 type BaseFeeSource interface {
--- a/gossip/emitter/hooks.go
+++ b/gossip/emitter/hooks.go
@@ -86,6 +86,9 @@ func (em *Emitter) OnNewEpoch(newValidators *pos.Validators, newEpoch idx.Epoch)
 			return updMetric(median, current, update, validatorIdx, newValidators)
 		})
 	em.payloadIndexer = ancestor.NewPayloadIndexer(PayloadIndexerSize)
+
+	// forget all seen proposals of the previous epoch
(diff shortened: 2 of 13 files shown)


LuisPH3 (approved):
:+1:

PR #246, opened 2025-05-29
feat(demo): add resume script for restarting existing blockchain nodes

toanalien:
I've created a new script demo/resume.sh that can resume the blockchain nodes without regenerating genesis or rebuilding binaries. The script:

1. Uses existing datadirs and binaries
2. Validates that datadirs exist before starting (exits with error if not found)
3. Starts each node with the same configuration as start.sh
4. Reconnects the nodes in a ring topology
5. Verifies peer connections at the end

Usage flow:

1. First time setup: Use `./start.sh`
2. Stop nodes: Use `./stop.sh`
3. Resume nodes: Use `./resume.sh`

### DIFF
--- a/demo/resume.sh
+++ b/demo/resume.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+cd $(dirname $0)
+. ./_params.sh
+
+set -e
+
+echo -e "\nResuming $N nodes:\n"
+
+for ((i=0;i<$N;i+=1))
+do
+    DATADIR="${PWD}/sonic$i.datadir"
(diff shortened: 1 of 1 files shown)


facuMH:
Hi @toanalien, 
Unfortunately we do not have established all the legal preconditions for accepting third-party contributions yet. Work on this is in progress, but there is no clarity yet on the desired and/or required policies. Thus, your contribution  cannot be accepted right now.

Nevertheless, I would like to thank you for your suggestion. We will analyse it and consider it for future releases of Sonic.

Thanks!

PR #247, opened 2025-05-29
adds minimal value validation for MaxEmptyBlockSkipPeriod

kjezek:
This PR adds validation for minimal value of MaxEmptyBlockSkipPeriod.

If the minimal value was  too low, it could happen that empty blocks are proposed continuously without giving validators a chance to propose real blocks between two empty blocks. 

This danger applies to the Single-Proposer protocol.

### DIFF
--- a/gossip/blockproc/verwatcher/version_number_test.go
+++ b/gossip/blockproc/verwatcher/version_number_test.go
@@ -39,7 +39,7 @@ func TestVersionNumber_AreOrderedFollowingSemanticVersioningRules(t *testing.T) 
 	for i := range len(versions) - 1 {
 		a := toVersionNumber(versions[i])
 		b := toVersionNumber(versions[i+1])
-		if !(a < b) {
+		if a >= b {
 			t.Errorf("unexpected result, %s < %s (%d.%d.%d.%d < %d.%d.%d.%d) does not hold",
 				versions[i], versions[i+1],
 				(a>>48)&0xffff, (a>>32)&0xffff, (a>>16)&0xffff, (a>>0)&0xffff,


HerbertJordan (dismissed):

opera/validate.go
```diff
@@ -160,6 +161,15 @@ const (
...
 	maximumMaxBlockGas = math.MaxInt64
+
+	// minEmptyBlockSkipPeriod sets the minimum time in seconds to produce an empty block when there are no transactions.
+	// The Single-Proposer protocol creates a block about every third frame.
+	// If MaxEmptyBlockSkipPeriod is too low, frequent empty block emissions may overwhelm pending proposals.
+	// Example: A block N exists; a validator proposes a block N+1.
+	// Before this block becomes accepted, a new empty block is created, which replaces the N+1 block.
+	// If MaxEmptyBlockSkipPeriod is too low, empty blocks are created too frequently,
+	// which starves proposals from validators.
+	minEmptyBlockSkipPeriod = 4
```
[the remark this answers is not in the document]
kjezek: good point, changed

opera/validate_test.go
```diff
@@ -238,10 +242,12 @@ func TestBlocksRulesValidation_DetectsIssues(t *testing.T) {
 }
 
 func TestBlocksRulesValidation_AcceptsValidRules(t *testing.T) {
+	const maxEmptyBlockSkipPeriod = inter.Timestamp(minEmptyBlockSkipPeriod * time.Second)
```
HerbertJordan, line 245: nit: now that `minEmptyBlockSkipPeriod` is in seconds, this conversion is misleading; it produces a duration of 4 billion seconds.

opera/validate_test.go
```diff
@@ -226,6 +226,10 @@ func TestBlocksRulesValidation_DetectsIssues(t *testing.T) {
 			rules: BlocksRules{MaxBlockGas: math.MaxUint64},
 			issue: "MaxBlockGas is too high",
 		},
+		"max empty block skip period too low": {
+			rules: BlocksRules{MaxEmptyBlockSkipPeriod: 3},
```
HerbertJordan, line 230: nit: fyi, these are 3 nano-seconds; 

In the past, we had people configuring times using the wrong time unit; we should be consistent;

HerbertJordan (dismissed):

opera/validate_test.go
kjezek: good point, thanks

HerbertJordan (approved):

PR #248, opened 2025-05-30
Refactor event signer tools

LuisPH3:
This PR is a preparation for the integration of the new randao_reveal.
The interfaces used by the emitter have been modified to facilitate access to the signing utility and the validator key. 

- Unused interface embeddings are removed.
- Affected llr tests (previously skipped)  are removed.

### DIFF
--- a/valkeystore/signer.go
+++ b/valkeystore/signer.go
@@ -12,8 +12,11 @@ import (
 
 //go:generate mockgen -source=signer.go -destination=signer_mock.go  -package=valkeystore
 
+// SignerAuthority is an interface for signing operations with a validator's key.
 type SignerAuthority interface {
+	// Sign signs a given digest using the validator's private key.
 	Sign(digest common.Hash) ([]byte, error)
+	// Returns the public key of the validator.
 	PublicKey() validatorpk.PubKey
 }
 
@@ -22,6 +25,8 @@ type signerAuthorityImpl struct {
 	pubkey  validatorpk.PubKey
 }
 
+// Constructs a new SignerAuthority using the provided keystore and public key.
+// The validator's private key is expected to be stored in the keystore
 func NewSignerAuthority(store KeystoreI, pubkey validatorpk.PubKey) SignerAuthority {
 	return &signerAuthorityImpl{
 		backend: store,
@@ -52,6 +57,8 @@ func (s *signerAuthorityImpl) PublicKey() validatorpk.PubKey {
 	return s.pubkey.Copy()
 }
 
+// VerifySignature verifies that the provided signature is valid for the given digest
+// using the provided public key. It returns true if the signature is valid, false otherwise.
 func VerifySignature(digest common.Hash, signature []byte, pubkey validatorpk.PubKey) bool {
 	if pubkey.Type != validatorpk.Types.Secp256k1 {
 		return false


LuisPH3:
I intend to add some tests to the signer itself

valkeystore/signer.go
```diff
@@ -3,41 +3,58 @@ package valkeystore
...
 	}
 	sigRS := sigRSV[:64]
 	return sigRS, err
 }
+
+func (s *signerAuthorityImpl) PublicKey() validatorpk.PubKey {
+	return s.pubkey.Copy()
+}
+
+func VerifySignature(digest common.Hash, signature []byte, pubkey validatorpk.PubKey) bool {
```
HerbertJordan, line 55: Should the signature type be strengthened by turning it into an array `[64]byte` to make the typing more restrictive?

gossip/basiccheck_test.go
```diff
@@ -256,3 +256,21 @@ func TestBasicCheckValidate(t *testing.T) {
...
+	me.SetCreationTime(e.CreationTime())
+	me.SetMedianTime(e.MedianTime())
+	me.SetPrevEpochHash(e.PrevEpochHash())
+	me.SetExtra(e.Extra())
+	me.SetGasPowerLeft(e.GasPowerLeft())
+	me.SetGasPowerUsed(e.GasPowerUsed())
+	me.SetPayloadHash(e.PayloadHash())
+	me.SetSig(e.Sig())
+	me.SetTxs(e.Txs())
+	return me
```
HerbertJordan, line 275: Is "SetPayload" missing?

gossip/heavycheck_test.go
HerbertJordan, line 1: This file seems like it should actually belong to the "eventcheck/heavycheck" package.

As a follow-up task we might want to modernize the implementation of these tests and move them to the respective package. I guess we can cover this as part of our event validation re-evaluation.

valkeystore/signer.go
```diff
@@ -3,41 +3,58 @@ package valkeystore
...
 
 	"github.com/0xsoniclabs/sonic/inter/validatorpk"
 	"github.com/0xsoniclabs/sonic/valkeystore/encryption"
 )
 
-type SignerI interface {
-	Sign(pubkey validatorpk.PubKey, digest []byte) ([]byte, error)
+//go:generate mockgen -source=signer.go -destination=signer_mock.go  -package=valkeystore
+
+type SignerAuthority interface {
```
HerbertJordan, line 15: Is the idea to avoid the shorter name `Signer` to not get confused with `types.Signer`?

HerbertJordan (dismissed):
Nice cleanup, thanks!

Please consider all comments suggestions.

valkeystore/signer.go
LuisPH3: Yes, additionally the signer object only contains signing logic, this object pairs it with a key, making it an authority.

valkeystore/signer.go
LuisPH3: There is already a package: inter/validatorpk/pubkey.go
Key type is defined by being a slice, I wanted to scope the changes to the signing tool. 
In the future, this should be cleaned and "merged" with the types.signer.

valkeystore/signer.go
facuMH: could you add documentation to this interface ? as to make its intent a little more explicit.

valkeystore/signer_test.go
```diff
@@ -0,0 +1,173 @@
...
+	_, publicKey := generateKeyPair(t)
+
+	t.Run("with invalid key format", func(t *testing.T) {
+		testKey := publicKey.Copy()
+		testKey.Type = validatorpk.Types.Secp256k1 + 1
+
+		signer := NewSignerAuthority(store, testKey)
+
+		var digest common.Hash
+		copy(digest[:], []byte("digest"))
```
facuMH, line 58: small nit: this could be a helper function that takes a string and returns a hash, as it is used in multiple tests.

LuisPH3: I am quite concerned about namespace pollution with many small tools. If the code is two lines and does not contain branches, I think it is more maintainable inlined. 
There will be some effort in the future to unify keys and signatures under strong types.

gossip/heavycheck_test.go
HerbertJordan: The name says LLR, but the tests it is covering are general event signature tests. When we work through the event validation code, we might have to eventually write similar tests again.

HerbertJordan (approved):

PR #254, opened 2025-05-30
Improve handling of early returning transactions

simonlechner:
This PR adds a snapshot and rollback to the transaction execution like it is done in the [t8n tool](https://github.com/ethereum/go-ethereum/blob/2a1784baef0e9d0c0641bc97d1a3478048de5c23/cmd/evm/internal/t8ntool/execution.go#L262) from ethereum.

### DIFF
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -296,9 +296,17 @@ func applyTransaction(
 		msg.BlobHashes = nil
 	}
 
+	isAllegro := evm.ChainConfig().IsPrague(blockNumber, evm.Context.Time)
+	var snapshot int
+	if isAllegro {
+		snapshot = statedb.Snapshot()
+	}
 	// Apply the transaction to the current state (included in the env).
 	result, err := core.ApplyMessage(evm, msg, gp)
 	if err != nil {
+		if isAllegro {
+			statedb.RevertToSnapshot(snapshot)
+		}
 		return nil, 0, result == nil, err
 	}
 	// Notify about logs with potential state changes.
--- a/evmcore/state_processor_test.go
+++ b/evmcore/state_processor_test.go
@@ -293,6 +293,88 @@ func TestApplyTransaction_InternalTransactionsSkipBaseFeeCharges(t *testing.T) {
 	}
 }
 
+func TestApplyTransaction_BlobHashesNotSupportedAndSkipped(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	state := state.NewMockStateDB(ctrl)
+	evm := vm.NewEVM(vm.BlockContext{}, state, &params.ChainConfig{}, vm.Config{})
+	gp := new(core.GasPool).AddGas(1000000)
+
+	msg := &core.Message{
+		From:       common.Address{1},
+		To:         &common.Address{2},
+		GasLimit:   21000,
+		GasPrice:   big.NewInt(1),
+		BlobHashes: []common.Hash{{0x01}},
+	}
+	usedGas := uint64(0)
+	receipt, gasUsed, skipped, err := applyTransaction(msg, gp, state, big.NewInt(1), nil, &usedGas, evm, nil)
+	require.ErrorContains(t, err, "blob data is not supported")
+	require.Nil(t, receipt)
+	require.Equal(t, uint64(0), gasUsed)
+	require.True(t, skipped)
+}
+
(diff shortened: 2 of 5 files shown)


evmcore/state_processor.go
```diff
@@ -296,9 +296,17 @@ func applyTransaction(
...
+	var snapshot int
+	if isAllegro {
+		snapshot = statedb.Snapshot()
+	}
 	// Apply the transaction to the current state (included in the env).
 	result, err := core.ApplyMessage(evm, msg, gp)
 	if err != nil {
+		if isAllegro {
+			statedb.RevertToSnapshot(snapshot)
+		}
```
HerbertJordan, line 309: Can you check whether we should also call `statedb.EndTransaction()` at this point?

evmcore/tx_pool.go
```diff
@@ -185,6 +185,8 @@ type TxPoolConfig struct {
 	GlobalQueue  uint64 // Maximum number of non-executable transaction slots for all accounts
 
 	Lifetime time.Duration // Maximum amount of time non-executable transaction are queued
+
+	DisableTxPoolValidation bool
```
HerbertJordan, line 189: Please document this option. In particular, mention that this is only to be used for integration tests.

evmcore/state_processor_test.go
```diff
@@ -293,6 +293,88 @@ func TestApplyTransaction_InternalTransactionsSkipBaseFeeCharges(t *testing.T) {
...
+				SkipFromEOACheck: true,
+			}
+
+			state.EXPECT().GetBalance(msg.From).Return(uint256.NewInt(1000000))
+			state.EXPECT().SubBalance(any, any, any)
+
+			if isPrague {
+				// Set up snapshot and revert expectations
+				state.EXPECT().Snapshot().Return(42)
+				state.EXPECT().RevertToSnapshot(42)
```
HerbertJordan, line 366: In this context the order of the execution of these expectations is essential. It might warrant the usage of `gomock.InOrder(...)`

tests/create_skipped_test.go
```diff
@@ -0,0 +1,111 @@
...
+				Value:    big.NewInt(0),
+				Data:     initCode,
+			}
+			tx := signTransaction(t, chainId, txData, sender)
+
+			preBalance, err := client.BalanceAt(t.Context(), sender.Address(), nil)
+			require.NoError(t, err)
+
+			receipt, err := net.Run(tx)
+			require.Error(t, err, "init code size too large")
```
HerbertJordan, line 56: I think you might wanted to sue "require.ErrorContains" here.

The run above takes a long time until it times out. Maybe you can find a way to reduce this, as it adds 100s of seconds to the test time.

HerbertJordan (changes requested):

evmcore/state_processor.go
simonlechner: I am not familiar with the `statedb.EndTransaction()` function, but I had a look at the [carmen implementation](https://github.com/0xsoniclabs/carmen/blob/fabde4233b62b2b5a0728098a097636b7b9b39f4/go/state/state_db.go#L1087).
Since the state should not and did not change a call to this function is not required, as it takes care of written storage slots and deleted accounts. In the last line it calls `s.resetTransactionContext()` which resets internal fields which should not have been set before the return of the failed transaction. 
I still added a call to `statedb.EndTransaction()` to ensure future change to the internal fields, which might be set even before failing, are reset correctly.

evmcore/state_processor_test.go
```diff
@@ -359,11 +362,14 @@ func TestApplyTransaction_ApplyMessageError_RevertsSnapshotIfPrague(t *testing.T
...
 			state.EXPECT().SubBalance(any, any, any)
+			state.EXPECT().EndTransaction()
 
 			if isPrague {
 				// Set up snapshot and revert expectations
-				state.EXPECT().Snapshot().Return(42)
-				state.EXPECT().RevertToSnapshot(42)
+				gomock.InOrder(
+					state.EXPECT().Snapshot().Return(42),
+					state.EXPECT().RevertToSnapshot(42),
```
HerbertJordan, line 371: Please add to the expectation that the mutating operation `SubBalance` is indeed between the creation of the snapshot and its revert.

HerbertJordan (approved):
Looks good, thanks!

facuMH (approved):

PR #255, opened 2025-05-30
Fix incorrectly formatted error strings

facuMH:
This PR fixes findings by `staticcheck` rule `ST1005`. 
This rule says that error strings must not begin with Capitalized letters, nor end with punctuation or new lines.
This rule will be activated with #250


HerbertJordan (changes requested):
Some of these error messages are intended to be directly consumed by end-users of the command line utilities. This is not great, as usually the command line utility should manage the full interaction with the user, but it is a common pattern in the code base.

We might want to apply exceptions for all messages intended to reach the user of the command line tool.

LuisPH3:
I have been before, we added an exception to prevent changing the command line interface to other processes.

LuisPH3 (approved):

HerbertJordan (approved):

PR #262, opened 2025-06-02
Add utility functions to run lists of transactions in parallel

HerbertJordan:
Adds a set of quality-of-life functions to the `IntegrationTestNetSession` to simplify the implementation of tests including the handling of multiple accounts, in particular for stress-tests.

The added utilities include:
- an `EndowAccounts` function enabling the parallel endowment of multiple accounts; 
- a `RunAll` function sending in a list of transactions and waiting for their receipts
- a `GetChainId` function, eliminating the need to query the chain-ID through an RPC call

### DIFF
--- a/tests/integration_test_net.go
+++ b/tests/integration_test_net.go
@@ -46,11 +46,24 @@ type IntegrationTestNetSession interface {
 	// mainly intended to provide funds to accounts for testing purposes.
 	EndowAccount(address common.Address, value *big.Int) (*types.Receipt, error)
 
+	// EndowAccounts sends the requested amount of tokens to each of the given
+	// accounts. This is a faster than calling EndowAccount for each account since
+	// multiple endowments may get bundled in the same block.
+	EndowAccounts(addresses []common.Address, value *big.Int) ([]*types.Receipt, error)
+
 	// Run sends the given transaction to the network and waits for it to be processed.
 	// The resulting receipt is returned.
 	Run(tx *types.Transaction) (*types.Receipt, error)
+
+	// RunAll sends the given transactions to the network and waits for them to be processed.
+	// The resulting receipts are returned.
+	RunAll(tx []*types.Transaction) ([]*types.Receipt, error)
+
 	// GetReceipt waits for the receipt of the given transaction hash to be available.
 	GetReceipt(txHash common.Hash) (*types.Receipt, error)
+
+	// GetReceipts waits for the receipts of the given transaction hashes to be available.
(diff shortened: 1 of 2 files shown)


tests/integration_test_net.go
```diff
@@ -46,12 +46,25 @@ type IntegrationTestNetSession interface {
...
+	// multiple endowments may get bundled in the same block.
+	EndowAccounts(addresses []common.Address, value *big.Int) ([]*types.Receipt, error)
+
 	// Run sends the given transaction to the network and waits for it to be processed.
 	// The resulting receipt is returned.
 	Run(tx *types.Transaction) (*types.Receipt, error)
+
+	// RunAll sends the given transactions to the network and waits for them to be processed.
+	// The resulting receipts are returned.
+	RunAll(tx []*types.Transaction) ([]*types.Receipt, error)
```
LuisPH3, line 60: instead of having two signatures, I propose that we have one method
`Run(tx ... *types.Transaction)`

tests/integration_test_net.go
```diff
@@ -46,12 +46,25 @@ type IntegrationTestNetSession interface {
...
+
 	// GetReceipt waits for the receipt of the given transaction hash to be available.
 	GetReceipt(txHash common.Hash) (*types.Receipt, error)
 
+	// GetReceipts waits for the receipts of the given transaction hashes to be available.
+	GetReceipts(txHash []common.Hash) ([]*types.Receipt, error)
```
LuisPH3, line 66: Idem:
use variadic arguments to avoid having too many symbols

tests/integration_test_net.go
```diff
@@ -62,10 +75,14 @@ type IntegrationTestNetSession interface {
 	// GetSessionSponsor returns the default account of the session. This account is used
 	// to sign transactions and pay for gas when using the Apply and EndowAccount methods.
 	GetSessionSponsor() *Account
+
 	// GetClient provides raw access to a fresh connection to the network.
 	// The resulting client must be closed after use.
 	GetClient() (*ethclient.Client, error)
 
+	// GetChainId returns the chain ID of the network.
+	GetChainId() *big.Int
```
LuisPH3, line 84: thanks

LuisPH3 (commented):
These are well needed improvements. 
please check comments:

tests/integration_test_net.go
HerbertJordan: That does not play well with the return type. Returning always a slice is annoying in the many cases where really just one transaction is executed.

PR #263, opened 2025-06-02
Apply DeMorgan's law

facuMH:
This PR fixes findings by `staticcheck` rule `QF1001`.
This rule reports possible applications of [De Morgan's Law](https://en.wikipedia.org/wiki/De_Morgan%27s_laws). 
In particular, there is only one reported finding, in a test for of version number ordering, where a positive proposition is preferred. 

This rule will be activated with https://github.com/0xsoniclabs/sonic/pull/250


LuisPH3 (approved):
this is much better

PR #270, opened 2025-06-03
Test: Internal transactions don't impact receipt index

simonlechner:
This PR adds an integration test to ensure that internal transactions do not alter the transaction index inside of the receipt.
Simple value transfers are performed with a epoch change in the middle, all produced blocks are checked if the transaction index in the block aligns with the one inside the receipt.

### DIFF
--- a/tests/transaction_receipt_idx_test.go
+++ b/tests/transaction_receipt_idx_test.go
@@ -22,29 +22,17 @@ func TestReceipt_InternalTransactionsDoNotChangeReceiptIndex(t *testing.T) {
 	require.NoError(t, err)
 	defer client.Close()
 
-	chainId, err := client.ChainID(t.Context())
-	require.NoError(t, err)
+	chainId := net.GetChainId()
 	gasPrice, err := client.SuggestGasPrice(t.Context())
 	require.NoError(t, err)
 	sender := makeAccountWithBalance(t, net, big.NewInt(1e18))
 
-	startBlockNumber, err := client.BlockNumber(t.Context())
-	require.NoError(t, err)
-
 	numSimpleTxs := 10
+	transactions := prepareTransactions(t, chainId, sender, gasPrice, numSimpleTxs)
 
-	// Send simple transactions
-	for nonce := range numSimpleTxs / 2 {
-		txData := &types.LegacyTx{
-			Nonce:    uint64(nonce),
-			Gas:      100000,
-			GasPrice: gasPrice,
-			To:       &common.Address{0x42},
-			Value:    big.NewInt(1),
-		}
-
-		tx := signTransaction(t, chainId, txData, sender)
-		err = client.SendTransaction(t.Context(), tx)
+	// Send first half of simple transactions
+	for i := range numSimpleTxs / 2 {
+		err = client.SendTransaction(t.Context(), transactions[i])
 		require.NoError(t, err)
 	}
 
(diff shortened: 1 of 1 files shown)


tests/transaction_receipt_idx_test.go
```diff
@@ -0,0 +1,102 @@
...
+	chainId, err := client.ChainID(t.Context())
+	require.NoError(t, err)
+	gasPrice, err := client.SuggestGasPrice(t.Context())
+	require.NoError(t, err)
+	sender := makeAccountWithBalance(t, net, big.NewInt(1e18))
+
+	startBlockNumber, err := client.BlockNumber(t.Context())
+	require.NoError(t, err)
+
+	numSimpleTxs := 10
```
facuMH, line 34:
```suggestion
	numSimpleTxs := uint64(10)
```
as everywhere it is used it is casted to `uint64` anyways.

tests/transaction_receipt_idx_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// Send simple transactions
```
facuMH, line 36: nit:
```suggestion
	// Send first half of simple transactions
```

tests/transaction_receipt_idx_test.go
```diff
@@ -0,0 +1,102 @@
...
+	// Send internal transaction (advance epoch)
+	contract, err := driverauth100.NewContract(driverauth.ContractAddress, client)
+	require.NoError(t, err)
+	txOpts, err := net.GetTransactOptions(&net.account)
+	require.NoError(t, err)
+	tx, err := contract.AdvanceEpochs(txOpts, big.NewInt(int64(1)))
+	require.NoError(t, err)
+	require.NotNil(t, tx)
+
+	// Send simple transactions
```
facuMH, line 60: nit:
```suggestion
	// Send second half of simple transactions
```

facuMH (commented):
Just a couple of minor suggestions.

tests/transaction_receipt_idx_test.go
```diff
@@ -0,0 +1,102 @@
...
+	for nonce := range numSimpleTxs / 2 {
+		txData := &types.LegacyTx{
+			Nonce:    uint64(numSimpleTxs/2 + nonce),
+			Gas:      100000,
+			GasPrice: gasPrice,
+			To:       &common.Address{0x42},
+			Value:    big.NewInt(1),
+		}
+
+		tx := signTransaction(t, chainId, txData, sender)
```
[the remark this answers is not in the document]
simonlechner: Getting the transactions in the same block was not an issue, but I think it is much cleaner to prepare all before sending.

tests/transaction_receipt_idx_test.go
```diff
@@ -0,0 +1,102 @@
...
+		err = client.SendTransaction(t.Context(), tx)
+		require.NoError(t, err)
+	}
+
+	// Use blocking call to ensure all transactions have been processed
```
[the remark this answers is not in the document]
simonlechner: I changed it to wait for the internal transaction receipt.

tests/transaction_receipt_idx_test.go
simonlechner: It is only used as an `uint64` once in the current version, I prefer to keep it as an `int` and cast it once.

tests/transaction_receipt_idx_test.go
```diff
@@ -0,0 +1,92 @@
...
+		receipt, err := client.TransactionReceipt(t.Context(), tx.Hash())
+		require.NoError(t, err)
+
+		// Check that the receipt index is equal to the transaction index
+		require.Equal(t, uint(i), receipt.TransactionIndex,
+			"Receipt index does not match transaction index for tx %d", i)
+	}
+}
+
+func prepareTransactions(t *testing.T, chainId *big.Int, sender *Account, gasPrice *big.Int, num int) []*types.Transaction {
```
facuMH, line 77: given the context of the `tests` package, I would suggest to inline this, but not a hard requirement for merging.

HerbertJordan (approved):
Looks good, thanks!

tests/transaction_receipt_idx_test.go
```diff
@@ -0,0 +1,92 @@
...
+	require.NotNil(t, tx)
+
+	// Send second half of simple transactions
+	for i := range numSimpleTxs / 2 {
+		err = client.SendTransaction(t.Context(), transactions[numSimpleTxs/2+i])
+		require.NoError(t, err)
+	}
+
+	// Wait for receipt of the internal transaction
+	receipt, err := net.GetReceipt(tx.Hash())
```
HerbertJordan, line 55: Does the block containing the "AdvanceEpoch" transaction indeed containing the epoch-sealing transactions?

Usually, the epoch-sealing transactions are at the begin of a block. This would place them after the transaction signaling the epoch change.

HerbertJordan (approved):

facuMH (approved):

PR #271, opened 2025-06-04
`GasLimit` to `MinTip`

facuMH:
This PR addresses a naming inconsistency in the `TxPool` that could lead to confusion about the field’s actual purpose.

The `gasPrice` field is used to represent the *minimum gas tip* required for a transaction to be accepted in the transaction pool, but its name suggested it was a general gas price value. In practice, it’s compared directly against the *transaction tip* in `validateTx` and `TxPool.Pending`.

To fix this misleading naming (and align with the naming used in Geth), it shall be renamed to `minTip`. The `GasPrice()` method in the `TxPool` interface shall also be renamed to `MinTip()` to reflect what it's actually returning. These changes help make the code easier to understand at a glance and avoid misleading assumptions.

A new command line flag (`--txpool.mintip`) is added as an alternative to the already existing `--txpool.pricelimit`.
`--txpool.pricelimit` has been documented as deprecated..

The corresponding changes have been made in the dummy_tx_pool implementation as well for consistency.
The function `SetGasPrice` is moved to `tx_poo_test.go` since that is it only used there. 

This addresses https://github.com/0xsoniclabs/sonic-admin/issues/149.

### DIFF
--- a/evmcore/tx_pool.go
+++ b/evmcore/tx_pool.go
@@ -176,7 +176,7 @@ type TxPoolConfig struct {
 	Journal   string           // Journal of local transactions to survive node restarts
 	Rejournal time.Duration    // Time interval to regenerate the local transaction journal
 
-	MinimumTip uint64 // Minimum tip to required to be accepted into the pool
+	MinimumTip uint64 // Minimum tip required to be accepted into the pool
 	PriceBump  uint64 // Minimum price bump percentage to replace an already existing transaction (nonce)
 
 	AccountSlots uint64 // Number of executable transaction slots guaranteed per account
--- a/tests/integration_test_net_test.go
+++ b/tests/integration_test_net_test.go
@@ -198,7 +198,7 @@ func TestIntegrationTestNet_CanStartWithCustomConfig(t *testing.T) {
 	// validation behaves as expected.
 	net := StartIntegrationTestNet(t, IntegrationTestNetOptions{
 		ModifyConfig: func(config *config.Config) {
-			// disable minimum tip check for non-local tx submission
+			// enable minimum tip check for local tx submission
 			config.TxPool.NoLocals = true
 			// increase minimum tip, default is 1
 			config.TxPool.MinimumTip = 10


LuisPH3:
Hi @thaarok,
I took the liberty of adding you for this review. The reason is that this PR changes the command line interface and the config files of the sonic client. 
It is important for us that this gets the operations' approval. Feel free to scale this question up.

evmcore/tx_pool.go
```diff
@@ -176,7 +176,7 @@ type TxPoolConfig struct {
 	Journal   string           // Journal of local transactions to survive node restarts
 	Rejournal time.Duration    // Time interval to regenerate the local transaction journal
 
-	PriceLimit uint64 // Minimum gas price to enforce for acceptance into the pool
+	MinimumTip uint64 // Minimum tip to required to be accepted into the pool
```
thaarok, line 179: `tip to required to be` -> `tip required to be`

tests/integration_test_net_test.go
```diff
@@ -198,10 +198,10 @@ func TestIntegrationTestNet_CanStartWithCustomConfig(t *testing.T) {
 	// validation behaves as expected.
 	net := StartIntegrationTestNet(t, IntegrationTestNetOptions{
 		ModifyConfig: func(config *config.Config) {
-			// enable minimum tip check for local tx submission
+			// disable minimum tip check for non-local tx submission
 			config.TxPool.NoLocals = true
```
thaarok, line 202: I think the original comment was correct.
Settings `NoLocals` to true tells the pool to handle local transactions the same way it handles remotes - which should enables checks for local txs, as the original comment claim.

evmcore/tx_pool_test.go
```diff
@@ -391,6 +392,28 @@ func deriveSender(tx *types.Transaction) (common.Address, error) {
...
+		// pool.priced is sorted by GasFeeCap, so we have to iterate through pool.all instead
+		drop := pool.all.RemotesBelowTip(price)
+		for _, tx := range drop {
+			pool.removeTx(tx.Hash(), true)
+		}
+		pool.priced.Removed(len(drop))
+	}
+
+	log.Info("Transaction pool price threshold updated", "price", price)
+}
```
thaarok, line 415: I am not sure, but I think once we remove the SetGasPrice method from the production code, we no longer need `TestTransactionPoolUnderpricing*` tests (which are the only usage of setMinTip()).

thaarok (approved):
Seems ok to me. Just a few notes mostly to comments.

evmcore/tx_pool_test.go
facuMH: I agree, the `repricing` and `underpricing` tests should be refactor and `setGasPrice` removed. But this should be done in a follow-up (https://github.com/0xsoniclabs/sonic-admin/issues/207) as it is not the focus of this PR.

LuisPH3 (approved):
the semantics of this flag now reflect its usage :+1:

PR #276, opened 2025-06-04
Refactor chainconfig construction

LuisPH3:
`ChainConfig` is a Geth structure which holds temporal information of the current set of features enabled.
`opera.Rules` do not hold any information about when such upgrades are introduced and are valid within the scope of the epoch. The historical information is stored in the `UpgradeHeight` collection of the key value store. 

There is a bug affecting historical runs, where a block in the past can be executed with later enabled but not yet active features. The reason is that sonic uses nanoseconds for timestamps and after receiving an update it makes it available from the current timestamp +1 (ns). The chainconfig is constructed, and each query conduced (mostly inside Geth code)  crops timestamps to seconds (`.Unix()` call). There are two suboptimal solutions:
- existing code stores nanoseconds inside the `ChainConfig` instance, computation is invalid. 
- we can crop the timestamp stored in the `UpgradeHeight` to seconds, but that will effectively roll back the block where the upgrade is first active, potentially including up to 4 blocks in the new semantics. 

The solution here relies on the fact, that within sonic, `Chainconfig` instances are transient, and should not be stored or reused between blocks: The new signature `CreateTransientEvmChainConfig` returns a `chainconfig` instance which satisfies the features at the provided block height. 

Changes are grouped by modules to facilitate review. 

this change is tested by #231


facuMH (commented):
This PR refactors how to get chain configs in a cleaner manner. 
Adds necessary tests. I have a couple of code comments though, nothing critical.
One design consideration, given all the convertions from `uint64` to `idx.Block`, would it be cleaner to user `uint64` as a third parameter for `CreateTransientEvmChainConfig` instead ? wdyt?

gossip/emitter/proposals_test.go
```diff
@@ -105,20 +105,19 @@ func TestWorldAdapter_GetEventPayload_ForwardsCallToGetExternalEventPayload(t *t
...
 
-	adapter := worldAdapter{world}
-	got := adapter.GetEvmChainConfig()
-	want := rules.EvmChainConfig(updateHeights)
-	require.Equal(want, got)
+	// adapter := worldAdapter{world}
+	// got := adapter.GetEvmChainConfig()
+	// want := rules.EvmChainConfig(updateHeights)
+	// require.Equal(want, got)
 }
```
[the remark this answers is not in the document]
LuisPH3: arg.. i missed this one. good catch

opera/rules.go
```diff
@@ -247,77 +247,68 @@ type UpgradeHeight struct {
...
-		}
+	// reverse iterate through the upgrade heights
+	for i := len(upgradeHeights) - 1; i >= 0; i-- {
+		height := new(big.Int).SetUint64(uint64(upgradeHeights[i].Height))
 
-		if cfg.CancunTime == nil && h.Upgrades.Sonic {
-			cfg.ShanghaiTime = timestamp
-			cfg.CancunTime = timestamp
-		}
+		if height.Cmp(new(big.Int).SetUint64(uint64(currentBlockHeight))) <= 0 {
```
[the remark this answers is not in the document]
HerbertJordan: It is not an equality but a `<=` comparison, but otherwise I agree, the detour through a `big.Int` seems unnecessary.

opera/rules.go
```diff
@@ -247,77 +247,68 @@ type UpgradeHeight struct {
...
-			cfg.ShanghaiTime = nil
-			cfg.CancunTime = nil
-		}
+			if upgrade.Allegro {
+				cfg.PragueTime = &timestampInThePast
+			}
 
-		if !h.Upgrades.Allegro {
-			cfg.PragueTime = nil
+			break
```
HerbertJordan, line 311: This assumes that the `upgradeHeights` are sorted. This requirement is not stated in the documentation, and should probably be checked -- or enforced -- if needed.

opera/rules_test.go
```diff
@@ -198,3 +202,40 @@ func TestRules_MinBaseFee_NoCopy_PreAllegro(t *testing.T) {
...
+			require.Equal(t, upgrades.Allegro, chainConfigAfterUpdate.IsPrague(new(big.Int).SetUint64(uint64(upgradeHeight.Height)), timestamp), "Allegro upgrade should match")
+
+			chainConfigBeforeUpdate := CreateTransientEvmChainConfig(chainID, []UpgradeHeight{upgradeHeight}, idx.Block(blockNumber-1))
+			require.NotNil(t, chainConfigBeforeUpdate, "chainConfig should not be nil")
+
+			if upgrades.Allegro {
+				require.True(t, chainConfigBeforeUpdate.IsCancun(new(big.Int).SetUint64(uint64(upgradeHeight.Height)), timestamp), "Before Allegro upgrade, Cancun should be true")
+				require.False(t, chainConfigBeforeUpdate.IsPrague(new(big.Int).SetUint64(uint64(upgradeHeight.Height)), timestamp), "Before Allegro upgrade, Prague should be false")
+			}
+		})
```
HerbertJordan, line 239: Factoring out common sub-expressions and giving them names would help readability here. In particular
```
new(big.Int).SetUint64(uint64(upgradeHeight.Height))
```
is something that could get a name and be reused.

Also, shortening the name of a few other variables might help readability.

opera/rules_test.go
```diff
@@ -198,3 +202,40 @@ func TestRules_MinBaseFee_NoCopy_PreAllegro(t *testing.T) {
 		t.Errorf("original and copied rules must be the same - shallow copy for preAllegro: got %d, want %d", got, want)
 	}
 }
+
+func TestCreateTransientEvmChainConfig(t *testing.T) {
```
HerbertJordan, line 206: There should be at least one test dealing with more than one entry in the `[]UpgradeHeight` slice. Also, one with an empty input list should be added to make clear what the expectation in such a case is.

Since this function essentially becomes the canonical interpretation of the `[]UpgradeHeight`, it should also be tested how the `heights` in the array are interpreted. If an upgrade is listed as being send in for block X, is it valid already for block X or for block X+1?

gossip/emitter/mock/world.go
HerbertJordan, line 1: This file should not be needed.

ethapi/api.go
```diff
@@ -1814,7 +1817,8 @@ func (s *PublicTransactionPoolAPI) sign(addr common.Address, tx *types.Transacti
 		return nil, err
 	}
 	// Request the wallet to sign the transaction
-	return wallet.SignTx(account, tx, s.b.ChainConfig().ChainID)
+	chainID := s.b.ChainConfig(0).ChainID
```
HerbertJordan, line 1820: It might really be worth adding a `ChainId()` function to the `Backend` interface. In all those changes the chain config needs to be created, which involves fetching the UpgradeHeight's list, just to get the chain ID.

ethapi/transaction_args.go
```diff
@@ -82,36 +82,23 @@ func (args *TransactionArgs) setDefaults(ctx context.Context, b Backend) error {
...
-		} else {
-			if args.MaxFeePerGas != nil || args.MaxPriorityFeePerGas != nil {
-				return errors.New("maxFeePerGas or maxPriorityFeePerGas specified but london is not active yet")
-			}
-			if args.GasPrice == nil {
-				price := b.SuggestGasTipCap(ctx, gasprice.AsDefaultCertainty)
-				price.Add(price, b.MinGasPrice())
-				args.GasPrice = (*hexutil.Big)(price)
-			}
+		if args.MaxPriorityFeePerGas == nil {
```
HerbertJordan, line 89: This change seems to ignore the case where `args.GasPrice != nil`.

This change also seems unrelated to the new handling of the ChainConfig.

ethapi/transaction_args.go
```diff
@@ -158,7 +145,7 @@ func (args *TransactionArgs) setDefaults(ctx context.Context, b Backend) error {
 		log.Trace("Estimate gas usage automatically", "gas", args.Gas)
 	}
 	if args.ChainID == nil {
-		id := (*hexutil.Big)(b.ChainConfig().ChainID)
+		id := (*hexutil.Big)(b.ChainConfig(0).ChainID)
```
HerbertJordan, line 148: We really should add `GetChainID` to the `Backend` interface.

gossip/emitter/scheduler/processor.go
```diff
@@ -39,8 +40,9 @@ type Chain interface {
 	// TODO: follow-up task - simplify this to a GetBlockHash(idx.Block) method.
 	evmcore.DummyChain
 
-	// GetEvmChainConfig returns the current chain configuration for the EVM.
-	GetEvmChainConfig() *params.ChainConfig
+	// GetEvmChainConfig returns the current chain configuration for the EVM at
+	// the given block height.
```
HerbertJordan, line 44:
```suggestion
	// GetEvmChainConfig returns the chain configuration for the EVM at the 
	// given block height.
```

integration/makegenesis/genesis_test.go
```diff
@@ -0,0 +1,107 @@
...
+	builder.AddBalance(address, big.NewInt(1e18))
+
+	finalizeBlockZero(t, builder, rules)
+
+	setCodeTx := makeSetCodeTransaction(t, new(big.Int).SetUint64(rules.NetworkID), key)
+	blockProc := DefaultBlockProc()
+
+	// With sonic features and attempting to execute setcode tx: log.Crit is called and program exits.
+	// https://github.com/0xsoniclabs/sonic/blob/03bd8b828db3ac51cb9b06254f9d33c75c12c8bd/gossip/blockproc/evmmodule/evm.go#L130
+	// TODO: investigate the suitability of containing log.Crit inside of block processing
```
HerbertJordan, line 40: Can you link this to an issue?

HerbertJordan (changes requested):
Thanks for improving this, and clearing up the relation between the network rules and the chain configuration.

I think there are still a number of changes that should be applied to this PR before it is ready to be merged. Please have a look at the comments.

opera/rules_test.go
```diff
@@ -198,3 +202,40 @@ func TestRules_MinBaseFee_NoCopy_PreAllegro(t *testing.T) {
...
+			require.True(t, chainConfigAfterUpdate.IsCancun(new(big.Int).SetUint64(uint64(upgradeHeight.Height)), timestamp))
+			require.Equal(t, upgrades.Allegro, chainConfigAfterUpdate.IsPrague(new(big.Int).SetUint64(uint64(upgradeHeight.Height)), timestamp), "Allegro upgrade should match")
+
+			chainConfigBeforeUpdate := CreateTransientEvmChainConfig(chainID, []UpgradeHeight{upgradeHeight}, idx.Block(blockNumber-1))
+			require.NotNil(t, chainConfigBeforeUpdate, "chainConfig should not be nil")
+
+			if upgrades.Allegro {
+				require.True(t, chainConfigBeforeUpdate.IsCancun(new(big.Int).SetUint64(uint64(upgradeHeight.Height)), timestamp), "Before Allegro upgrade, Cancun should be true")
+				require.False(t, chainConfigBeforeUpdate.IsPrague(new(big.Int).SetUint64(uint64(upgradeHeight.Height)), timestamp), "Before Allegro upgrade, Prague should be false")
+			}
```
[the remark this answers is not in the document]
LuisPH3: A couple of lines before, the chainconfig is created for a block earlier than the upgrade. this query should not accept prague.
I added some comments to make the test intention more clear.

opera/rules.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/276/commits/cdf23e7c33522bef43d4c8f376719ef324541efb

UpgradeHeights are sorted, to be conservative.

opera/rules_test.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/276/commits/cdf23e7c33522bef43d4c8f376719ef324541efb

gossip/emitter/scheduler/processor.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/276/commits/008378c4f66f5a8d9dd90946d428a8d2fcf63ff7

gossip/emitter/proposals_test.go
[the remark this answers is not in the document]
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/276/commits/008378c4f66f5a8d9dd90946d428a8d2fcf63ff7

integration/makegenesis/genesis_test.go
LuisPH3: https://github.com/0xsoniclabs/sonic-admin/issues/208

ethapi/transaction_args.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/276/commits/60e72e5da1d2c96cea1e8568b959a0524a228ffa

ethapi/api.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/276/commits/60e72e5da1d2c96cea1e8568b959a0524a228ffa

gossip/emitter/scheduler/processor.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/276/commits/86752d7a7beeb9f1378906187b3a093ace5f9ca8

ethapi/transaction_args.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/276/commits/0a70857242e8cbb6731faed0860709c4ac668a99

opera/rules.go
```diff
@@ -222,6 +225,7 @@ type Upgrades struct {
 	// -- Sonic Chain Hard Forks --
 	Sonic   bool // < launch version of the Sonic chain, introducing Cancun features
 	Allegro bool // < first hard fork of the Sonic chain, introducing Prague features
+	Brio    bool // < second hard fork of the Sonic chain, introducing Osaka features
```
HerbertJordan, line 228: A bit surprised to find this in this PR, but I guess it is fine to introduce this with this PR. 

What should be added in a follow-up PR (see #243 as an example):
- in the rule validation, as a hard-fork, it should not be allowed to disable it one it is enabled
- coverage in the `marshal_test.go`
- coverage in the `rules_test.go`

HerbertJordan (approved):

HerbertJordan (approved):

HerbertJordan (dismissed):

PR #278, opened 2025-06-04
Fix race-condition on test flag for transaction pool validation switch

HerbertJordan:
This change fixes a race condition in the propagation of the transaction-pool-validation disabling flag.

Before this PR, a global constant defining the default value of this flag was updated multiple times while running integration tests. With this change, this race condition is eliminated.

This is part of an effort to re-establish support for running race-detection tests on our integration tests.

### DIFF
--- a/cmd/sonicd/app/app.go
+++ b/cmd/sonicd/app/app.go
@@ -3,7 +3,7 @@ package app
 import (
 	"os"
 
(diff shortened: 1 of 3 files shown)


simonlechner (approved):
I moved the flag definition from the `flags.go` file to not make it as apparent, I didn't know that this would lead to race condition. Thanks for the fix!

PR #279, opened 2025-06-05
Search for internal block-sealing transactions

HerbertJordan:
Update to look for internal transactions.


simonlechner (approved):
Looks like I misunderstood epoch changes and internal transaction. The changes look good. At first I thought that adding endowments in a blocking way will not get them in the same block, but it is sufficient to get one in the same block.

PR #282, opened 2025-06-06
Accept zero-hashes as the state root of empty accounts

HerbertJordan:
This PR updates the integration test checking witness proofs to be able to handle zero-hashes being reported in proofs for empty accounts. This is a feature recently added to Carmen ([#23](https://github.com/0xsoniclabs/carmen/pull/23)).


LuisPH3 (approved):
it makes kind of sense

PR #285, opened 2025-06-06
Refactor txpool reprice tests 

facuMH:
As mentioned in https://github.com/0xsoniclabs/sonic/pull/271#discussion_r2126929274, the function `setMinTip` is only used in the repricing tests, and those repricing tests only tests the correct functionality of `setMinTip`. 

This PR removes unused functions `setMinTip` and `RemotesBelowTip` (which is only used in `setMinTip`), and refactors the repricing tests into two new tests.

New tests have been introduced instead:
- `TestTransactionPool_RejectsUnderTippedTransactions`
- `TestTransactionPool_AcceptsUnderTippedLocals`
Which as their names state, check that acceptance of transactions respects the min tip of the transaction pool for non local transactions.

### DIFF
--- a/evmcore/tx_pool.go
+++ b/evmcore/tx_pool.go
@@ -2023,18 +2023,6 @@ func (t *txLookup) RemoteToLocals(locals *accountSet) int {
 	return migrated
 }
 
-// RemotesBelowTip finds all remote transactions below the given gas price threshold.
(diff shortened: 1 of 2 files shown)


thaarok (approved):
Seems ok to me.

PR #288, opened 2025-06-10
Fix termination of parent-hash update transaction

HerbertJordan:
This PR adds a missing transaction termination call to complete the internal transaction keeping track of block hashes.

This feature has been introduced as part of the Prague feature package and is not yet actively used on the network. Without the call to `EndTransaction`, effects of the hash-tracking transaction are visible to the subsequent transaction in the same block. For instance, accessed accounts or value slots would be considered warm. Also, if there is no subsequent transactions, clean-up steps would get skipped. In particular, empty accounts touched by the internal transaction would not get cleaned up as expected.

### DIFF
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -91,6 +91,7 @@ func (p *StateProcessor) Process(
 	// execute EIP-2935 HistoryStorage contract.
 	if p.config.IsPrague(blockNumber, time) {
 		ProcessParentBlockHash(block.ParentHash, vmenv)
+		statedb.EndTransaction()
 	}
 
 	// Iterate over and process the individual transactions
@@ -137,6 +138,7 @@ func (p *StateProcessor) BeginBlock(
 	// execute EIP-2935 HistoryStorage contract.
 	if p.config.IsPrague(blockNumber, time) {
 		ProcessParentBlockHash(block.ParentHash, vmEnvironment)
+		stateDb.EndTransaction()
 	}
 
 	return &TransactionProcessor{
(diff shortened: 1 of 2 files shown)


LuisPH3 (commented):
what do you think about adding this call inside the function?

```
func ProcessParentBlockHash(prevHash common.Hash, evm *vm.EVM) {
	msg := &core.Message{
		From:      params.SystemAddress,
		GasLimit:  30_000_000,
		GasPrice:  common.Big0,
		GasFeeCap: common.Big0,
		GasTipCap: common.Big0,
		To:        &params.HistoryStorageAddress,
		Data:      prevHash.Bytes(),
	}

	txContext := NewEVMTxContext(msg)
	evm.SetTxContext(txContext)

	evm.StateDB.AddAddressToAccessList(params.HistoryStorageAddress)
	_, _, _ = evm.Call(msg.From, *msg.To, msg.Data, 30_000_000, common.U2560)
	evm.StateDB.Finalise(true)
}
```
There is access to `StateDB as well.

HerbertJordan:
> what do you think about adding this call inside the function?

I tried to add it there initially, but the `evm.StateDB` available in there is of type [vm.StateDB](https://github.com/0xsoniclabs/go-ethereum/blob/2400937cc3b1c370ea8735d7af86eac621e4eb70/core/vm/interface.go#L30C1-L85C1) while the one we need access to the extended version of the [state.StateDB](https://github.com/0xsoniclabs/sonic/blob/3b58d61d04210f438888c21dc37b5de5a9e67eed/inter/state/adapter.go#L13-L30) defined in Sonic.

To do the `EndTransaction` call inside the `ProcessParentBlockHash` function it would require an type cast.

PR #292, opened 2025-06-10
Introduce brio Hardfork

LuisPH3:


kjezek (approved):
looks good to me, thanks

opera/validate.go
```diff
@@ -261,5 +261,10 @@ func validateUpgrades(upgrade Upgrades) error {
 
 	// The SingleProposerBlockFormation feature can be freely modified.
 
+	if upgrade.Brio && !upgrade.Allegro {
+		//nolint:staticcheck // ST1005: allow capitalized error message to preserve proper name
+		issues = append(issues, errors.New("Brio upgrade requires Allegro"))
+	}
+
```
HerbertJordan, line 268: There is a new situation with "Brio" that was not there before: it should also check that if "Brio" was enabled before the update, it should also be enabled after. To check this, the validate function needs two parameters: "before" and "after".

Please create an issue for 2.3 for this to remind us about this required check.

HerbertJordan (approved):

opera/validate.go
LuisPH3: https://github.com/0xsoniclabs/sonic-admin/issues/216

PR #293, opened 2025-06-10
Move end of transaction into ProcessParentBlockHash function

HerbertJordan:
Moves the signaling of the end of the transaction updating the parent block hash into the respective helper function.

Signaling the end of the transaction was introduced by #288 for the scheduler and block processor, but supporting it in the API handlers was overlooked. This PR adds support for those use cases as well.


LuisPH3 (approved):
As discussed, all these concepts could benefit from a migration to the new carmen db api

PR #295, opened 2025-06-11
Update `validateAuth`

facuMH:
This PR introduces changes in the `validateAuth` function, to make the expected behaviour independent of the order of admission of authorizations, improving the predictability of the resulting state and the usability of the system with it.

This PR brings forth the changes introduced in geth by https://github.com/ethereum/go-ethereum/pull/31373. Not all changes are taken, only the fix in `validateAuth` and the relevant test cases. 

#### Proposed changes
The queued/pending transactions issued from any authority of the transaction being validated have to be counted, allowing one transaction from accounts submitting an authorization.

As noted by the original PR:
- The expected behavior was explicitly forbidden in the case "reject-delegation-from-pending-account", I believe that this behavior conflicts to the definition of the limitation, and it is removed in this PR. The expected behavior is tested in "accept-authorization-from-sender-of-one-inflight-tx".
- Replacement tests have been separated to improve readability of the acceptance test.

### DIFF
--- a/evmcore/tx_pool_test.go
+++ b/evmcore/tx_pool_test.go
@@ -688,11 +688,25 @@ func TestSetCodeTransactions(t *testing.T) {
 		},
 		"allow setcode tx with queued tx from delegated account": {
 			test: func(t *testing.T, pool *TxPool) {
-				if err := pool.addRemoteSync(setCodeTx(1, keyC, []unsignedAuth{{0, keyA}})); err != nil {
-					t.Fatalf("failed to add with remote setcode transaction: %v", err)
+				if err := pool.addRemoteSync(pricedTransaction(1, 100000, big.NewInt(10), keyC)); err != nil {
+					t.Fatalf("failed to add legacy transaction: %v", err)
 				}
 				if err := pool.addRemoteSync(setCodeTx(1, keyB, []unsignedAuth{{1, keyC}})); err != nil {
-					t.Fatalf("failed to add conflicting delegation: %v", err)
+					t.Fatalf("failed to add non conflicting delegation transaction: %v", err)
+				}
+			},
+			queued: 2,
+		},
+		"reject setcode tx with more than one queued tx from delegated account": {
+			test: func(t *testing.T, pool *TxPool) {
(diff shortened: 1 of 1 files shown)


evmcore/tx_pool_test.go
```diff
@@ -643,6 +686,17 @@ func TestSetCodeTransactions(t *testing.T) {
 			},
 			pending: 1,
 		},
+		"allow setcode tx with queued tx from delegated account": {
+			test: func(t *testing.T, pool *TxPool) {
+				if err := pool.addRemoteSync(setCodeTx(1, keyC, []unsignedAuth{{0, keyA}})); err != nil {
+					t.Fatalf("failed to add with remote setcode transaction: %v", err)
+				}
+				if err := pool.addRemoteSync(setCodeTx(1, keyB, []unsignedAuth{{1, keyC}})); err != nil {
+					t.Fatalf("failed to add conflicting delegation: %v", err)
```
LuisPH3, line 695: non conflicting? if this should not happen, then it is not a conflict

evmcore/tx_pool_test.go
LuisPH3, line 1: commit message introducing this change is not very helpful. 
I am trying to figure out what is this test testing

LuisPH3 (changes requested):
the last commit adds testing, i pressume for coverage reasons, we need to talk about it because I cannot figure out what do we need to test.

evmcore/tx_pool_test.go
facuMH: you are correct, this error message should say "non-conflicting". message fixed in https://github.com/0xsoniclabs/sonic/pull/295/commits/611bac82fc9309ea24183df6b00fa9af579e59bd

evmcore/tx_pool_test.go
facuMH: All other test cases only check for transactions in the `pending` list, but not in `queued`. The intention was to test that a delegation can be accepted even when there is one queued transaction from that same account. A new test is added in https://github.com/0xsoniclabs/sonic/pull/295/commits/611bac82fc9309ea24183df6b00fa9af579e59bd as a counterpart, rejecting a delegation if there are 2 queued transactions from the delegated account.

PR #298, opened 2025-06-11
Validate against turning off hard-forks.

LuisPH3:
This PR prevents from Hard-fork feature flags to be deactivated once active.


HerbertJordan (approved):
Looks good, thanks!

kjezek (approved):
this PR propagates current network rules to the validator to assess that the Brio update cannot be disabled.  This was not necessary for previous updates, because the validation is enabled with Allegro only, and for this reason, it was checked only that previous updates cannot be disabled, without the need to reason about the current state. 

Looks good, thanks.

PR #299, opened 2025-06-11
Restrict input of gas base fee computation to required fields

HerbertJordan:
This PR restricts the parameters required for computing base fees to the essential properties of the preceding block required to do so. The aim is to clarify the required input and to eliminate the implicit requirement of provisioning a full block header.


LuisPH3 (approved):
The proposed changes factor the block header fields required for basefee computation into a new type. 
The code benefits from scope and readability.

PR #300, opened 2025-06-11
Fix handling of block time

HerbertJordan:
While checking the alignment of block parameters between the scheduler and the block processor a difference in the used block time was discovered. The block processor was not using the time included in the proposal.

Before this PR, the time to be seen within transaction executions was set up [here](https://github.com/0xsoniclabs/sonic/blob/2221dd4be5e60f83cbd8f46b13f3c47bb86ef122/gossip/c_block_callbacks.go#L245-L246), by passing the [blockCtx](https://github.com/0xsoniclabs/sonic/blob/2221dd4be5e60f83cbd8f46b13f3c47bb86ef122/gossip/c_block_callbacks.go#L246) to the factory of the EVM-processor. This `blockCtx` contains the time of the atropos event, as it is set up [here](https://github.com/0xsoniclabs/sonic/blob/2221dd4be5e60f83cbd8f46b13f3c47bb86ef122/gossip/c_block_callbacks.go#L158), which is then used by the EVM as the block time. However, when running the single-proposer protocol, the block time stated within the proposal is expected to be used.

Since the `blockCtx.Time` is used multiple times between its definition an the creation of the EVM, it is not sufficient to update the time in the context once the proposal has been selected.

To fix this, this PR moves the proposal identification section (lines [193-242](https://github.com/0xsoniclabs/sonic/blob/2221dd4be5e60f83cbd8f46b13f3c47bb86ef122/gossip/c_block_callbacks.go#L193-L242)) further up in the code base, to be placed before the initialization of the [blockCtx](https://github.com/0xsoniclabs/sonic/blob/2221dd4be5e60f83cbd8f46b13f3c47bb86ef122/gossip/c_block_callbacks.go#L158) and the [determination](https://github.com/0xsoniclabs/sonic/blob/2221dd4be5e60f83cbd8f46b13f3c47bb86ef122/gossip/c_block_callbacks.go#L153) of the time to be used in the block context.

As a side-effect, this PR also makes sure that the block time is always increasing.

**Not to reviewers:**
The main change of this PR is the movement of the proposal identification code further up inside the same function. This should be reviewed first. All the rest is test infrastructure testing the correct handling and reproducibility of block parameters.

### DIFF
--- a/tests/block_parameters_test.go
+++ b/tests/block_parameters_test.go
@@ -1,0 +1,133 @@
+package tests
+
+import (
+	"fmt"
+	"math/big"
+	"testing"
+
+	"github.com/0xsoniclabs/sonic/opera"
+	"github.com/0xsoniclabs/sonic/tests/contracts/block_parameters"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBlockParameters_BlockHeaderMatchesObservableBlockParameters(t *testing.T) {
+	hardForks := map[string]opera.Upgrades{
+		"sonic": {
+			Berlin:  true,
+			London:  true,
+			Llr:     false,
+			Sonic:   true,
+			Allegro: false,
+		},
+		"allegro": {
+			Berlin:  true,
+			London:  true,
+			Llr:     false,
+			Sonic:   true,
+			Allegro: true,
+		},
+	}
+
+	for name, upgrades := range hardForks {
+		t.Run(name, func(t *testing.T) {
+			for _, singleProposer := range []bool{false, true} {
+				t.Run(fmt.Sprintf("single_proposer=%t", singleProposer), func(t *testing.T) {
+					upgrades := upgrades
+					upgrades.SingleProposerBlockFormation = singleProposer
+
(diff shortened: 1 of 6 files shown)


tests/block_parameters_test.go
```diff
@@ -0,0 +1,133 @@
...
+			Sonic:   true,
+			Allegro: false,
+		},
+		"allegro": {
+			Berlin:  true,
+			London:  true,
+			Llr:     false,
+			Sonic:   true,
+			Allegro: true,
+		},
```
LuisPH3, line 31: can we enable single block proposal for this test? 
I should be capable of executing correct paramenters, right?

LuisPH3 (commented):
The

tests/block_parameters_test.go
HerbertJordan: Both of these configurations are tested with and without the single proposer mode.

LuisPH3 (dismissed):
The refactor of `c_block_callbacks` seems appropriate:
- initialized facilities are not used until after proposal selection.
   - logger, sealer...
- skip block logic is delayed after proposal resolution, this has no effect on block
- chain config, evm instance are constructed with the proposal-corrected values, for the entirety of the block processing.

Testing seems adequate, new contract can be useful to test hardforks.

facuMH (dismissed):
looks good, just a tiny comment.

facuMH (approved):

PR #301, opened 2025-06-12
Make time of last emitted event atomic

facuMH:
This PR fixes a data race condition in the emitter's `prevEmitedAtTime` variable, accessed during initialization and in a ticker.


HerbertJordan (approved):
Looks good, thanks.

PR #302, opened 2025-06-12
Test: Check skipped transaction index in receipt

simonlechner:
This PR adds an integration test checking that a skipped transaction does not alter the index of transactions inside a receipt.


tests/transaction_receipt_idx_test.go
```diff
@@ -120,3 +120,111 @@ func getSenderOfTransaction(
...
+			after = receipt.BlockNumber.Uint64()
+		}
+	}
+
+	// Make sure the skipped transaction was included in the block by checking the balance of the sender.
+	balanceBefore, err := client.BalanceAt(t.Context(), senderSkipped.Address(), big.NewInt(int64(before)))
+	require.NoError(t, err)
+	balanceAfter, err := client.BalanceAt(t.Context(), senderSkipped.Address(), big.NewInt(int64(after)))
+	require.NoError(t, err)
+	require.Greater(t, balanceBefore.Uint64(), balanceAfter.Uint64(),
```
facuMH, line 208: greater or equal ?

facuMH (approved):
This PR introduces a tests that verifies that a skipped transaction does not alter the index of the receipts for non-skipped transactions.

HerbertJordan (approved):
Thanks for verifying that this is handled correctly.

tests/transaction_receipt_idx_test.go
simonlechner: Some amount of balance needs to be deducted since the gas of the skipped transaction is bought, so strictly greater.

PR #308, opened 2025-06-12
Fix error handling in debug_traceBlock - use error field

thaarok:
This is main-branch variant of #307

Fixes debug_traceBlock for situation where tracing of a single tx fails:

**Before fix:**
```
curl 'http://localhost/' -X POST -H 'Content-Type: application/json' --data '{     
    "jsonrpc": "2.0",
    "method": "debug_traceBlockByNumber",
    "params": [
        "0x200873b",
        {
            "tracer": "callTracer",
            "tracerConfig": {
                "withLog": true
            }
        }
    ],
    "id": 1
}' | jq
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -32000,
    "message": "tracing failed: blob transaction missing blob hashes"
  }
}
```

### DIFF
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -2329,17 +2329,18 @@ func (api *PublicDebugAPI) traceBlock(ctx context.Context, block *evmcore.EvmBlo
 		}
 		res, err := api.traceTx(ctx, tx, msg, txctx, block.Header(), statedb, config, nil)
 		if err != nil {
-			return nil, err
+			results[i] = &txTraceResult{TxHash: tx.Hash(), Error: err.Error()}
+			resultsLength += len(err.Error())
+		} else {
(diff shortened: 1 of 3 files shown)


ethapi/api.go
```diff
@@ -2329,7 +2329,8 @@ func (api *PublicDebugAPI) traceBlock(ctx context.Context, block *evmcore.EvmBlo
 		}
 		res, err := api.traceTx(ctx, tx, msg, txctx, block.Header(), statedb, config, nil)
 		if err != nil {
-			return nil, err
+			results[i] = &txTraceResult{TxHash: tx.Hash(), Error: err.Error()}
+			continue
```
jenikd, line 2333: Before continue, `statedb.EndTransaction()` should be called as it is in the end of this for loop. Also limits for response size are not takiing efect because of this.

jenikd (dismissed):

PR #313, opened 2025-06-16
Improve access to transactions in event payloads

HerbertJordan:
This PR replaces the previous `Tx()` function defined on event payloads with two new functions:
- `Transactions()` retrieving transactions carried by the event, independently of whether these are on the top level or nested in the a proposal
- `TransactionToMeter()` retrieving transactions for which individual validators are required to pay gas costs for in the distributed proposal mode

As part of this update, various logger and monitoring features got updated to report proper values for the number of processed transactions by events, independently of the user block formation mode (distributed or single-proposer).


LuisPH3 (approved):
I think it is way more readable now, it introduces a new naming for two semantics that we already had in a hidden form.

PR #316, opened 2025-06-16
Add permissibility check for transactions in block processor

HerbertJordan:
This PR introduces an extra permissibility check as part of the block processor making sure that undesired transaction types are getting prevented from being included in blocks.

This is a defensive coding layer added as a second line of defense in case event-validation fails. It should also be the responsibility of the block formation process to check transaction permissibility.

### DIFF
--- a/gossip/c_block_callbacks.go
+++ b/gossip/c_block_callbacks.go
@@ -213,7 +213,7 @@ func consensusCallbackBeginBlockFn(
 				}
 
 				// Filter invalid transactions from the proposal.
-				proposal.Transactions = filterInvalidTransactions(
+				proposal.Transactions = filterNonPermissibleTransactions(
 					proposal.Transactions, &es.Rules, log.Root(), invalidTxsMeter,
 				)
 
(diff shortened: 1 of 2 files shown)


gossip/c_block_callbacks.go
```diff
@@ -656,3 +666,84 @@ func extractProposalForNextBlock(
...
+// rules. This is the canonical validation set for transactions before they are
+// accepted in blocks. It is used to verify static properties according to the
+// current network rules.
+//
+// Transactions rejected by this step are considered invalid transactions.
+// Honest validators should never produce blocks containing invalid transactions.
+//
+// Valid transactions may still be rejected by the block processor due to nonce
+// or balance issues. In such cases, the transaction is considered a skipped
+// transaction. Skips should be minimized, but can not be completely avoided.
```
LuisPH3, line 707: This paragraph attempts to make a difference between dropping the tx here or later, but both cases lead to skipped txs. In the sense of not being reflected by a receipt after block execution. 
I do not see the need to create a new semantic for this case, maybe it is better to reject the translations later in block processing. This would also guarantee that different entry points to processor have the same behavior (RPC, genesis)

LuisPH3 (commented):
The exclusion of tx types out of their allowed block time is important, and I am even surprised that it was not enforced before. 
Please see comment

gossip/c_block_callbacks.go
LuisPH3: wouln't it be better to execute this check here:
https://github.com/0xsoniclabs/sonic/blob/f1f600d5308ff0805728ea034ebf9f26e6d48fcf/evmcore/state_processor.go#L71
and report it as skipped directly?  semantically, should be the same, right?

HerbertJordan: I will attempt to move it to the state processor.

LuisPH3 (approved):
The changes proposed protect the block execution from including transactions not allowed but the current system rules. 
This change shift the responsibility from the consensus event validation to the block processor, removing delegation of vulnerable decisions into another module.

PR #319, opened 2025-06-17
Continue processing block after a failing transaction

simonlechner:
The `log.Crit` call in the `Execute` function of the `OperaEVMProcessor` has been removed, by moving the error handling into the `Process` function of the `evmProcessor`. This allows to handle failing transactions separately and not on a block level. Failing transactions are ignored and the block processing continues with the next transaction.

### DIFF
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -80,7 +80,6 @@ func (p *StateProcessor) Process(
 	var (
 		gp           = new(core.GasPool).AddGas(block.GasLimit)
 		receipt      *types.Receipt
-		skip         bool
 		header       = block.Header()
 		time         = uint64(block.Time.Unix())
 		blockContext = NewEVMBlockContext(header, p.bc, nil)
@@ -105,12 +104,7 @@ func (p *StateProcessor) Process(
 		}
 
 		statedb.SetTxContext(tx.Hash(), i)
-		receipt, _, skip, err = applyTransaction(msg, gp, statedb, blockNumber, tx, usedGas, vmenv, onNewLog)
-		if skip {
-			skipped = append(skipped, uint32(i))
-			receipts = append(receipts, nil)
-			continue
-		}
+		receipt, _, err = applyTransaction(msg, gp, statedb, blockNumber, tx, usedGas, vmenv, onNewLog)
 		if err != nil {
 			log.Error("Failed to apply transaction", "tx", tx.Hash().Hex(), "err", err)
 			skipped = append(skipped, uint32(i))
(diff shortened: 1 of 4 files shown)


evmcore/state_processor_test.go
```diff
@@ -187,12 +188,11 @@ func TestProcess_DetectsTransactionThatCanNotBeConvertedIntoAMessage(t *testing.
...
-			receipts, logs, skipped, err := process(block, state, vmConfig, usedGas, nil)
-			require.ErrorContains(err, "invalid transaction v, r, s values")
+			receipts, logs, skipped := process(block, state, vmConfig, usedGas, nil)
 
-			require.Nil(receipts)
-			require.Nil(logs)
-			require.Nil(skipped)
+			require.ElementsMatch(receipts, []*types.Receipt{nil})
+			require.Empty(logs)
+			require.Empty(skipped)
```
HerbertJordan, line 195: In this case `skipped` should not be empty, right?

evmcore/state_processor_test.go
```diff
@@ -258,6 +257,67 @@ func TestProcess_TracksParentBlockHashIfPragueIsEnabled(t *testing.T) {
...
+	// Process the block
+	usedGas := new(uint64)
+	receipts, logs, skipped := processor.Process(block, state, vm.Config{}, usedGas, nil)
+
+	require.Len(t, receipts, 2)
+	require.Nil(t, receipts[0])
+	require.NotNil(t, receipts[1])
+
+	require.Empty(t, logs)
+	require.Empty(t, skipped)
```
facuMH, line 318: looking at this, it gives the impression that there this change introduces a somewhat hidden semantic, which is "an erroneus transaction - which has no recepit, nor is counted as skipped, but it also does not produce a failed receipt", was this the intent?

evmcore/state_processor.go
```diff
@@ -97,7 +98,9 @@ func (p *StateProcessor) Process(
 	for i, tx := range block.Transactions {
 		msg, err := TxAsMessage(tx, signer, header.BaseFee)
 		if err != nil {
-			return nil, nil, nil, fmt.Errorf("could not apply tx %d [%v]: %w", i, tx.Hash().Hex(), err)
+			log.Error("Failed to convert transaction to message", "tx", tx.Hash().Hex(), "err", err)
+			receipts = append(receipts, nil)
+			continue // skip this transaction, but continue processing the rest of the block
```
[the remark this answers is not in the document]
simonlechner: I had another look at `OperaEVMProcessor.Execute` function and how skipped transactions are handled and added the failed transactions to the skipped ones.

HerbertJordan (approved):
Looks good to me. I would suggest to also include @LuisPH3 into the review, as he originally identified the issue.

LuisPH3 (approved):

facuMH (approved):

facuMH (approved):

PR #324, opened 2025-06-18
Use event version 3 for allegro and 4 for single proposer

facuMH:
In order to update transaction validation in event checker, we need to be able to differentiate between `Sonic`, `Allegro` and `SingleProposer` in a context where we do not have network rules. 
To enable this differentiation, this PR makes the necessary changes to use events version 3 to represent `Allegro` events, and events of version 4 to rerepsent `SingleProposer` events.
 
This PR also moves the function `evmcore.IntrinsicGas` to the `basiccheck` package, as it is there that it is only ever used, and it should only be used there. A new function is added `intrinsicGasAllegro`, as an updated version of it, that also accounts for init code size of create transactions and set auth length cost.

Note for review: tests `TestEmitter_CreateEvent_CreatesCorrectEventVersion`, `TestEventPayloadSerialization`, `TestEventUnmarshalCSER_Version2Or3FailsIfHashOfEmptyPayloadIsIncluded` and `testSingleProposerProtocol_CanBeEnabledAndDisabled` are only modified to test for both `Sonic` and `Allegro` with their correspoding version 2 and 3. 

NOTE: the existing version of `evmcore.IntrinsicGas` assumed Homestead, Istanbul and Shanghai were active, but this was not documented. This assumptions are maintained in the new functions, but now are also documented.


facuMH:
This changes will not be needed

PR #326, opened 2025-06-20
Add gas limit to single-proposer mode

HerbertJordan:
This RP introduces a check for the admissible gas limit for block proposals in the single-proposer mode.

**For Reviewers**
This PR applies three modifications:
- it moves the formerly private [`GetEffectiveGasLimit`](https://github.com/0xsoniclabs/sonic/blob/561abb92b658a7bc5cfe67a10af776528ca42ab4/gossip/emitter/proposals.go#L325) function from the `emitter` package into the `inter` package to make it available to the `gossip` package
- it extends the EVM State-Processor's [Process](https://github.com/0xsoniclabs/sonic/blob/561abb92b658a7bc5cfe67a10af776528ca42ab4/evmcore/state_processor.go#L72) function by an explicit gas-limit parameter for running the given transactions, independent of the overall block limit. While the block limit must be consistent for all transactions in a block, and can be accessed within the EVM, this gas limit is specific for the execution of the concrete set of transactions passed to the EVM
- it updates the calls to the `Process` function in the scheduler, the block formation code, and in tests to provide this additional parameter

In particular, the [c_block_callbacks.go](https://github.com/0xsoniclabs/sonic/blob/04f24e1406bb3abd3eb8bb8c1ce2e68988e500d1/gossip/c_block_callbacks.go#L208-L211) file is extended with the computation of the effective gas limit for the current block.

### DIFF
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -70,7 +70,8 @@ func NewStateProcessor(config *params.ChainConfig, bc DummyChain) *StateProcesso
 // Future hard-forks may be used to clean up the rules and make them more
 // consistent.
 func (p *StateProcessor) Process(
-	block *EvmBlock, statedb state.StateDB, cfg vm.Config, usedGas *uint64, onNewLog func(*types.Log),
+	block *EvmBlock, statedb state.StateDB, cfg vm.Config, gasLimit uint64,
+	usedGas *uint64, onNewLog func(*types.Log),
 ) (
 	types.Receipts, []*types.Log, []uint32,
 ) {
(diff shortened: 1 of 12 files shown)


LuisPH3 (dismissed):
The pr corrects the metrics regarding gas used by block:
pre-internal, internal, and user transactions are executed in three different gas pools, which allows executing max block gas  internal and preinternal transactions (both)
and max block gas user transactions, this parameter is time-adjusted for single block proposals. 

Because of the distinction in semantics for the single block proposal argument, the block max block gas value cannot longer be used an a new paramenter needs to be wired.

LuisPH3 (approved):

PR #327, opened 2025-06-20
Replace outdated Fantom mentions by Sonic

facuMH:
This PR replaces outdated mentions of Fantom by Sonic.

### DIFF
--- a/opera/rules.go
+++ b/opera/rules.go
@@ -147,11 +147,11 @@ type EconomyRules struct {
 
 	// MinGasPrice defines a lower boundary for the gas price
 	// on the network. However, its interpretation is different
-	// in the context of the Fantom and Sonic networks.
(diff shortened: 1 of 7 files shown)


opera/rules.go
```diff
@@ -181,11 +181,11 @@ type EconomyRules struct {
 
 	// MinGasPrice defines a lower boundary for the gas price
 	// on the network. However, its interpretation is different
-	// in the context of the Fantom and Sonic networks.
+	// in the context of the Sonic and Sonic networks.
```
simonlechner, line 184:
```suggestion
	// in the context of the Sonic network.
```
Or keep the original (maybe in reverse order) as it is also the case for the fantom network.

simonlechner (commented):
Looks good except for one replacement.

simonlechner (approved):

PR #330, opened 2025-06-20
Remove redundant Time field from proposals

HerbertJordan:
This PR removes the redundant "Time" field from the proposal. 

In any valid event carrying a proposal, this field is required to be equivalent to the median time stored in the same event. Thus, this field is redundant and can be removed to reduce disk and network usage.


LuisPH3 (approved):
good cleanup.

LuisPH3 (approved):

PR #331, opened 2025-06-23
Make epoch and validators atomics 

facuMH:
Since the emitter's `epoch` index and `validators` tracker are updated in `OnNewEpoch` hook and read in `createEvent`, they need to be safe for concurrency.

This PR makes those two atomics to prevent race conditions.


HerbertJordan (approved):
Looks good to me. I think the missing code coverage can be ignored for this change, since (almost) all changes are 1:1 replacements of pre-existing code.

PR #332, opened 2025-06-23
Limit total size of transactions in events

HerbertJordan:
This PR introduces a load test for large transactions and establishes a protection against too large events being produced by the emitter.

**Load Test**
This PR introduces a new load test producing hundreds of large transactions (in terms of data size, each being 125 KB large) and sending them into the network with the aim to trigger the emission of an event exceeding the 10 MiB limit. 

This tests revealed that with high enough gas limits the distributed block formation mode is not protected against this. A corresponding fix is part of this PR.

**For Reviewers**
In order to reliably trigger large events, the transaction-submission mechanism of the integration test infrastructure needed to be updated to facilitate the parallel submission of transactions. The corresponding changes are covered in the `tests/integration_test_net.go`.

Beyond that, `tests/network_rules_update_test.go` was extended to support the retrieval of the current network rules from the network. The operation was grouped with other rule-related operations. However, these might be moved into the integration network infrastructure eventually.

The rest of the PR is the actual stress test (`tests/large_transactions_test.go`) and the fixes applied to the `gossip/emitter` implementation.

### DIFF
--- a/tests/large_transactions_test.go
+++ b/tests/large_transactions_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/0xsoniclabs/sonic/opera"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/params"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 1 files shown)


tests/large_transactions_test.go
```diff
@@ -0,0 +1,184 @@
...
+	transactions := []*types.Transaction{}
+	data := make([]byte, 125_000) // 125 KB of data, all zeros (cheapest)
+	for nonce := range uint64(numRounds) {
+		for i := range accounts {
+			tx := types.MustSignNewTx(
+				accounts[i].PrivateKey,
+				signer,
+				&types.AccessListTx{
+					ChainID:  chainId,
+					Gas:      125_000*10 + 21_000, // 125 KB of data + base gas
```
facuMH, line 164: maybe add a comment to clarify that `*10` is the price of zero bytes in data ? wdyt ?

facuMH (approved):
This PR extends the functionality of the integration test net to submit transactions in parallel, as well as getting the current rules. 
It also adds an upper limit for the size of an event and an integration test verifying this limit is enforced in a flooded network.

PR #333, opened 2025-06-24
Update license

facuMH:
This PR introduces the sonic LGPL to all sonic files that do not already have a `go-ethereum` license, and a script that automates this process. 

Notes to reviewers:
- To facilitate reviews, the https://github.com/0xsoniclabs/sonic/commit/99f36f2535e497f174f5de2c4e3f40f4b593e7f0 commit contains the 3 relevant files to the automation of this process. That is the license header in plain text, a go script that checks all files and adds the license if they need it, and a test for the scrip, with the 3 main cases (already has geth license, already has sonic license and does not have either).
- The second commit is only adding the header license to all relevant files. The license header is 15 lines plus one empty line after it, [this](https://github.com/0xsoniclabs/sonic/commit/0d5f391ec5c7dcd90e4bc1442d2724c7bfd760bf) commit adds it to 565 files, `565*16=9040` which es exactly the number of lines added by this commit. No lines are deleted, that means no geth license header has been removed.


.codecov.yml
```diff
@@ -1,3 +1,19 @@
+# Copyright 2025 Sonic Operations Ltd
+# This file is part of Sonic Client
```
b-scholz, line 2: It might be too late, but "This file is part of the Sonic Client." might read better. Both versions are fine.

b-scholz (approved):

facuMH:
This PR needs a bit of cleaning.

PR #334, opened 2025-06-24
Update license

facuMH:
This PR introduces the sonic LGPL to all sonic files that do not already have a go-ethereum license

Notes to reviewers:
The license header is 15 lines and must always be followed by one empty line. This PR adds the license header to 565 files, `566*16=9056` which es exactly the number of lines added by this commit. No lines are deleted, that means no geth license header has been removed.


HerbertJordan (commented):
I will review this once the script is merged and I can quickly verify this PR by running the script myself.

HerbertJordan (approved):
Looks good. Please add the command used to generate those to the PR description. I suspect something like
```
go run ./scripts/license -dir ./
```
Also: a direct follow-up task - add the check for the presence of header files to the github CI and a command to fix the scripts folder or the make file.

PR #336, opened 2025-06-24
Refactor in testing helper functions

facuMH:
This PR implements a refactor in two integrations tests helper functions.
It shall pass `t` instead of `require.Assertion` in test functions.


simonlechner (approved):
Nice to see the arguments are unified and `t` is now passed to all the test helper functions. Also learned that `testing.TB` is an interface that excepts testing, benchmarking and fuzzing pointers.

PR #337, opened 2025-06-24
Use `t.Context` in integration tests

facuMH:
The `testings` package offers a `context()`, this should be used in all tests. 
This PR replaces the uses of `context.Background()` for the testing alternative. 

This PR depends on https://github.com/0xsoniclabs/sonic/pull/336

### DIFF
--- a/tests/block_header_test.go
+++ b/tests/block_header_test.go
@@ -534,7 +534,6 @@ func getStateRoot(client *ethclient.Client, blockNumber int) (common.Hash, error
 
 func testHeaders_SystemContractsHaveNonZeroNonce(t *testing.T, headers []*types.Header, client *ethclient.Client) {
 	require := require.New(t)
-	ctxt := t.Context()
 	for i := range headers {
 		block := big.NewInt(int64(i))
 		for _, addr := range []common.Address{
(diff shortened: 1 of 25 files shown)


tests/block_header_test.go
```diff
@@ -535,7 +534,7 @@ func getStateRoot(client *ethclient.Client, blockNumber int) (common.Hash, error
 
 func testHeaders_SystemContractsHaveNonZeroNonce(t *testing.T, headers []*types.Header, client *ethclient.Client) {
 	require := require.New(t)
-	ctxt := context.Background()
+	ctxt := t.Context()
```
simonlechner, line 537: `ctxt` is only used once and could be inlined.

tests/gas_price_suggestion_test.go
```diff
@@ -19,7 +18,7 @@ func TestGasPrice_SuggestedGasPricesApproximateActualBaseFees(t *testing.T) {
 
 	fees := []uint64{}
 	suggestions := []uint64{}
-	ctxt := context.Background()
+	ctxt := t.Context()
```
simonlechner, line 21: As it is inlined everywhere else I would like to see the same thing here, although it is used twice. Just a suggestion though.

simonlechner (approved):
Nice clean up, added two picky comments regarding unity.

simonlechner (approved):

PR #338, opened 2025-06-24
Add faulty validators to liveness test of the single proposer protocol

HerbertJordan:
This PR extends the existing liveness test for the single-proposer protocol by the option of faulty nodes.

After this PR, the following cases are covered:
- up to `N-1` silent validators, not making any proposals
- up to `N-1` faulty validators, making proposals for the wrong blocks
- up to `N-1` randomly silent or faulty validators

In each case, the test verifies that progress on the block chain can be retained as long as there is at least one honest participant.


facuMH (dismissed):
This PR modifies single proposer protocol testing to test the network progresses even if there are validators that do not propose anything, or send proposals for the wrong block.

facuMH (approved):

PR #341, opened 2025-06-25
Fix timing sensitive skipped transaction test

HerbertJordan:
This PR should fix a flaky test as described in https://github.com/0xsoniclabs/sonic-admin/issues/235.

The test implicitly assumed that all test-transactions end up in the same block. However, depending on the timing of the block formation, transactions may get distributed among multiple blocks.


simonlechner (dismissed):
Looks good, thanks for the fix.

simonlechner (approved):

PR #349, opened 2025-07-01
Update to geth dependency to version 1.16.0

simonlechner:
This PR updates the geth dependency to 1.16.0.
The following changes have been included
Removal of the `tx.EffectiveGasTipValue` function which used to ignore errors internally
Addition of the `deleteRange` function in lachesis-base
Addition of the `SyncKeyValue` function in the `KeyValueStore` interface


gossip/gasprice/reactive.go
```diff
@@ -261,8 +261,9 @@ func (gpo *Oracle) calcTxpoolStat() txpoolStat {
 	p := uint64(0)
 	for _, tx := range sorted {
 		for p < uint64(len(s.percentiles)) && gasCounter >= p*maxGasToIndex/uint64(len(s.percentiles)) {
-			s.percentiles[p] = tx.EffectiveGasTipValue(minGasPrice)
-			if s.percentiles[p].Sign() < 0 {
+			gasTip, err := tx.EffectiveGasTip(minGasPrice)
+			s.percentiles[p] = gasTip
```
HerbertJordan, line 265: This does not look right. If there is an error, `gasTip` should not be used, right?

gossip/blockproc/drivermodule/driver_txs.go
```diff
@@ -178,10 +178,11 @@ func (p *DriverTxListener) OnNewReceipt(tx *types.Transaction, r *types.Receipt,
 }
 
 func effectiveGasPrice(tx *types.Transaction, baseFee *big.Int) *big.Int {
-	if baseFee == nil {
+	gasTip, err := tx.EffectiveGasTip(baseFee)
```
HerbertJordan, line 181: Reading the documentation of this function, it is not a 100% replacement of the `EffectiveGasTipValue`. There are cases where the tip could be negative, in which case a negative tip is returned + an error. In the old version, the negative tip would be added to the baseFee, producing the effective gas price. In the new version, the `GasPrice` of the transaction would be used, which is different.

To be equivalent to the past, the error should indeed be ignored. Looking into the implementation of the `EffectiveGasTip` function, it also seems like the only possible error is that the tip might be negative.

HerbertJordan (changes requested):

ethapi/api.go
```diff
@@ -1737,7 +1737,9 @@ func (s *PublicTransactionPoolAPI) formatTxReceipt(header *evmcore.EvmHeader, tx
 	if header.BaseFee == nil {
 		fields["effectiveGasPrice"] = hexutil.Uint64(tx.GasPrice().Uint64())
 	} else {
-		gasPrice := new(big.Int).Add(header.BaseFee, tx.EffectiveGasTipValue(header.BaseFee))
+		gasTip, err := tx.EffectiveGasTip(header.BaseFee)
+		_ = err // EffectiveGasTip returns an error for negative values, this is no problem here
```
facuMH, line 1741:
```suggestion
		// EffectiveGasTip returns an error for negative values, this is no problem here
		gasTip, _ := tx.EffectiveGasTip(header.BaseFee)
```

HerbertJordan (approved):
Looks good, thanks!

PR #350, opened 2025-07-01
Add Integration tests heap profiling

facuMH:
As more functionality is added to Sonic, it needs to be thoroughly tested. One of the main testing approaches is integration tests, or End-to-End. Since lately many integration tests have been added, running them all takes longer and consumes more memory.

This PR introduces a profiler to the integration test net struct that (if activated through an environment variable) will generate a heap dump of the moment of peak consumption, probing every 100ms. 

To enable it run `SONIC_TEST_HEAP_PROFILE=on go test ./tests`. Acceptable values to enable this option are `1`, `on` or `true` (case sensitivity is ignored) and any other value will be ignored.
It only samples memory used by the integration test net and sub routines of it. 
The output will be stored in `sonic/build/profile/` and the name of the dump will be `mem_myTestName.pprof`, where `myTestName` is the name of the test that started the integration test net. 

to visualize the output one can use `go tool pprof -http "localhost:8000" build/profile/mem_myTestName.pprof`

This PR is part of https://github.com/0xsoniclabs/sonic-admin/issues/244


tests/integration_test_net.go
```diff
@@ -169,6 +177,66 @@ type integrationTestNode struct {
...
 
+////////////////////////////////////////////////////////////////////////////////
+// Memory profiler.
+// if enabled with the `-test.heap.profile` flag, it will write a heap dump to
+// the `../build/profile/` directory at the end of the test run.
+// The file will be named `mem_<test_name>.pprof` where `<test_name>` is the name
+// of the test that started the profiling.
+////////////////////////////////////////////////////////////////////////////////
+
+type netProfiler struct {
```
HerbertJordan, line 188: nit: since it is collecting memory data and not network statistics, should this be renamed to `memoryProfiler`? The fact that it is used by the network should be of lesser importance. Wdyt?

tests/integration_test_net.go
```diff
@@ -289,6 +357,11 @@ func startIntegrationTestNet(
 	// the network's session needs to know about the network itself
 	net.net = net
 
+	if testProfileEnabled != nil && *testProfileEnabled {
+		net.profiler = netProfiler{tb: t}
+		go profileRoutine(t.Context(), &net.profiler)
```
HerbertJordan, line 362: The member field is pointless. You could just pass the `t` reference to the `profileRoutine` function and track the `highSeen` as a local variable inside the `profileRoutine`.

tests/integration_test_net.go
```diff
@@ -169,6 +177,66 @@ type integrationTestNode struct {
...
+}
+
+// since `go test` already parses the flag, this variable needs to be declared
+// at the package level, so that it can be parsed when go test parses other flags.
+var testProfileEnabled = flag.Bool("test.heap.profile", false,
+	`Enable memory profiling for integration tests. 
+	If set, a memory profile will be written to the build/profile directory 
+	at the end of the test run.`)
+
+func profileRoutine(ctx context.Context, n *netProfiler) {
```
HerbertJordan, line 202: Identifiers should not include information about what it is identifying. You would not name something a "AddressStruct" or a "counterVariable". So this should also not be a "profileRoutine". Furthermore, functions should start with a verb. Please find a better name for this.

tests/integration_test_net.go
```diff
@@ -169,6 +177,66 @@ type integrationTestNode struct {
...
+			if memStats.HeapAlloc <= n.highestSeen {
+				continue
+			}
+			buffer.Reset()
+			n.highestSeen = memStats.HeapAlloc
+			require.NoError(n.tb, pprof.WriteHeapProfile(buffer))
+
+		case <-ctx.Done():
+			// write a file with the name of the test case that started the profiling
+			buildProfile := "../build/profile/"
```
HerbertJordan, line 224: I am not a particular fan of this relative addressing. It can mess up things if you run tests from a different working directory. It should either be an absolute directory, relative to the test-case file, or in the current working directory.

HerbertJordan (changes requested):

tests/integration_test_net.go
facuMH: struct removed, only `t` pass and `highSeen` is now just a local variable

tests/integration_test_net.go
facuMH: struct removed.

tests/integration_test_net.go
```diff
@@ -289,6 +350,13 @@ func startIntegrationTestNet(
 	// the network's session needs to know about the network itself
 	net.net = net
 
+	heapProfile := os.Getenv(heapProfileEnvVar)
+	if heapProfile == "1" ||
+		strings.EqualFold(heapProfile, "on") ||
+		strings.EqualFold(heapProfile, "true") {
+		go startHeapProfiler(t)
+	}
```
HerbertJordan, line 358: nit: may be moved into the function since any re-use will have to do the same.

HerbertJordan (approved):

tests/integration_test_net.go
facuMH: done

HerbertJordan (approved):

PR #351, opened 2025-07-02
Rename Opera mentions to Sonic

facuMH:
This PR the old client name "Opera" from comments and error messages, and replaces it by "Sonic".

Note: most mentions of Opera in the code are due to the package of the same name. Renaming that package is a task for the future. This PR aims only for comments and error messages.

Note to reviewers:  the command 
` grep -i " opera" -r . | grep -v html | grep -v ".git/" | grep -v -i operation | grep -v opera. | grep -v OperaE*`
can be used to look for mentions of opera that **do not** include `html` files, files in `.git` folder, and ignore mentiones of "operation", `opera.` to ignore mentions of the package `opera`  and structs named `OperaEvm` or `OperaEpoch`.

### DIFF
--- a/cmd/sonicd/app/launcher.go
+++ b/cmd/sonicd/app/launcher.go
@@ -236,7 +236,7 @@ func initApp() *cli.App {
 	return app
 }
 
-// opera is the main entry point into the system if no special subcommand is ran.
+// sonicd is the main entry point into the system if no special subcommand is ran.
 // It creates a default node based on the command line arguments and runs it in
 // blocking mode, waiting for it to be shut down.
 func lachesisMain(ctx *cli.Context) error {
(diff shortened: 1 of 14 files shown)


cmd/sonicd/app/launcher.go
```diff
@@ -236,7 +236,7 @@ func initApp() *cli.App {
 	return app
 }
 
-// opera is the main entry point into the system if no special subcommand is ran.
+// sonicd is the main entry point into the system if no special subcommand is ran.
```
HerbertJordan, line 239:
```suggestion
// lachesisMain is the main entry point into the system if no special sub-command is ran.
```

HerbertJordan (approved):

PR #352, opened 2025-07-02
Apply cache size flag when importing genesis

facuMH:
The command line flag `--statedb.cache` was not properly forwarded in the case of importing a json genesis file.
This PR fixes that issue by adding the cache size to `genesis.ImportParams` and `db.GossipDbParameters`


kjezek (approved):
thanks for investigating this

PR #353, opened 2025-07-02
archive cache size 

facuMH:
Thanks to the profiler tool introduced by #350 it was made obvious that the archive cache had a larger size than needed for the integration tests. In order to reduce the resources needed by the integration tests we limit the archive cache with the same value that the live cache is limited to.


kjezek (approved):
thx

PR #354, opened 2025-07-02
Add stop mechanism to new nodes

facuMH:
While investigating how to reduce memory consumption it was also noticed that some goroutines were not properly closed. This PR adds a stope mechanism to `startNode` in `lachesisMainInternal`.

This PR is part of https://github.com/0xsoniclabs/sonic-admin/issues/244


HerbertJordan (approved):

HerbertJordan:
Looks good, and I think the missing test coverage in the launcher can be ignored for this PR.

PR #355, opened 2025-07-03
Update carmen and specify db cache capacity

facuMH:
The profiling introduced by #350 showed that live and archive caches were unnecessarily big in integration tests.
Since [this](https://github.com/0xsoniclabs/carmen/pull/36) change of carmen users of carmen can specify the capacity when calling the mpt import methods.
So this PR updates carmen dependency to include this change and specifies the minimum recommended for import, while keeping the default value (`0`) for the heal import.

As a custom [test run ](https://scala.fantom.network/job/Carmen/job/Validate-State-Toolchain-Archive/373/) showed that this configuration is stable for big databases.

### DIFF
--- a/gossip/evmstore/statedb_import.go
+++ b/gossip/evmstore/statedb_import.go
@@ -52,7 +52,7 @@ func (s *Store) ImportLiveWorldState(liveReader io.Reader) error {
 	if err := os.MkdirAll(liveDir, 0700); err != nil {
 		return fmt.Errorf("failed to create carmen dir during FWS import; %v", err)
 	}
-	if err := mptio.ImportLiveDb(mptio.NewLog(), liveDir, liveReader); err != nil {
+	if err := mptio.ImportLiveDb(mptio.NewLog(), liveDir, liveReader, 2_000); err != nil {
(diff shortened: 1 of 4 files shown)


gossip/evmstore/statedb_import.go
```diff
@@ -52,7 +52,7 @@ func (s *Store) ImportLiveWorldState(liveReader io.Reader) error {
 	if err := os.MkdirAll(liveDir, 0700); err != nil {
 		return fmt.Errorf("failed to create carmen dir during FWS import; %v", err)
 	}
-	if err := mptio.ImportLiveDb(mptio.NewLog(), liveDir, liveReader); err != nil {
```
HerbertJordan, line 55: I think the decision on how much cache is needed for an import should be left to Carmen in the productive case. So here, the default should be used.

I can see a use case in tests, where you would like to use less. But as I understand, this function is the productive version that is also the one utilized by end-users.

HerbertJordan (changes requested):
I do not think that the user of the Carmen (Sonic) should be responsible for setting meaningful cache sizes. This should be he job of Carmen.

facuMH:
This changes will be dropped.

PR #359, opened 2025-07-03
Add clean up for new node service protocols

facuMH:
This PR adds the `close` of each P2P protocol to the list of cleanups each node has to keep track of, and execute when closing. 

This PR contributes to https://github.com/0xsoniclabs/sonic-admin/issues/244


gossip/service.go
```diff
@@ -479,7 +479,7 @@ func (s *Service) RegisterEmitter(em *emitter.Emitter) {
 }
 
 // MakeProtocols constructs the P2P protocol definitions for `opera`.
-func MakeProtocols(svc *Service, backend *handler, disc enode.Iterator) []p2p.Protocol {
+func MakeProtocols(svc *Service, backend *handler, disc enode.Iterator) ([]p2p.Protocol, func()) {
```
HerbertJordan, line 482: Should we name the second return parameter to document that this is a cleanup? Or is there a type definition somewhere that defines `type Cleanup func()` we could use here? Or should it be added?

HerbertJordan (approved):

gossip/service.go
facuMH: I like this idea, if it does not exist already I will add it since it will be needed for clean up of other dangling routines

HerbertJordan (approved):
Nice!

PR #360, opened 2025-07-03
Add missing time-out counter in test

HerbertJordan:
This change fixes an issue in the test-setup of the single-proposer protocol tests which causes a nil-access error in case scheduling is delayed beyond the set time-out limit.

This issue was observed during a data-race detection run ([#29](https://scala.fantom.network/job/Sonic/job/Unit-test-race-detection/29/console)) and was reported as issue [#245](https://github.com/0xsoniclabs/sonic-admin/issues/245).


facuMH (approved):
This PR introduces a a non-functional structure just for tests.

PR #361, opened 2025-07-03
Add exit to timeout loop 

facuMH:
This PR adds weak pointer reference to `PublicFilterAPI`s timeout loop so that the API can be destroyed, the go routine of the timeout loop can also exit gracefully.

### DIFF
--- a/gossip/filters/api.go
+++ b/gossip/filters/api.go
@@ -24,6 +24,7 @@ import (
 	"math/big"
 	"sync"
 	"time"
+	"weak"
 
 	"github.com/0xsoniclabs/sonic/evmcore"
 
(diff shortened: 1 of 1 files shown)


gossip/filters/api.go
```diff
@@ -83,19 +84,25 @@ func NewPublicFilterAPI(backend Backend, cfg Config) *PublicFilterAPI {
 		events:  NewEventSystem(backend),
 		filters: make(map[rpc.ID]*filter),
 	}
-	go api.timeoutLoop(5 * time.Minute)
+	go timeoutLoop(weak.Make(api), 5*time.Minute)
```
HerbertJordan, line 87: For the PR comment: keeping only a weak reference in the timeout-Loop is the actual reason why the API instance (and the Backend including the Carmen DB) can actually be finalized. In the old version, the timeout look kept those in use.

PR #362, opened 2025-07-04
Update generated files

facuMH:
This PR updates all generated files. 
Only modified files are mocks, pb files and generated contracts for integration tests 
to reproduce run:
```
go generate ./...
make license-add
```
unit and integration tests run successfully with these changes.


HerbertJordan (approved):
Thanks for running this clean-up pass. 

The description should be extended to mention that also the smart contracts got refreshed by this change.

PR #363, opened 2025-07-04
Add generated-check target

facuMH:
This PR adds a Makefile target `make generated-check` which checks that all generated files (mocks, protobuf and contracts) are up to date, and fails if not. 
This new target is also used as a github workflows step

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -37,5 +37,8 @@ jobs:
       - name: License Check
         run: make license-check
 
+      - name: Generated Check
+        run: make generated-check
+
       - name: Build
         run: make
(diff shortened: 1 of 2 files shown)


HerbertJordan (commented):
Nice feature, but I am worried about version conflicts in the generator scripts.

To enable this, we need to somehow fix the versions of gomock, protoc, and solc such that everybody is working on the same versions.

HerbertJordan (commented):
@LuisPH3 can you have a look into this?

LuisPH3 (commented):
I have to raise a question here:
If we are committing the generated files, and requiring them to be generated with the same version of the generated tool. Why don't we just stop committing them and generate them instead in build time.

I foresee a problem here when migrating between linux distros, where the exact versioning may not be available. This PR will make it impossible to commit from a distribution which is not ubuntu 24.

PR #366, opened 2025-07-04
Merge tests regarding blobs

facuMH:
In an effort to reduce resources needed by the extensive integration tests in our project, tests with similar/same domain can be ran into the same test net. 
As is the case for the tests for blob transaction acceptance and blob base fee, both regarding the domain of the blobs.


LuisPH3 (approved):

LuisPH3:
The proposed change place blob tests within the same file and execute tests within the same test net, speeding up testing.

PR #367, opened 2025-07-04
Add sync point to wait for idle reorg thread in TxPool Test

HerbertJordan:
This change fixes a timing-sensitive flaky unit test for the transaction pool.

The TxPool test `` asserts in line [2423](https://github.com/0xsoniclabs/sonic/blob/11240d3f00bb95a7963e3a35a1557351d516023c/evmcore/tx_pool_test.go#L2423) that the preceding [`AddRemote`](https://github.com/0xsoniclabs/sonic/blob/11240d3f00bb95a7963e3a35a1557351d516023c/evmcore/tx_pool_test.go#L2413) operation is triggering the emission of an event signaling the presence of a new transaction. However, this only happens if when calling `AddRemote` the TxPool re-org loop running in a background thread is idle. If the re-org loop is still running, the asserting is not satisfied.

This timing issue is occurring more frequent in runs searching for data races and can be provoked in regular runs by adding a `time.Sleep(50 * time.Milliseconds)` at the end of the re-org loop [here](https://github.com/0xsoniclabs/sonic/blob/11240d3f00bb95a7963e3a35a1557351d516023c/evmcore/tx_pool.go#L1319).

This fix adds support for explicitly waiting for the reorg-routine to be idle before performing the TxPool modifications intended to emit new events. With this change, the operation sequence of a non-delayed test execution is enforced, eliminating the flakiness of this test even if an extra sleep is present.


evmcore/tx_pool.go
```diff
@@ -1161,6 +1167,15 @@ func (pool *TxPool) requestPromoteExecutables(set *accountSet) chan struct{} {
 	}
 }
 
+// waitForIdleReorgLoop_forTesting allows tests to wait for the reorg loop to
+// finish its current run. This is useful for tests that want to control the
+// timing of reorgs and promotions, ensuring that the pool is in a stable state
+// before proceeding with further assertions or actions.
+func (pool *TxPool) waitForIdleReorgLoop_forTesting() {
```
facuMH, line 1174: Nice comment. 
nit: considering this is a test only function, it should be in the corresponding testing file. Since it is in the same package it could take the txpool as a parameter and still refer to the channels.

facuMH (approved):
Nice fix !! thanks for figuring this one out. Just a small comment

facuMH (approved):
Thanks for the fix !!

LuisPH3 (approved):
The proposed changes add test-private sync infrastructure to fix a flaky test.

PR #374, opened 2025-07-08
Unify counter integration tests

facuMH:
In the efforts to reduce runtime of the integration tests, this PR uses the same integration test net instance for the two tests regarding the counter contract.
Theses cannot be executed in parallel since both change the counter value and expect to have each their own iterations contiguously.


LuisPH3 (approved):

LuisPH3 (approved):

HerbertJordan (approved):
Thanks for the cleanup!

PR #379, opened 2025-07-10
Only remove authorities when removing SetCodeTx

facuMH:
Before this change, our system would attempt to remove authorities every time any transaction was deleted. This led to unnecessary and misleading error logs whenever a non-SetCode transaction (which doesn't have authorities to remove) was processed.

This PR introduces a more precise approach: `removeAuthorities` will now only attempt to remove known authorities if the transaction being removed actually contains associated authorities. This cleans up our logs by preventing irrelevant error messages.

Additionally, to facilitate better testing and observability, a `log.Logger` has been introduced as a parameter to the `txLookup.removeAuthorities` and `txLookup.Remove` functions.


LuisPH3 (changes requested):
The fix is correct, but there are some details related to the testing that we have to get in sync with other testing happening in the project :
When introducing the dependency injection of the log, one of the goals is to test that the logs are emitted when expected. If you see other tests using injected logger, there is are test cases for the logging events. 

Additionally, introducing `pool.all.Remove(old.Hash(), log.Root())` in untested code is inconvenient, it lacks of coverage. (this is odd because these locations are collected, including the integration tests) 

Can we see if adding the logger interface to the lookup structure (instead of passing it to the remove function) facilitates getting a better coverage report?

facuMH:
Logger is added to the `TxPool` and `TxLookup` in https://github.com/0xsoniclabs/sonic/pull/382. This PR will be rebased onto that one so that it only introduces the tests and the actual fix.

PR #384, opened 2025-07-10
Unify and update gas price integration tests

facuMH:
In the efforts to reduce runtime of the integration tests, this PR reuses the same integration test net instance for both gas price suggestion integration tests. 
This PR also updates the test net configuration to `Allegro` fork and adds `SetCodeTx` to the tests.


tests/gas_price_suggestion_test.go
```diff
@@ -104,10 +139,16 @@ func TestGasPrice_UnderpricedTransactionsAreRejected(t *testing.T) {
 	require.NoError(send(factory.makeAccessListTransactionWithPrice(t, nonce+1, feeCap, 0)))
 	require.NoError(send(factory.makeDynamicFeeTransactionWithPrice(t, nonce+2, feeCap, 0)))
 	require.NoError(send(factory.makeBlobTransactionWithPrice(t, nonce+3, feeCap, 0)))
+	// SetCode transactions are restricted to a max of one in-flight transaction
+	// per address, so we need to use a different account.
+	require.NoError(send(makeSetCodeTransactionWithPrice(t, chainId, 0, feeCap, 0,
+		setCodeAccount.Address(), setCodeAccount.PrivateKey)))
```
LuisPH3, line 145: Although this is correct, the code is divergent from the patter used before. 
Instead of adding one new function to create transactions (I have no idea how many are there already), please add it to the tx factory. 

To get the code to work, it would be sufficient to initialize the factory with a new account per iteration instead of the `net.GetSessionSponsor()`
Feel free to modify the factory to store a complete account object instead of the private key, this will reduce the number of arguments of `makeSetCodeTransactionWithPrice` as well.

LuisPH3 (changes requested):
Although the changes achieve parallel execution, the new code diverges of the patterns utilized before, it would be better to keep the code consistent.

tests/gas_price_suggestion_test.go
facuMH: `makeSetCodeTransactionWithPrice` is not a member of `factory`, the test now simply instantiates 2 factories one for `SetCodeTx` and one for the others.

LuisPH3: please make it a member. I do not see why this type of transaction would need a different infrastructure.

LuisPH3 (changes requested):

PR #387, opened 2025-07-14
Reuse net and parallel in test net tests

facuMH:

In order to reduce the runtime of the integration tests, this PR reuses a single instance of the test net for various tests.
To avoid side effects or transaction nonce collitions, `EndowAccountsWithTokens`, `DeployContracts` and `InteractWithContract` tests take a `Session` instead of a network, so that they can Endow a new account without account or tx nonce colision, or overlap over deployed contracts.


HerbertJordan (dismissed):
Looks good, thanks for the improvement.

HerbertJordan (approved):

PR #389, opened 2025-07-15
Detect `log.Error` calls during integration tests

facuMH:
During work on improving integration test runtime, it was noticed that some tests logged errors. While these errors were not breaking errors, they should also not be ignored. Hence this PR introduces the integration test net `fakeLogger`, an implementation of the [go-ethereum logger](https://github.com/ethereum/go-ethereum/blob/master/log/logger.go#L106), which will fail any test during which a call to `log.Error`, `log.Critical` or `log.Write` with log level of `Error` or `Critical`, are registered.


facuMH:
For now this changes will be halted. To be re-analysed for future releases.

PR #391, opened 2025-07-15
Fix TxPool drop underpriced flaky test

facuMH:
This PR fixes a time sensitivity issue in a TxPool test. 

This PR fixes https://github.com/0xsoniclabs/sonic-admin/issues/260

The tests assumes that after each call to `addRemote` the reorder is executed. In this particular case of 2 transactions that kick out other transactions, the txpool needs to be reordered in between the two additions as well.


facuMH:
to reproduce the flaky tests, comment out the calls to `waitForIdleReorgLoop_forTesting` that are added in this PR.
And then the following tests will report it:
```
func TestRepeatFlaky(t *testing.T) {
	for i := 0; i < 1000; i++ {
		t.Run(fmt.Sprintf("TestTransactionPool_DropUnderpricedTransactionsWhenPoolIsFull_%d", i), func(t *testing.T) {
			TestTransactionPool_DropUnderpricedTransactionsWhenPoolIsFull(t)
		})
	}
}
```

PR #396, opened 2025-07-21
FIX: fix flaky test in integration test net tests

facuMH:
This PR fixes a flaky test, removing a session spawning from inside a parallel section to outside of it.


LuisPH3 (approved):
thanks

PR #406, opened 2025-07-24
Ignore large transaction test when checking for race conditions

facuMH:
This PR fixes https://github.com/0xsoniclabs/sonic-admin/issues/287, which describes a flaky test, due to the amount of parallelism in the test `testLargeTransactionLoadTest`, running it with `-race` took longer than anticipated and as a side effect the receipt for the transactions first submitted (last ones to be processed) got a long delay (between 1 and 6 minutes).

Now the tests is ignored when running with `-race`.

### DIFF
--- a/tests/large_transactions_test.go
+++ b/tests/large_transactions_test.go
@@ -97,9 +97,9 @@ func TestLargeTransactions_CanHandleLargeTransactions(t *testing.T) {
 
 func TestLargeTransactions_LargeTransactionLoadTest(t *testing.T) {
 
-	if isRaceTest() {
+	if isDataRaceDetectionEnabled() {
 		t.Skip(`Due to the concurrency requirements of this test, 
-		it becomes unstable when running with the race condition detector.`)
+		it becomes unstable when running with enabled data race detection.`)
 	}
 
 	hardForks := map[string]opera.Upgrades{
@@ -115,7 +115,6 @@ func TestLargeTransactions_LargeTransactionLoadTest(t *testing.T) {
 	for name, upgrades := range hardForks {
 		for mode, singleProposer := range modes {
 			t.Run(fmt.Sprintf("%s/%s", name, mode), func(t *testing.T) {
-				t.Parallel()
 				effectiveUpgrades := upgrades
 				effectiveUpgrades.SingleProposerBlockFormation = singleProposer
 				testLargeTransactionLoadTest(t, &effectiveUpgrades)
--- a/tests/large_transactions_test.go
+++ b/tests/large_transactions_test.go
@@ -96,6 +96,12 @@ func TestLargeTransactions_CanHandleLargeTransactions(t *testing.T) {
 }
 
 func TestLargeTransactions_LargeTransactionLoadTest(t *testing.T) {
+
+	if isRaceTest() {
+		t.Skip(`Due to the concurrency requirements of this test, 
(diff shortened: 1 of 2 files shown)


tests/large_transactions_test.go
```diff
@@ -194,7 +195,28 @@ func testLargeTransactionLoadTest(
 	// load peak.
 	slices.Reverse(transactions)
 
-	receipts, err := net.RunAll(transactions)
+	client, err := net.GetClient()
```
HerbertJordan, line 198: The parallel "RunAll" was originally specifically introduced for this load test to make sure that transactions are submitted fast enough into the network (through multiple client connections) such that sufficient transactions ended up in the same event stress-testing the message size limit. See #332 for details.


Making it sequential is probably eliminating the tests ability to provoke such situations reliably.

If this test makes problems in the race-detection case, we could consider skipping it if data race detection is enabled.

HerbertJordan (changes requested):

tests/large_transactions_test.go
LuisPH3: > The parallel "RunAll" was originally specifically introduced for this load test to make sure that transactions are submitted fast enough into the network (through multiple client connections) such that sufficient transactions ended up in the same event stress-testing the message size limit. See #332 for details.

Because transactions are being submitted in reverse order. They won't make it into an event until the first nonce is accepted into the pool. The parallel section is not functional but an optimization. 

I agree that this test may not be exploiting the issue under "-race". And skipping the test entirely may be a good option.

facuMH: Further experimentation proved that running the tests with the serialized sending did not trigger the original issue. Hence I think the best solution now is to keep the test as it was originally, and simply ignore it when checking for race conditions.

HerbertJordan: > Because transactions are being submitted in reverse order. They won't make it into an event until the first nonce is accepted into the pool. The parallel section is not functional but an optimization.

The reverse submission is not sufficient to trigger the desired effect reliably. Maybe a tuning of the number of transactions per account may improve the chances that a sequential submission is fast enough.

tests/large_transactions_test.go
```diff
@@ -109,6 +115,7 @@ func TestLargeTransactions_LargeTransactionLoadTest(t *testing.T) {
 	for name, upgrades := range hardForks {
 		for mode, singleProposer := range modes {
 			t.Run(fmt.Sprintf("%s/%s", name, mode), func(t *testing.T) {
+				t.Parallel()
```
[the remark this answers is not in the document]
facuMH: when sending transaction serially, it failed regardless o this parallel. I'm not sure if this is what you are asking

tests/large_transactions_test.go
```diff
@@ -96,6 +96,12 @@ func TestLargeTransactions_CanHandleLargeTransactions(t *testing.T) {
 }
 
 func TestLargeTransactions_LargeTransactionLoadTest(t *testing.T) {
+
+	if isDataRaceDetectionEnabled() {
+		t.Skip(`Due to the concurrency requirements of this test,
```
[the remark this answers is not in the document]
facuMH: it prints as a new line.

LuisPH3 (approved):

HerbertJordan (dismissed):

PR #407, opened 2025-07-24
Test net node reuse client

facuMH:
In order to reduce the amount of `ethclient.Client` that need to be initialized and released per test net, this PR introduces the `SharedClient` type, which is a wrapper on `ethclient.Client`, which instead releasing the client on `Close()` will return to a pool of which is a part of each network node.

### DIFF
--- a/tests/integration_test_net.go
+++ b/tests/integration_test_net.go
@@ -55,6 +55,7 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 	geth_crypto "github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/ethereum/go-ethereum/rpc"
 )
 
 // IntegrationTestNetSession a collection of methods to run tests against the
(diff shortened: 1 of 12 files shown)


tests/integration_test_net.go
```diff
@@ -1101,3 +1128,26 @@ func isDataRaceDetectionEnabled() bool {
 	}
 	return false
 }
+
+type ethClient = ethclient.Client
+
+type SharedClient struct {
```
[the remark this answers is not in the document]
LuisPH3: Additionally, missing docstring.

tests/integration_test_net.go
```diff
@@ -659,6 +687,7 @@ func (n *IntegrationTestNet) RestartWithExportImport() error {
 
 // GetHeaders returns the headers of all blocks on the network from block 0 to the latest block.
 func (n *IntegrationTestNet) GetHeaders() ([]*types.Header, error) {
+
```
LuisPH3, line 690: nit: remove newline

tests/integration_test_net.go
```diff
@@ -1101,3 +1128,26 @@ func isDataRaceDetectionEnabled() bool {
 	}
 	return false
 }
+
+type ethClient = ethclient.Client
```
LuisPH3, line 1132: nit: comment aliasing reason

LuisPH3 (changes requested):

LuisPH3 (approved):

PR #409, opened 2025-07-26
Update sample genesis JSON file for Allegro upgrade

jmpike:
Updates sample JSON genesis file for the Allegro upgrade.

- Maximal epoch duration is 1 hour as required by rule validation criteria.
- Maximal event gas, which implies the maximal transaction gas, is set to 30Mgas. This exceeds EIP-7825 limit of 16,777,216 gas (2^24), but large (gas) transactions were requested by our users. Additional 30kgas are reserved for the event formation.
- ShortGasPower and LongGasPower allocations are unified. The rule validation criteria now require the validator gas economy for both categories to be the same.
- The gas pre-allocation period cannot exceed 1 minute.
- Allegro upgrade is enabled.
- Single proposer block formation upgrade is disabled.

### DIFF
--- a/example-genesis.json
+++ b/example-genesis.json
@@ -1,7 +1,7 @@
 {
   "_": "This is a sample JSON genesis file. Use 'sonictool genesis json example-genesis.json' to generate a database from it.",
   "rules": {
-    "Name": "sample-demonet",
+    "Name": "sample-sonic-network",
     "NetworkID": 4003,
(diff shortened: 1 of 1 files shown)


LuisPH3 (approved):
This PR changes the configuration for the example-genesis file.

HerbertJordan (approved):
Thanks!

PR #410, opened 2025-07-28
RPC API. Add Blob and SetCode transactions to transactionArgs to types.Transaction conversion.

LuisPH3:
This PR adds support for Blob and SetCode txs in RPC routines using conversion to `*types.Transaction`. The most important one being `AccessList` computation.

Notice that the conversion may fail with ill-formed `transactionArgs`, and that the conversion function relies on setDefaults being called before. This behavior has not been changed to minimize code changes at this stage of the release cycle. 
Future changes in the RPC may remove the need for this function. 

contributes to https://github.com/0xsoniclabs/sonic-admin/issues/284

### DIFF
--- a/ethapi/transaction_args_test.go
+++ b/ethapi/transaction_args_test.go
@@ -439,6 +439,215 @@ func TestTransactionArgs_ToMessage_RejectsConversionWithIncoherentGasPricing(t *
 			msg, err := tc.args.ToMessage(0, nil, nil)
 			require.Nil(t, msg)
 			require.EqualError(t, err, "both gasPrice and (maxFeePerGas or maxPriorityFeePerGas) specified")
+
+		})
+	}
+}
+
+func TestTransactionArgs_ToTransaction(t *testing.T) {
+
+	tests := map[string]struct {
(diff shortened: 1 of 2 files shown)


ethapi/transaction_args_test.go
```diff
@@ -439,6 +439,215 @@ func TestTransactionArgs_ToMessage_RejectsConversionWithIncoherentGasPricing(t *
...
+	}
+}
+
+func TestTransactionArgs_ToTransaction(t *testing.T) {
+
+	tests := map[string]struct {
+		args     TransactionArgs
+		expected *types.Transaction
+	}{
+		"legacy transaction": {
```
jenikd, line 453: Nit: As other types starts with capital letter, renaming this toi `Legacy transaction`

jenikd (approved):

ethapi/transaction_args_test.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/410/commits/a6439460c611be9605fbfa1da4c040817a8d61c5

PR #417, opened 2025-08-01
shared parallel

facuMH:


LuisPH3:
To be studied and broken down into smaller PRs

PR #418, opened 2025-08-04
Fix source of flakiness. Avoid blocking pool reorg loop.

LuisPH3:
This PR refactors the handling of signals introduced to sync reorg-loop and tests. 
Previous implementation would block the reorg-triggering loop, preventing it from starting while waiting. 

Flakiness of the test `TestTransactionPoolUnderpricingDynamicFee` was observed when increasing parallelism of the integration tests. Although this test is in a different package (`evmcore` `txPool`)  the system load would affect the signaling required for the test to succeed.

### DIFF
--- a/evmcore/tx_pool.go
+++ b/evmcore/tx_pool.go
@@ -1182,12 +1182,13 @@ func (pool *TxPool) scheduleReorgLoop() {
 	defer pool.wg.Done()
 
 	var (
-		curDone       chan struct{} // non-nil while runReorg is active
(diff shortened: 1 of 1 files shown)


LuisPH3:
After studying the code, I come to the conclusion that this change is not needed:

Having a blocking signal wait and bypassing the early return when there is no concurrent reorg running are equivalent. As a side effect, no new reorg is started if it was not running already. 

The original race situation still needs to be identified (if any, because I cannot reproduce it locally)

PR #419, opened 2025-08-04
Remove test context

facuMH:
In order to improve resource reutilization `testContext` had been added to the blob integration test to facilitate instantiation and reutilitzation of test network and client. These mechanisms are now integrated into the integration test net basic functionality and are no longer needed to be explicitly handled in each integration test. 

This PR removes `testContext` from blob integration test. No impactful change in performance was observed after running 25 iterations before/after these changes.


LuisPH3 (approved):
This PR simplifies integration test by using common infrastructure.

PR #420, opened 2025-08-04
Avoid closing RPC client of pooled network clients

facuMH:
Since https://github.com/0xsoniclabs/sonic/pull/407 introduced the Pooled clients, no test should close the internal RPC client of the ethClients.

This PR removes those unwanted calls to `Close()` and keeps the name of variables holding the clients consistent across tests.


LuisPH3 (approved):
This PR removes close operations on pooled clients, which could lead to errors if a closed client is returned to the pool. 

code exploration shows no other close use of the raw client .

PR #422, opened 2025-08-04
Refactor gas price suggestion integration test

facuMH:
This PR implements a series of refactors in `gas_price_suggestion_tests.go`.
- Use `session` from a shared test network
- remove file shared network and structure ( this was an improvement when using an instance per file, but now that we use a shared network among many files, is no longer needed)


LuisPH3 (approved):
:rocket:

PR #425, opened 2025-08-05
Fix correct typing for block query.

LuisPH3:
This PR avoids the confusion of addressing blocks by event.Hash. Instead, uses common.Hash which at the time is binary compatible. 

This avoids unnecessary casting as well .


LuisPH3:
I am calling this PR off, 
coverage is not good enough to touch this code.

PR #429, opened 2025-08-06
Use sessions in set code integration test

facuMH:
This PR implements two main changes in the set code transaction integration tests.
- Use a session from a shared network.
- Make all calls to `t.Parallel` after spawning new sessions to prevent same nonce flakyness.


LuisPH3 (approved):
As a follow-up, we can remove the dispatcher top level test!

PR #430, opened 2025-08-06
Introduce test utils file

facuMH:
As a preparation for migrating the heavier integrations tests into their own packages, this PR introduces the utils file where a series  of helper functions are gathered.

Technical note: Since functions declared in `_test.go` files are only visible to other test files in the same package, we need to move 
this functions to a non test file and make them public as we want to use them in sub-packages.

This PR collaborates to https://github.com/0xsoniclabs/sonic-admin/issues/258


tests/test_utils.go
```diff
@@ -0,0 +1,353 @@
...
+		if len(rules.Name) > 0 {
+			return rules
+		}
+	}
+
+	t.Fatal("Failed to retrieve network rules after 10 attempts")
+	return opera.Rules{}
+}
+
+func GetEpochOfBlock(client *PooledEhtClient, blockNumber int) (int, error) {
```
LuisPH3, line 278: I noticed a pattern that can be beneficial for us.
We do not want to use these facilities outside of testing, this can still be possible even if added to a package when the first argument is (*testing.t) (or testing.TB to allow use in benchmarks) 

I think we should enforce that every signature in this file requires a t object.

LuisPH3 (changes requested):

PR #432, opened 2025-08-07
block header integration test package

facuMH:
This PR depends on https://github.com/0xsoniclabs/sonic/pull/430.
This PR moves to block header test into their own package so that the testing runtime can manage it in parallel to other packages. 

### DIFF
--- a/tests/block_header/block_header_test.go
+++ b/tests/block_header/block_header_test.go
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with Sonic. If not, see <http://www.gnu.org/licenses/>.
 
-package block_header_test
+package block_header
 
 import (
 	"cmp"
--- a/tests/single_proposer_protocol_test.go
+++ b/tests/single_proposer_protocol_test.go
@@ -22,7 +22,7 @@ import (
 	"testing"
 
 	"github.com/0xsoniclabs/sonic/opera"
-	block_header_test "github.com/0xsoniclabs/sonic/tests/block_header"
+	"github.com/0xsoniclabs/sonic/tests/block_header"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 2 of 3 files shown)


tests/block_header/block_header_test.go
```diff
@@ -14,27 +14,25 @@
...
 	"github.com/0xsoniclabs/carmen/go/database/mpt"
-	"github.com/0xsoniclabs/sonic/gossip/gasprice"
-	"github.com/0xsoniclabs/sonic/inter"
 	"github.com/0xsoniclabs/sonic/opera"
 	"github.com/0xsoniclabs/sonic/opera/contracts/driver"
 	"github.com/0xsoniclabs/sonic/opera/contracts/driverauth"
 	"github.com/0xsoniclabs/sonic/opera/contracts/evmwriter"
 	"github.com/0xsoniclabs/sonic/opera/contracts/netinit"
 	"github.com/0xsoniclabs/sonic/opera/contracts/sfc"
+	. "github.com/0xsoniclabs/sonic/tests"
```
LuisPH3: lets avoid "." imports

tests/single_proposer_protocol_test.go
```diff
@@ -22,6 +22,7 @@ import (
 	"testing"
 
 	"github.com/0xsoniclabs/sonic/opera"
+	block_header_test "github.com/0xsoniclabs/sonic/tests/block_header"
```
LuisPH3, line 25: I think that this alias is a left over.

tests/block_header/block_header_test.go
```diff
@@ -14,27 +14,25 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with Sonic. If not, see <http://www.gnu.org/licenses/>.
 
-package tests
+package block_header_test
```
LuisPH3, line 17: Packages with _test suffix are a special namespace which creates a second testing package inside the same folder as some code. 
In this case, there is no code. is all tests, please use the same package name for all files (without _testing)

LuisPH3 (changes requested):

LuisPH3 (approved):
Ready!

PR #433, opened 2025-08-07
Large transactions integration test package

facuMH:
This PR depends on https://github.com/0xsoniclabs/sonic/pull/430.
This PR moves to large transactions test into their own package so that the testing runtime can manage it in parallel to other packages.


tests/large_transactions/large_transactions_test.go
```diff
@@ -23,6 +23,7 @@ import (
 	"testing"
 
 	"github.com/0xsoniclabs/sonic/opera"
+	. "github.com/0xsoniclabs/sonic/tests"
```
LuisPH3, line 26: The importer of a package will use the name to refer to its contents, so exported names in the package can use that fact to avoid repetition. (Don't use the import . notation, which can simplify tests that must run outside the package they are testing, but should otherwise be avoided.) For instance, the buffered reader type in the bufio package is called Reader, not BufReader, because users see it as bufio.Reader, which is a clear, concise name. Moreover, because imported entities are always addressed with their package name, bufio.Reader does not conflict with io.Reader. Similarly, the function to make new instances of ring.Ring—which is the definition of a constructor in Go—would normally be called NewRing, but since Ring is the only type exported by the package, and since the package is called ring, it's called just New, which clients of the package see as ring.New. Use the package structure to help you choose good names.

https://go.dev/doc/effective_go

LuisPH3 (changes requested):

LuisPH3 (approved):

PR #438, opened 2025-08-11
Prevent flaky check

facuMH:
This PR depends on #440 

Previously the test:
- generated some block
- queried for the headers using `HeaderByNumber`
- then tried to get witness proof using `eth_getProof`
and because of an asynchronous update in the DB in rare cases it could happen that the last block reported as finished, is not yet in the archive. 
To fix this issue, this PR adds a retry mechanism to fetching the witness proofs.

This PR fixes https://github.com/0xsoniclabs/sonic-admin/issues/316

### DIFF
--- a/tests/block_header/block_header_test.go
+++ b/tests/block_header/block_header_test.go
@@ -400,6 +400,9 @@ func testHeaders_LastBlockOfEpochContainsSealingTransaction(t *testing.T, header
 
 func testHeaders_StateRootsMatchActualStateRoots(t *testing.T, headers []*types.Header, client *tests.PooledEhtClient) {
 	require := require.New(t)
+	// wait until last block is received.
+	waitForProofOf(t, client, len(headers)-1)
+
 	for i, header := range headers {
 		// The direct way to get the state root of a block would be to request the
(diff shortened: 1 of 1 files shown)


LuisPH3:
this looks good, but we could integrate #440

tests/block_header/block_header_test.go
```diff
@@ -405,38 +407,44 @@ func testHeaders_StateRootsMatchActualStateRoots(t *testing.T, headers []*types.
...
-		fmt.Sprintf("%v", common.Address{}),
-		[]string{},
-		fmt.Sprintf("0x%x", blockNumber),
-	)
-	if err != nil {
-		return common.Hash{}, fmt.Errorf("failed to get witness proof; %v", err)
-	}
+
+	err := tests.WaitFor(context.Background(), func(ctx context.Context) (bool, error) {
+		err := client.Client().Call(
```
kjezek, line 422: I think the cleanest way would be to ask RPC for the latest block, and wait until the expected block is reached

kjezek (approved):
looks good, thanks

PR #439, opened 2025-08-11
Add `t.Parallel` to large transaction tests

facuMH:
This PR adds missing calls to `t.Parallel` in large transaction integration tests.


facuMH:
This PR will be halted since it is already pushing the limits of our testing infrastructre

PR #442, opened 2025-08-13
Remove deprecated default command arguments

LuisPH3:
Default arguments included deprecated API namespaces, which log a warning. 
This PR removes these from the default command line arguments.


jenikd (approved):
Removing `ftm` nemaspace from default node config is ok as it is deprecated and should not be used in the future. User still can override this with the starting flags and get deprecated warning in the log.

PR #443, opened 2025-08-13
add missing receipt status check

facuMH:


LuisPH3 (changes requested):
the case relates to status, if err was not nil it would have failed in the previous line. please fix text.

PR #446, opened 2025-08-19
Fix corruption of emitter state files during healing.

LuisPH3:
When killing a running client with SIGABRT :
Attempting to restart will certainly not work because of the inconsistent state of the DB,  restart reports:
`failed to initialize the node: failed to make consensus engine: failed to open existing databases: non-initialized DB state`
After healing the db and starting the client again, boot sequence gets a little further but execution stops shortly after, error is now:
`Local database does not contain last emitted event - sync the node before enabling validation to avoid double sign.`

I want to ask you about this, I am afraid there may be some bug in Heal or some misalignment. I tried to reproduce it with the test-net data, but no luck there.
Steps to reproduce:
make db dir `mkdir /tmp/dbtest`
init db (fake) `./build/sonictool -datadir /tmp/dbtest/ genesis fake 1`
run client (with short heal checkpoints to wait less) `./build/sonicd  -datadir /tmp/dbtest/ --statedb.checkpointinterval 7 -fakenet 1/1`
wait until 7 blocks are created (checkpoint)
from a second terminal kill the process: `kill -6 <pid>`
first attempt to restart will complain no-db error, use same command as before to start client
heal the db: `./build/sonictool -datadir /tmp/dbtest/ heal`
second attempt to restart shows the second error message.

The proposed solution removes these "last-X" files storing the last received event from each other validator. 
The assumption is that the healing process resets the system to an epoch boundary and genesis events need to be received to sync up to the current state.

### DIFF
--- a/cmd/sonictool/app/heal.go
+++ b/cmd/sonictool/app/heal.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"strings"
 	"sync"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


LuisPH3:
as discussed with @Tjofil, this change will most certainly lead to double signing.

PR #462, opened 2025-09-02
Fix expected error on wait loop

facuMH:
This PR fixes the expected error to keep waiting until a block is present in the archive. 
Fixes [#316](https://github.com/0xsoniclabs/sonic-admin/issues/316)


LuisPH3 (approved):

LuisPH3 (approved):
good catch

PR #463, opened 2025-09-02
Wait for processed transaction to leave tx pool

facuMH:
This PR fixes https://github.com/0xsoniclabs/sonic-admin/issues/340

This PR adds a wait for the transaction with nonce 1 to have left the pool, so that the new transaction will not be miss interpreted as a replacement.


LuisPH3 (approved):
:raccoon:

PR #466, opened 2025-09-04
Improve test support when using Apple OS

LuisPH3:
The following error was reported while attempting to test some other feature:
"    /Users/luis/code/sonic/cmd/sonicd/app/test_cmd.go:246: (stderr) failed to start the node: error starting protocol stack: listen Unix /var/folders/p9/b72p6b1n49x34qmdcgp8g1340000gn/T/TestUnlockFlagAmbiguousWrongPassword68034954/001/sonic.ipc:"

Although Mac supports unix-sockets, setup seems to be different. The tests edited are not dependent on unix-socket IPC and fixing it should be its own domain. With appropriate IPC testing.


facuMH (approved):
This PR adds a command line flag to some tests to make them AppleOs compatible.

PR #467, opened 2025-09-04
Simplify block override integration test

jenikd:
This PR is changing way how to instantiate a contract for a later interaction. The way how it was done before this PR might be possible source of test flakiness.


facuMH (approved):
This PR changes how the block override contract is initiated in its integration test and how the function the contract offers.

PR #473, opened 2025-09-10
TxPool purges transactions with gas over rule limit

facuMH:
This PR depends on https://github.com/0xsoniclabs/sonic/pull/472 and resolves https://github.com/0xsoniclabs/sonic-admin/issues/364

This PR adds two tests to verify that when `MaxEventGas` is updated, then all existing transactions with gas the over limit set by the rules should be discarded.

In the reference implementation this is done with a newly introduced purge before resets in [`runReorg`](https://github.com/ethereum/go-ethereum/blob/master/core/txpool/legacypool/legacypool.go#L1255). Geth needs to handle it this way because they have no other way of enforcing the new gas limit. 

Since in Sonic the new max gas cap (introduced with EIP7825) shall be instead set by a network parameter ([here](https://github.com/0xsoniclabs/sonic/blob/main/evmcore/tx_pool.go#L1422) in the TxPool, whose value contribution comes from [`MaxEventGas`](https://github.com/0xsoniclabs/sonic/blob/main/gossip/evm_state_reader.go#L55)) and `TxPool.currentMaxGas` parameter is used to drop transactions during [demoteUnexecutables](https://github.com/0xsoniclabs/sonic/blob/main/evmcore/tx_pool.go#L1658) and [promoteExecutables](https://github.com/0xsoniclabs/sonic/blob/main/evmcore/tx_pool.go#L1462), transactions with a higher gas will be dropped during the resets following the update of `MaxEventGas` without the need to introduce code that would execute only once in all history.

### DIFF
--- a/evmcore/tx_pool.go
+++ b/evmcore/tx_pool.go
@@ -1270,8 +1270,6 @@ func (pool *TxPool) runReorg(done chan struct{}, reset *txpoolResetRequest, dirt
 	}
 	pool.mu.Lock()
 	if reset != nil {
-		// need to call discard before reset to check if rules will change
-		pool.discardTxsWithTooLargeGasOsaka(reset.oldHead, reset.newHead)
 		// Reset from the old head to the new, rescheduling any reorged transactions
 		pool.reset(reset.oldHead, reset.newHead)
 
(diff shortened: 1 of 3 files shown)


evmcore/tx_pool.go
```diff
@@ -1698,6 +1699,27 @@ func (pool *TxPool) demoteUnexecutables() {
...
+// discardTxsWithTooLargeGasOsaka removes transactions that exceed the maximum gas limit,
+// if the reset is already in Osaka.
+func (pool *TxPool) discardTxsWithTooLargeGasOsaka(reset *txpoolResetRequest) {
+	if reset.newHead != nil && reset.oldHead != nil {
+
+		// Discard the transactions with the gas limit higher than the cap.
+		if pool.chainconfig.IsOsaka(reset.newHead.Number, uint64(reset.newHead.Time)) && !pool.chainconfig.IsOsaka(reset.oldHead.Number, uint64(reset.oldHead.Time)) {
+			var hashes []common.Hash
+			pool.all.Range(func(hash common.Hash, tx *types.Transaction, local bool) bool {
+				if tx.Gas() > params.MaxTxGas {
```
HerbertJordan, line 1711: As discussed offline, the limit should be taken from the network rules.

evmcore/tx_pool_test.go
```diff
@@ -3134,6 +3135,49 @@ func TestSampleHashesManySenders(t *testing.T) {
 	}
 }
 
+func TestTxPool_ResetDropsTransactionsWithHighGas(t *testing.T) {
```
HerbertJordan, line 3138: Please also add a check that the filtering is not conducted if no change in the state of the Osaka enabled flag. Thus, most of the time, this cleanup should not be conducted.

Effectively, this cleanup will only be run once, right? The first time there is a re-org after the Osaka hard-fork?

PR #475, opened 2025-09-15
Update max tx gas for processor config

facuMH:
This PR is part of https://github.com/0xsoniclabs/sonic-admin/issues/362, implementation EIP-7825, Max Transaction Gas.

This PR implements the necessary changes for the sonic client to forward a max gas cap for transactions to the processor. 
To accomplish this objective a series of changes are needed:
- updates the go-eth-sonic dependency.
- correctly fill the `vm.config` to reflect network gas cap for transactions (or special case for internal transactions).
- update tests and mocks due to updating several interfaces.
- add unit tests for the new properties and an integration test.

Reviewing the changes by just looking at the files ordered alphabetically can be a bit chaotic, hence the reviewers are encouraged to review each commit individually, since they modify different sets of files.

### DIFF
--- a/evmcore/state_processor_test.go
+++ b/evmcore/state_processor_test.go
@@ -1332,22 +1332,22 @@ func TestRunTransaction_InternalTransactions_SkipsTransactionChecksTrue(t *testi
 	ctrl := gomock.NewController(t)
 	state := state.NewMockStateDB(ctrl)
 	any := gomock.Any()
-	state.EXPECT().SetTxContext(any, any)
+	state.EXPECT().SetTxContext(any, any).Times(2)
 	state.EXPECT().GetBalance(any).Return(uint256.NewInt(math.MaxInt64))
-	state.EXPECT().EndTransaction()
+	state.EXPECT().EndTransaction().Times(2)
 	state.EXPECT().SubBalance(any, any, any)
 	state.EXPECT().Prepare(any, any, any, any, any, any)
-	state.EXPECT().GetNonce(any).Return(uint64(0))
+	state.EXPECT().GetNonce(any).Return(uint64(0)).Times(2)
 	state.EXPECT().SetNonce(any, any, any)
-	state.EXPECT().GetCode(any).Return([]byte{}).Times(2)
+	state.EXPECT().GetCode(any).Return([]byte{}).AnyTimes()
 	state.EXPECT().Snapshot().Return(1)
 	state.EXPECT().Exist(any).Return(true)
 	state.EXPECT().GetRefund().Return(uint64(0)).Times(2)
 	state.EXPECT().AddBalance(any, any, any)
 	state.EXPECT().GetLogs(any, any)
 	state.EXPECT().TxIndex().Return(0)
-
 	rules := opera.FakeNetRules(opera.GetBrioUpgrades())
+	rules.Economy.Gas.MaxEventGas = maxTxGas
 	vmConfig := opera.GetVmConfig(rules)
 	updateHeights := []opera.UpgradeHeight{
 		{Upgrades: rules.Upgrades, Height: 0, Time: 0},
@@ -1387,7 +1387,6 @@ func TestRunTransaction_InternalTransactions_SkipsTransactionChecksTrue(t *testi
 
 	// -- end of setup --
 
-	rules.Economy.Gas.MaxEventGas = maxTxGas
 	unsignedTx := types.NewTx(&types.LegacyTx{
 		Nonce: 0, To: &common.Address{1}, Gas: maxTxGas * 2, GasPrice: big.NewInt(1),
 	})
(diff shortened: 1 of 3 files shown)


evmcore/state_processor.go
```diff
@@ -92,7 +92,7 @@ type ProcessedTransaction struct {
 // consistent.
 func (p *StateProcessor) Process(
 	block *EvmBlock, statedb state.StateDB, cfg vm.Config, gasLimit uint64,
-	usedGas *uint64, onNewLog func(*types.Log),
+	usedGas *uint64, onNewLog func(*types.Log), maxTxGas *uint64,
```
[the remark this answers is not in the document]
LuisPH3: please extend documentation, what is gasLimit and what is maxTxGas?

tests/tx_max_gas/tx_max_gas_test.go
```diff
@@ -88,4 +104,62 @@ func TestTxMaxGas(t *testing.T) {
 		require.NoError(t, err, "Transaction should succeed")
 		require.Equal(t, types.ReceiptStatusSuccessful, receipt.Status, "Transaction should be successful")
```
LuisPH3: I think this check belongs to another case: 
"transactions follow new rules limit after rules change"

tests/tx_max_gas/tx_max_gas_test.go
```diff
@@ -88,4 +104,62 @@ func TestTxMaxGas(t *testing.T) {
...
+		gappedTx := tests.CreateTransaction(t, net, &types.LegacyTx{Nonce: 1, Gas: 1_500_000}, account)
+		t.Logf("About to send gapped tx %v with gas %v", gappedTx.Hash(), gappedTx.Gas())
+		err := client.SendTransaction(t.Context(), gappedTx)
+		require.NoError(t, err, "Transaction should be accepted into the pool")
+
+		// update rules to lower max gas below the transaction's gas
+		rules.Economy.Gas.MaxEventGas = int64(1_100_000)
+		tests.UpdateNetworkRules(t, net, rules)
+		tests.AdvanceEpochAndWaitForBlocks(t, net)
+		tests.AdvanceEpochAndWaitForBlocks(t, net)
```
LuisPH3: supefluous, only one needed

tests/tx_max_gas/tx_max_gas_test.go
```diff
@@ -88,4 +104,62 @@ func TestTxMaxGas(t *testing.T) {
...
+
+		err = client.Client().Call(&rules, "eth_getRules", "latest")
+		require.NoError(t, err)
+		require.Equal(t, int64(1_100_000), rules.Economy.Gas.MaxEventGas, "MaxEventGas should be updated")
+
+		// send a transaction with the missing nonce and low gas
+		lowGasTx := tests.CreateTransaction(t, net, &types.LegacyTx{Nonce: 0, Gas: 500_000}, account)
+		t.Logf("About to send low gas tx %v with gas %v", lowGasTx.Hash(), lowGasTx.Gas())
+		err = client.SendTransaction(t.Context(), lowGasTx)
```
LuisPH3: Use `net.Run`, this tx does not need async handling.

evmcore/state_processor.go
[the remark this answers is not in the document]
LuisPH3: | gasLimit is actually the blockGasLimit. it should be renamed, perhaps in a follow-up.
I would prefer to see it now, but if you feel like it is too much, please attach issue link

| actually having it as a pointer was exactly for allowing it to be an optional. There are many tests where this is the case.
there is no use for this value being optional in the sonic client, tests follow needs, not otherwhise

tests/tx_max_gas/tx_max_gas_test.go
```diff
@@ -0,0 +1,180 @@
...
+		block, err := client.BlockByNumber(t.Context(), big.NewInt(int64(blockNumber)))
+		require.NoError(t, err)
+
+		if len(block.Transactions()) == 0 {
+			continue
+		}
+
+		// if the first transaction is an internal transaction, we found the epoch seal block
+		if internaltx.IsInternal(block.Transactions()[0]) {
+			sealEpoch = block.Transactions()[0]
```
HerbertJordan: nit: return here and avoid the `sealEpoch == nil` check in the loop condition ...

```suggestion
			return block.Transactions()[0]
```

evmcore/state_processor.go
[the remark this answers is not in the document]
facuMH: https://github.com/0xsoniclabs/sonic/pull/475/commits/072f8220a8ff36a43bad6d41da95c8bb58a292b3

evmcore/state_processor_test.go
```diff
@@ -1323,6 +1324,196 @@ func TestRunSponsoredTransaction_CoveredTransaction_ProcessesTwoTransactionsSucc
...
+}
+
+func TestRunTransaction_RegularTransaction(t *testing.T) {
+
+	tests := map[string]struct {
+		rules      opera.Rules
+		stateSetup func(state *state.MockStateDB)
+		validation func(t *testing.T, got ProcessedTransaction)
+	}{
+		"Brio/Rejected": {
```
LuisPH3: nit:

```suggestion
		"Brio/Skipped": {
```

LuisPH3 (dismissed):
👍

opera/vm_config_test.go
```diff
@@ -39,3 +39,36 @@ func TestGetVmConfig_SingleProposerModeDisablesExcessGasCharging(t *testing.T) {
...
+				MaxEventGas: want,
+			},
+		},
+	}
+
+	vmConfig := GetVmConfig(rules)
+
+	require.NotNil(t, vmConfig.MaxTxGas)
+	require.Equal(t, want, *vmConfig.MaxTxGas)
+	require.Equal(t, &rules.Economy.Gas.MaxEventGas, vmConfig.MaxTxGas)
```
HerbertJordan: In the code you make a point of it that a copy is made of the value, and that a pointer to the copy is written into the `vmConfig`. Thus, this should be not equal, right?

Maybe `require.Equal` is ignoring the pointers and comparing the value it is pointing to?

tests/tx_max_gas/tx_max_gas_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+	var rules struct {
+		Economy struct {
+			Gas struct {
+				MaxEventGas int64
+			}
+		}
+	}
+	// As enforced by rules validation a change on MaxEventGas can not be lower than opera.UpperBoundForRuleChangeGasCosts()
+	// so we set it to 1_000_000 and see that internal transactions still work.
```
HerbertJordan:
```suggestion
	// so we set it to that limit and see that internal transactions still work.
```

PR #484, opened 2025-09-22
Extend unit tests for state processor

HerbertJordan:
This PR extends unit tests for the state processor to increase the sensitivity regarding the indexing of receipts and log messages.

In particular, it re-confirms that the current state processor is indeed not ignoring skipped transactions when indexing logs and receipts. This should be further investigated. For now, this test shall ensure that this property is preserved during upcoming refactor steps.


simonlechner (approved):
Good to have this tested before the refactoring.

facuMH (approved):

PR #485, opened 2025-09-22
Add test for off-setting of the tx-index in receipts and logs

HerbertJordan:
This PR introduces additional unit test coverage verifying the correct off-setting of transaction IDs in receipts and log messages in the EVM processor.

This PR is a preparation step for #483.


facuMH (dismissed):
This PR adds a tests verifying the correct indexing of executed transactions and their receipts.

LuisPH3 (approved):
The sharing of state over a captured variable in the mocks threw me off at first. it does not worth to write a manual fake object.
the 15 iterations counter was not trivial. it took me a moment to realized that the test runs 5 iteration (of 3 txs) in the same processor instance. 

It is all good now.

PR #489, opened 2025-09-24
Add development version of subsidies registry

HerbertJordan:
This PR adds a development version of the on-chain contract governing Gas-Subsidies. It is intended to be used until the production-ready contract is released.

The main components of this PR are:
- `subsidies_registry.sol`  ... the contract itself
- `subsidies_contract.bin` ... the binary version of the contract as it is to be deployed on the chain
- `registry.go` ... a set of constants and manipulation utilities for installing the contract in test environments and including it in genesis files
- `registry_test.go` ... tests for the manipulation utilities

This contract will serve as the Gas-Subsidies contract for the duration of  the development of this feature. As stated above, it will get replaced with the production ready version as soon as it is available.


gossip/blockproc/subsidies/registry/registry_test.go
```diff
@@ -0,0 +1,91 @@
...
+	fmt.Printf("Transaction data:\n")
+	fmt.Printf("  Nonce: %d\n", tx.Nonce())
+	fmt.Printf("  Gas Price: 0x%x = %d\n", tx.GasPrice().Uint64(), tx.GasPrice().Uint64())
+	fmt.Printf("  Gas Limit: 0x%x = %d\n", tx.Gas(), tx.Gas())
+	v, r, s := tx.RawSignatureValues()
+	fmt.Printf("  V: 0x%x\n", v)
+	fmt.Printf("  R: 0x%x\n", r)
+	fmt.Printf("  S: 0x%x\n", s)
+
+	//t.Fail() // uncomment to see the output when running "go test"
```
LuisPH3, line 90: dead code

gossip/blockproc/subsidies/registry/registry_test.go
```diff
@@ -0,0 +1,91 @@
...
+func TestGetCode_CodeIsNotEmpty(t *testing.T) {
+	require.NotEmpty(t, GetCode())
+}
+
+func TestContractAddressMatchesCreatorAddress(t *testing.T) {
+	require := require.New(t)
+	_, creator := GetDeploymentTransaction()
+	want := crypto.CreateAddress(creator, 0)
+	got := GetAddress()
+	require.Equal(want, got)
```
LuisPH3, line 38: I think that it is important to verify that the sender address can be retrieved from the signature. Since all parameters are hardcoded, this is the only property which is checked.

gossip/blockproc/subsidies/registry/subsidies_registry.sol
```diff
@@ -0,0 +1,178 @@
...
+        // call data (less than 4 bytes, not covering the function selector).
+        if (to == address(0) || callData.length < 4) {
+            return (false, bytes32(0));
+        }
+        bytes4 selector = bytes4(callData[:4]);
+        return (true, keccak256(abi.encodePacked("c", from, to, selector)));
+    }
+
+    // Approval sponsorships cover all ERC20 approve calls from a specific
+    // account to a specific token contract and spender with a non-zero
```
LuisPH3, line 140: what is a spender?

gossip/blockproc/subsidies/registry/registry_test.go
LuisPH3: right, it fits better in this context though.

LuisPH3 (approved):

PR #493, opened 2025-09-24
Fix in code hash check.

facuMH:
This PR fixes  `checkDelegationLimit` function, to consider both `types.EmptyHash` as well as the zero value for `common.Hash` as empty codes.


LuisPH3 (commented):
👍

LuisPH3 (approved):

PR #494, opened 2025-09-24
Update processor to only report number of skipped transaction

HerbertJordan:
This PR reduces the amount of information returned by the processor regarding skipped transactions to merely the number of skipped transactions.

Before this change, `EVMProcessor`s returned the list of indexes of issues transactions that got skipped. However, this information is effectively not used in the code. The two only uses are
- in the `c_block_callbacks.go` file, to report the number of skipped transactions
- in the `genesis.go` file, to list the indexes of the failed initialization issues in an error

Since with the upcoming introduction of transactions been introduced by the block processor the indexes of skipped transactions would even be harder to interpret correctly, this PR suggests to reduce the emitted information to merely the number of skipped transactions.

**Note:** as this PR reveals, there is not a single unit test that would verify the content of the skipped transaction list. This PR adds test coverage for the reported number of skipped transactions.

### DIFF
--- a/gossip/blockproc/evmmodule/evm.go
+++ b/gossip/blockproc/evmmodule/evm.go
@@ -174,12 +174,12 @@ func (p *OperaEVMProcessor) Execute(txs types.Transactions, gasLimit uint64) typ
 	return receipts
 }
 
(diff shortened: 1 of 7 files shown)


facuMH (approved):
This PR modifies the `Finalize` method of the processor to return the number of skipped transactions instead of their indexes and adds a test verifying that this result is correct.

simonlechner (approved):

PR #496, opened 2025-09-24
Remove tracking of indices of skipped transactions

HerbertJordan:
As #494 removed the need for reporting the indices of skipped transactions, the internal tracking of those within the `OperaEVMProcessor` can be simplified.

### DIFF
--- a/gossip/blockproc/evmmodule/evm.go
+++ b/gossip/blockproc/evmmodule/evm.go
@@ -174,12 +174,12 @@ func (p *OperaEVMProcessor) Execute(txs types.Transactions, gasLimit uint64) typ
 	return receipts
 }
 
-func (p *OperaEVMProcessor) Finalize() (evmBlock *evmcore.EvmBlock, skippedTxs []uint32, receipts types.Receipts) {
+func (p *OperaEVMProcessor) Finalize() (evmBlock *evmcore.EvmBlock, numSkipped int, receipts types.Receipts) {
 	evmBlock = p.evmBlockWith(
 		// Filter skipped transactions. Receipts are filtered already
 		filterSkippedTxs(p.incomingTxs, p.skippedTxs),
 	)
-	skippedTxs = p.skippedTxs
+	numSkipped = len(p.skippedTxs)
 	receipts = p.receipts
 
 	// Commit block
--- a/gossip/blockproc/evmmodule/evm.go
+++ b/gossip/blockproc/evmmodule/evm.go
@@ -92,10 +92,11 @@ type OperaEVMProcessor struct {
 
 	gasUsed uint64
 
-	incomingTxs types.Transactions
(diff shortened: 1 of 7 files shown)


gossip/blockproc/evmmodule/evm.go
```diff
@@ -92,10 +92,11 @@ type OperaEVMProcessor struct {
 
 	gasUsed uint64
 
-	incomingTxs types.Transactions
-	skippedTxs  []uint32
-	receipts    types.Receipts
-	prevRandao  common.Hash
+	numIncomingTxs int
+	numSkippedTxs  int
+	processedTxs   types.Transactions
```
LuisPH3, line 97: nit: skipped transactions are also processed. Rename to `executedTxs `

gossip/blockproc/evmmodule/evm.go
```diff
@@ -92,10 +92,11 @@ type OperaEVMProcessor struct {
...
+	receipts       types.Receipts
+	prevRandao     common.Hash
 
 	rules opera.Rules
```
LuisPH3, line 101: nit: repeated field (`net` )

LuisPH3 (approved):
This is the design decision that I would have liked to re-consider:
by changing the internal state from an exhaustive list of receipts to a list of receipts of only the executed transactions, we break the state into several members that need to be manually handled: a counter for incoming, a counter for skipped. 

By keeping a full list of txs and optional receipts, and building an exhaustive list, the finalize method can take the decision of converting this into a list of executed txs and number of skipped. this accounting does nto leak into any internal data structure and the algorithm is simplified.

If the goal is to return a list of executed transactions from evmProcessor.Process, the decision of skipping is carried into that function as well. and the counters need to be replicated yet again inside this function. 

By returning the exhaustive list of transactions and optional receipts there is one less thing to care about. 

I am willing to compromise here, but i still believe that this design hides manual counting which could be easily misinterpreted, my solution is to keep one single data structure.

gossip/blockproc/evmmodule/evm.go
```diff
@@ -140,46 +141,41 @@ func (p *OperaEVMProcessor) evmBlockWith(txs types.Transactions) *evmcore.EvmBlo
...
+	p.numIncomingTxs += len(txs)
+	p.numSkippedTxs += len(skipped)
+	for i, receipt := range receipts {
+		if receipt != nil && i < len(txs) {
 			p.receipts = append(p.receipts, receipt)
+			p.processedTxs = append(p.processedTxs, txs[i])
 		}
 	}
 
 	return receipts
```
LuisPH3, line 173: This variable really needs a new name now, because `p.receipts` accumulates a list without skipped receipts but this variable `receitps` returns the full list.

HerbertJordan:
Replaced by #501

PR #497, opened 2025-09-24
Factor out transaction processing from StateProcessor

HerbertJordan:
This PR moves the transaction processing loop of the `StateProcessor`s `Process` function into a stand-alone function that can be re-used by the iterative `TransactionProcessor`.

### DIFF
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -85,10 +85,8 @@ func (p *StateProcessor) Process(
 	block *EvmBlock, statedb state.StateDB, cfg vm.Config, gasLimit uint64,
 	usedGas *uint64, onNewLog func(*types.Log),
 ) []ProcessedTransaction {
-	processed := make([]ProcessedTransaction, len(block.Transactions))
 	var (
 		gp           = new(core.GasPool).AddGas(gasLimit)
-		receipt      *types.Receipt
 		header       = block.Header()
 		time         = uint64(block.Time.Unix())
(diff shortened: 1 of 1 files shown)


facuMH (approved):
This PR abstracts the conversion of a transaction into a message and its execution step into a stand alone function, so that the this loop can be reutilized when needed.

evmcore/state_processor.go
```diff
@@ -179,18 +202,11 @@ func (tp *TransactionProcessor) Run(i int, tx *types.Transaction) (
...
-	tp.stateDb.SetTxContext(tx.Hash(), i)
-	receipt, _, err = applyTransaction(
-		msg, tp.gp, tp.stateDb, tp.blockNumber, tx,
-		&tp.usedGas, tp.vmEnvironment, tp.onNewLog,
+	processed := runTransactions(
+		[]*types.Transaction{tx}, tp.signer, tp.header.BaseFee, tp.stateDb,
+		tp.gp, tp.blockNumber, &tp.usedGas, tp.vmEnvironment, tp.onNewLog, i,
 	)
-	return receipt, err != nil, err
+	return processed[0].Receipt, processed[0].Receipt == nil, nil
```
simonlechner, line 209: The error is no longer returned in here, since a non nil error and nil receipt are treated equally in the calling  `evmProcessor.run` function, this does not introduce a difference in the processing.

simonlechner (commented):
Considering the calling function, no difference in this refactor is introduced.

LuisPH3 (approved):

PR #498, opened 2025-09-25
Have EVM return a list of included transactions with receipts

HerbertJordan:
This PR updates the `EVMProcessor` interface to produce a list of included transactions, instead of a list of receipts.

This modification has the following effects:
- it allows the EVM processor to introduce additional transactions as a side-effect of processing other transactions
- it eliminates the need for the users of the EVM processor to manually match transactions with receipts using indices
- it eliminates the need for the users of the EVM processor to filter out skipped transactions based on nil-receipts

### DIFF
--- a/gossip/blockproc/evmmodule/evm.go
+++ b/gossip/blockproc/evmmodule/evm.go
@@ -138,7 +138,7 @@ func (p *OperaEVMProcessor) evmBlockWith(txs types.Transactions) *evmcore.EvmBlo
 	return evmcore.NewEvmBlock(h, txs)
 }
 
-func (p *OperaEVMProcessor) Execute(txs types.Transactions, gasLimit uint64) types.Receipts {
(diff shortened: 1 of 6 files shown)


gossip/blockproc/evmmodule/evm.go
```diff
@@ -171,7 +171,16 @@ func (p *OperaEVMProcessor) Execute(txs types.Transactions, gasLimit uint64) typ
 		}
 	}
 
-	return receipts
+	result := make([]blockproc.IncludedTransaction, 0, len(txs)-len(skipped))
+	for i, receipt := range receipts {
+		if receipt != nil && i < len(txs) {
```
LuisPH3, line 176: this is fishy. Why would there be an extra condition to not store a receipt?
If there is a receipt at this point in time, it needs to go into the block, the work is already done
This will happen when subsidies txs are added to the block. 

I think that we need to refactor `evmProcessor.Process` first

HerbertJordan:
Obsolete; replaced by #501 , #502 and #503.

PR #501, opened 2025-09-25
Remove reporting of indices of skipped transactions

HerbertJordan:
This PR removes the redundant reporting of the indices of skipped transactions from the `StateProcessor`'s `Process` function. 

The required information regarding skipped transactions is also encoded in the returned receipts and can be derived from those. 

This is a preparation step eliminating toward allowing the processor to introduce additional transactions. By eliminating redundant results, less causes for inconsistencies can arise.

### DIFF
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -51,8 +51,10 @@ func NewStateProcessor(config *params.ChainConfig, bc DummyChain) *StateProcesso
 }
 
 // Process processes the state changes according to the Ethereum rules by running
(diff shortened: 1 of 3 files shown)


LuisPH3 (approved):

facuMH (approved):
This PR changes the return value of the state processor process function so that it returns simply a list of receipts (and nil for the skipped transactions) which is purged of nils when the block is finalized.

simonlechner (approved):

PR #502, opened 2025-09-25
Make state processor report list of processed transactions

HerbertJordan:
This PR updates the `StateProcessor` 's `Process` function to report the list of transactions it attempted to process. 

While in the current version this list is equivalent to the list of transactions being passed to it, upcoming changes will require the function to introduce additional transactions. The introduction of the ability to report the list of processed transactions facilitates this requirement.

The list of processed transactions are also reported by the `EVMProcessor`'s `Execute` function to allow those transactions to be accounted for in the block-formation covered by `c_block_callbacks.go`.

### DIFF
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -50,6 +50,15 @@ func NewStateProcessor(config *params.ChainConfig, bc DummyChain) *StateProcesso
 	}
 }
 
+// ProcessedTransaction represents a transaction that was considered for
+// inclusion in a block by the state processor. It contains the transaction
+// itself and the receipt either confirming its execution, or nil if the
+// transaction was skipped.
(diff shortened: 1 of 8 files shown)


evmcore/state_processor.go
```diff
@@ -107,9 +117,9 @@ func (p *StateProcessor) Process(
 			log.Debug("Failed to apply transaction", "tx", tx.Hash().Hex(), "err", err)
 			continue // skip this transaction, but continue processing the rest of the block
 		}
-		receipts[i] = receipt
+		processed[i].Receipt = receipt
```
LuisPH3, line 120: This PR adds a side effect not explained.
By continuing in both error cases, skipped transactions are not included in the processed list. 

This renders later filtering superfluous.

PR #505, opened 2025-09-26
Refactor TxPool to use the sonic state interface

LuisPH3:
This PR changes the source of state information from the pool from a scoped interface to the canonical sonic state.

This PR has two compromises:
- New interface has to manually be faked to avoid rewriting the large amount of testing inherited from Geth. 
- The pool has access to the full blockchain, and it is now capable of writing more intrusive and complex operations against the state. 

This PR facilitates later integration of the sponsored txs validation.

### DIFF
--- a/evmcore/tx_pool_test.go
+++ b/evmcore/tx_pool_test.go
@@ -29,14 +29,21 @@ import (
 	"testing"
 	"time"
 
+	"github.com/0xsoniclabs/carmen/go/common/witness"
+	"github.com/0xsoniclabs/sonic/inter/state"
 	"github.com/0xsoniclabs/sonic/utils"
 	"github.com/ethereum/go-ethereum/common"
+
+	geth_state "github.com/ethereum/go-ethereum/core/state"
+	"github.com/ethereum/go-ethereum/core/stateless"
+	"github.com/ethereum/go-ethereum/core/tracing"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/core/vm"
 	"github.com/ethereum/go-ethereum/crypto"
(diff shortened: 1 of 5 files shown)


evmcore/tx_pool_test.go
```diff
@@ -113,18 +120,161 @@ func (t testTxPoolStateDb) GetCodeHash(addr common.Address) common.Hash {
...
+func (t testTxPoolStateDb) Witness() *stateless.Witness {
+	panic("not implemented")
+}
+func (t testTxPoolStateDb) AccessEvents() *geth_state.AccessEvents {
+	panic("not implemented")
+}
+func (t testTxPoolStateDb) Finalise(bool) {
+	panic("not implemented")
+}
+
```
HerbertJordan, line 277: You could get the same by defining the `testTxPoolStateDb` as
```
type testTxPoolStateDb struct {
	state.StateDB
	balances   map[common.Address]*uint256.Int
	nonces     map[common.Address]uint64
	codeHashes map[common.Address]common.Hash
}
```
which would only be a one-line change.

By embedding the `state.StateDB` into the struct, you "inherit" all the functions. By not initializing this embedding, it will remain `nil`, and you get a `nil` access panic instead of the `not implemented` message, which functionally should be equivalent.

Furthermore, by defining it like this, you could set up a mock in the `newTestTxPoolStateDb` function if in the future some of those function accesses should be supported.

Also, if the `StateDB` interface changes in the future, and we such changes with every geth update, you will automatically inherit those changes.

HerbertJordan (approved):
Looks good.

The boilerplate code for testing could be minimized.

facuMH (approved):

PR #508, opened 2025-09-29
Update incremental processor to return processed transactions

HerbertJordan:
This PR updates the incremental version of the transaction processor to produce a list of processed transactions.

Before this change, the incremental processor only reported the result of the given transaction. However, with the introduction of sponsored transactions, the execution of a single transaction may result in the inclusion of multiple transactions in a block. Situations like this need to be supported by the incremental transaction processor.

To that end, this PR replaces the return type of the incremental processor's `Run` function with a list of processed transactions. Furthermore, updates the integration in the emitter's scheduler accordingly.

This change is a required modification on the path towards supporting sponsored transactions.


facuMH (dismissed):
This PR changes the interfaces of the `TransactionProcessor.Run` method so that it now returns a list of transactions and their receipts. The changes on the code and testing seem sufficient. I only have a couple of minor comments.

simonlechner (dismissed):

facuMH (approved):

PR #510, opened 2025-09-29
Refactor data sources for tx validate

LuisPH3:
This PR refactors the tx validation routines to use existing interfaces to query state and chain data instead of pre-filled data structures. 

These changes are needed for the integration of the gas subsidies, the tooling of gas subsidies requires access to the `state.StateDB` interface, which is a redundant source of information.

Because of the low granularity of the existing unit tests, the change in the interface forced to rewrite all tests for tx_validate. The new tests are declarative, self-contained and reduce the need for in the house tooling by using mocks.


evmcore/tx_validation_test.go
```diff
@@ -18,151 +18,131 @@ package evmcore
...
+		&types.LegacyTx{
+			Value: big.NewInt(-1),
+		},
+		&types.AccessListTx{
+			Value: big.NewInt(-1),
+		},
+		&types.DynamicFeeTx{
+			Value: big.NewInt(-1),
+		},
+		&types.DynamicFeeTx{
```
facuMH, line 49: duplicated ?

evmcore/tx_validation.go
```diff
@@ -50,48 +56,44 @@ type NetworkRules struct {
 	eip4844 bool // Fork indicator whether we are using EIP-4844 type transactions.
 	eip7623 bool // Fork indicator whether we are using EIP-7623 floor gas validation.
 	eip7702 bool // Fork indicator whether we are using EIP-7702 set code transactions.
```
facuMH, line 58: maybe as a follow-up we can rename this ones to revisions.

evmcore/tx_validation_test.go
```diff
@@ -234,159 +218,295 @@ func TestValidateTxForNetwork_Blobs_RejectsTxWith(t *testing.T) {
...
+				eip1559: true,
+				eip4844: true,
+				eip7702: true,
+				eip7623: true,
+				osaka:   true,
+			}
+			err := ValidateTxForNetwork(types.NewTx(tx), rules, chain, signer)
 			require.ErrorIs(t, err, ErrGasLimitTooHigh)
+
+			// check that the error is not happening with osaka disabled
```
facuMH, line 363:
```suggestion
			// check that the error is not reported with osaka disabled
```
or 
```suggestion
			// check that the error does not happen with osaka disabled
```

evmcore/tx_validation_test.go
```diff
@@ -18,151 +18,131 @@ package evmcore
```
facuMH, line 18: In many tests `common.Address{}` is used as an example value for non-nil, but sometimes address 0 has a special meaning, could you use a different address? other tests use `common.Address{123}`  or `common.Address{42}`.

evmcore/tx_validation_test.go
```diff
@@ -18,151 +18,131 @@ package evmcore
...
 
-			setGasPriceOrFeeCap(t, tx, big.NewInt(1))
-			setEffectiveTip(t, tx, big.NewInt(2))
+	t.Run("gas fee lower than gas tip", func(t *testing.T) {
+		txs := []types.TxData{
+			&types.DynamicFeeTx{
+				GasFeeCap: big.NewInt(1),
+				GasTipCap: big.NewInt(2),
+			},
+			&types.DynamicFeeTx{
```
facuMH, line 113: duplicated.

evmcore/tx_validation_test.go
```diff
@@ -18,151 +18,131 @@ package evmcore
...
+		}
+	})
 
-			err := ValidateTxStatic(types.NewTx(tx))
+	t.Run("gas tip larger than 256 bits", func(t *testing.T) {
+		txs := []types.TxData{
+			&types.DynamicFeeTx{
+				GasTipCap: extremelyLargeValue,
+			},
+			&types.DynamicFeeTx{
```
facuMH, line 94: duplicated.

evmcore/tx_validation_test.go
```diff
@@ -18,151 +18,131 @@ package evmcore
...
+			&types.LegacyTx{
+				GasPrice: extremelyLargeValue,
+			},
+			&types.AccessListTx{
+				GasPrice: extremelyLargeValue,
+			},
+			&types.DynamicFeeTx{
+				GasFeeCap: extremelyLargeValue,
+			},
+			&types.DynamicFeeTx{
```
facuMH, line 76: duplicated.

evmcore/tx_validation_test.go
```diff
@@ -18,151 +18,131 @@ package evmcore
...
 // Static Validation
 
 func TestValidateTxStatic_Value_RejectsTxWithNegativeValue(t *testing.T) {
-	for name, tx := range getTxsOfAllTypes() {
-		t.Run(name, func(t *testing.T) {
-			if isBlobOrSetCode(tx) {
-				t.Skip("blob and setCode transactions cannot have negative value because they use uint256 Value")
-			}
-			setValueToNegative(t, tx)
+	txs := []types.TxData{
```
facuMH, line 39: would it make sense to add a comment in each tests as for why there are some types of transactions not tested for that property ??

evmcore/tx_validation_test.go
```diff
@@ -234,159 +218,295 @@ func TestValidateTxForNetwork_Blobs_RejectsTxWith(t *testing.T) {
...
+			Gas: gasLimit + 1,
+		},
+		&types.DynamicFeeTx{
+			To:  &common.Address{}, // not a contract creation
+			Gas: gasLimit + 1,
+		},
+		&types.BlobTx{
+			Gas: gasLimit + 1,
+		},
+		&types.BlobTx{
```
facuMH, line 336: duplicated

evmcore/tx_validation_test.go
```diff
@@ -612,393 +892,322 @@ func TestValidateTxForPool_AcceptsNonLocalTxWithBigTip(t *testing.T) {
...
+			signer.EXPECT().Sender(gomock.Any()).Return(common.Address{}, nil).AnyTimes()
+			signer.EXPECT().Equal(gomock.Any()).Return(false).AnyTimes()
+
+			chain := NewMockStateReader(ctrl)
+			chain.EXPECT().GetCurrentBaseFee().Return(big.NewInt(5)).AnyTimes()
+			chain.EXPECT().MaxGasLimit().Return(uint64(100_000)).AnyTimes()
+			state := state.NewMockStateDB(ctrl)
+
+			opts := poolOptions{
+				minTip: big.NewInt(20), // higher than tx tip
```
facuMH, line 1064: where is the tip defined ? defaulted to zero? then why not 1 ?

evmcore/tx_validation_test.go
```diff
@@ -612,393 +892,322 @@ func TestValidateTxForPool_AcceptsNonLocalTxWithBigTip(t *testing.T) {
...
-		Data:      []byte("some data"),
-		AuthList:  []types.SetCodeAuthorization{{}},
-	}), netRules.signer, key)
-	require.NoError(b, err)
-
-	b.ResetTimer()
-	for i := 0; i < b.N; i++ {
-		err = validateTx(tx, opts, blockState, netRules)
-		require.NoError(b, err)
+		return "unknown"
```
facuMH, line 1211: is this ever checked/used ?

facuMH (changes requested):
This PR implements a refactor in the interface of `validateTx`, which involves modifying some of the structures used to pass information from the txpool to the validate function.
The core changes and logic seem okay, but I have some comments about the testing.

facuMH (dismissed):
This PR refactors the Tx Validation tests. It maintains the coverage and testing of the properties.

simonlechner (dismissed):
Looks good, just comments regarding typos.

evmcore/tx_validation_test.go
LuisPH3: fixed

simonlechner (approved):

PR #512, opened 2025-09-29
Add fee overflow check to IsCovered utility

HerbertJordan:
Adds an overflow check for fees when requesting sponsorship coverage.


LuisPH3 (approved):
Although overflow is checked in tx pool tx validation, an equivalent check is needed in the block processor. This would normally be done by pre-checks, but that would be too late for calling this function.

HerbertJordan:
> Although overflow is checked in tx pool tx validation, an equivalent check is needed in the block processor. This would normally be done by pre-checks, but that would be too late for calling this function.

Also, the gas costs considered here are increased by the overhead that is not checked for in other stages.

Also: defensive coding.

PR #516, opened 2025-09-30
Fix Tx indices in receipts in case of introduced transactions

HerbertJordan:
Updates the `OperaEVMProcessor` to be able to deal with transactions being introduced by the state processor.

With the introduction of sponsored transactions, additional internal transactions may be added at various places during the processing of lists of transactions. These additional transactions should be accounted for when assigning the transaction indices to log messages and receipts.


LuisPH3 (approved):
The introduced changes test that the transaction indices are correctly accumulated when execute is called multiple times.

facuMH (dismissed):

simonlechner (dismissed):

LuisPH3 (approved):

PR #517, opened 2025-09-30
Deploy gas subsidies registry contract when upgrades enabled.

facuMH:
This PR adds the deployment the gas subsidies registry contract in the integration test net when gas subsidies feature is enabled. 

Contract should only be deployed when the upgrade flag is enabled. FakeGenesis ignores optional feature flags and can only handle base `Sonic` or `Allegro` upgrades, hence it is not added nor tested in this PR. To add support for FakeGenesis to optional feature flags is a work outside of the scope of this PR.

### DIFF
--- a/tests/gas_subsidies/gas_subsidies_test.go
+++ b/tests/gas_subsidies/gas_subsidies_test.go
@@ -132,7 +132,7 @@ func TestGasSubsidies_CallingRegistryBeforeDeploy_FailsTransaction(t *testing.T)
 			}
 			ok, _, err := registryInstance.AccountSponsorshipFundId(&opts, sponsee.Address())
 			require.Error(err)
-			require.False(ok, "registry should have a fund ID")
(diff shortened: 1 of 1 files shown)


tests/gas_subsidies/gas_subsidies_test.go
```diff
@@ -95,3 +97,26 @@ func TestGasSubsidies_CanBeEnabledAndDisabled(
...
 	}
 }
+
+func TestGasSubsidies_Enabled_DeploysRegistryContract(t *testing.T) {
+	require := require.New(t)
+
+	upgrades := opera.GetSonicUpgrades()
+	upgrades.GasSubsidies = true
+
+	net := tests.StartIntegrationTestNet(t, tests.IntegrationTestNetOptions{
```
LuisPH3, line 107: Missing one of the two genesis methods.

LuisPH3 (changes requested):

PR #523, opened 2025-10-01
Integrate getGasConfig call into Subsidies Registry contract

HerbertJordan:
This PR upgrades the integration of the Subsidies Registry contract to also fetch the gas-limit configuration from the contract.

This update enables the configuration of gas limits on-chain, eliminating the need for hard-forks if a change would be required.


LuisPH3 (dismissed):
The proposed changes remove hardcoded gas limits for utility contract calls used during subsidies checks.
adjacent code was updated.

facuMH (dismissed):

simonlechner (approved):

LuisPH3 (approved):

simonlechner (approved):

PR #525, opened 2025-10-02
Gas subsidies: add all tx types test

simonlechner:
This PR adds an integration test to ensure that gas subsidies are supported with all transaction types.

### DIFF
--- a/tests/gas_subsidies/all_tx_types_test.go
+++ b/tests/gas_subsidies/all_tx_types_test.go
@@ -34,9 +34,8 @@ func TestGasSubsidies_SupportAllTxTypes(t *testing.T) {
 			To:  &common.Address{0x42},
 		},
 		"AccessListTx": &types.AccessListTx{
-			Gas:        21000 + 2400,
-			To:         &common.Address{0x42},
-			AccessList: []types.AccessTuple{{}},
+			Gas: 21000,
+			To:  &common.Address{0x42},
 		},
 		"DynFeeTx": &types.DynamicFeeTx{
 			Gas: 21000,
@@ -72,7 +71,7 @@ func TestGasSubsidies_SupportAllTxTypes(t *testing.T) {
 			// The sponsorship donation needs to be high enough to cover the gas
 			// fees of the sponsored tx and the fees of the internal payment tx.
 			donation := big.NewInt(1e16)
-			Fund(t, &net.Session, sponsee.Address(), donation)
+			Fund(t, net, sponsee.Address(), donation)
 
 			signedTx := makeSponsorRequestTransaction(t, tx, net.GetChainId(), sponsee)
 			require.NoError(t, client.SendTransaction(t.Context(), signedTx))


tests/gas_subsidies/all_tx_types_test.go
```diff
@@ -0,0 +1,83 @@
...
+			client, err := net.GetClient()
+			require.NoError(t, err)
+			defer client.Close()
+
+			sponsee := tests.NewAccount()
+
+			// The sponsorship donation needs to be high enough to cover the gas
+			// fees of the sponsored tx and the fees of the internal payment tx.
+			donation := big.NewInt(1e16)
+			Fund(t, &net.Session, sponsee.Address(), donation)
```
facuMH, line 75:
```suggestion
			Fund(t, net, sponsee.Address(), donation)
```
`IntegrationTestNetSession` is an interface which both nets and sessions implement.

facuMH (dismissed):
This PR adds a tests which ensures that transactions of every kind can be subsidized.

LuisPH3 (approved):
The test adds a table test to check subsidies feature interaction with each tx type.

tests/gas_subsidies/all_tx_types_test.go
```diff
@@ -0,0 +1,83 @@
...
+func TestGasSubsidies_SupportAllTxTypes(t *testing.T) {
+	transactions := map[string]types.TxData{
+		"LegacyTx": &types.LegacyTx{
+			Gas: 21000,
+			To:  &common.Address{0x42},
+		},
+		"AccessListTx": &types.AccessListTx{
+			Gas:        21000 + 2400,
+			To:         &common.Address{0x42},
+			AccessList: []types.AccessTuple{{}},
```
LuisPH3, line 39: nit: superfluous, empty access list is valid (empty auth list is not )

LuisPH3 (changes requested):

tests/gas_subsidies/all_tx_types_test.go
```diff
@@ -0,0 +1,83 @@
...
+		},
+	}
+
+	for name, tx := range transactions {
+		t.Run(name, func(t *testing.T) {
+			// Enable allegro to support setCode transactions.
+			upgrades := opera.GetAllegroUpgrades()
+			upgrades.GasSubsidies = true
+
+			net := tests.StartIntegrationTestNet(t, tests.IntegrationTestNetOptions{
```
LuisPH3, line 62: please use a session, this test does not change epochs or restarts the net.
@facuMH  can you check this ?

facuMH: because these tests are not in the `tests` package, they cannot benefit of the shared network. In a different PR we could add a similar file to `tests/session_dispenser.go` to have a shared network only for `gas_subsidies` package. But that would be outside of the scope of this PR. 
As we have one test per file, each test needs to have their own network.

facuMH (approved):

LuisPH3 (approved):

PR #529, opened 2025-10-02
Gas Subsidies: verifying funds get deducted.

facuMH:
This PR adds a test which runs a subsidized transaction and verifies that the cost has been deducted from the fund.


LuisPH3 (changes requested):
I want to encourage you going back to table tests, this new imperative testing idiom makes it very difficult to differentiate properties tested from expectations.

Please add tests: (please use a table) 
- sponsored transaction calls contract
- sponsored transaction transfers money (gas is paid by subsidies, value by sender) 
- sponsored transaction does nothing
- sponsored transaction calls contract which reverts 
- sponsored transaction is last in block.  
- sponsored transaction is not last in block.

PR #530, opened 2025-10-02
Add block replay test

HerbertJordan:
This PR adds an integration test running a mix of sponsored transactions and epoch changes on an integration network, followed by a verification of all produced blocks.

In particular, the following properties are checked:
- the effects of sponsored transactions can be recreated by a block processor unaware of the sponsoring
- fee-deduction transactions can be processed like regular transactions
- the nonces of internal transactions are consecutive


LuisPH3 (approved):
The PR introduces an off-client reproduction of a block history, which includes sponsorships. 

generated headers, block time, and hash are then compared.

PR #534, opened 2025-10-02
Filter transactions with lost sponsorship in reorg

HerbertJordan:
This PR adds support for removing transactions that lost their sponsorship from the transaction pool.

- [ ] test with norma

### DIFF
--- a/evmcore/tx_list.go
+++ b/evmcore/tx_list.go
@@ -23,6 +23,7 @@ import (
 	"sort"
 	"time"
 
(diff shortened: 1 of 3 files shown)


LuisPH3 (approved):
The changes update demote and promote to be sponsorship aware. These functions are called in the reorg of the pool when state changes. 

Future changes:
This change may justify to add the construction of the createSubsidiesChecker to the reset method, so that it is constructed once when the state change.
The construction should contain the instantiation of the evm to avoid that overhead.

PR #550, opened 2025-10-08
Signers cleanup.

LuisPH3:
This PR cleans some issues related to usage of signer facilities to reduce system workload. 

The PR #544 pointed out suboptimal usage of the signature cache, and research of the produced traces lead to the following conclusions:
-  The global_cache implementation overlaps with transactions sender cache mechanism.
- The caching can be avoided by using `types.Sender` freestanding function. 
- This function will still recompute the sender if the used signer is different to the used one to compute the cached value
- Lachesis events produce new transaction instances which do benefit from caching the sender

This PR does the following changes:
- Remove the sender cache 
- Use `types.Sender` instead of `signer.Sender` 
- Use `lastSignerForChainId` as canonical function to construct a signer, this prevents recomputation 
- Replace new instances of transactions received in lachesis events with their equivalent from the tx_pool
Different commits have been included to facilitate review. 

**Profiles** 

- baseline (main), observe the highlighted calls to sender retrieval
Screenshot from 2025-10-08 12-10-00
- After removing cache and using homogeneous signer tool: 
Screenshot from 2025-10-08 11-43-50
- After replacing txs in events from equivalent instances from the pool, the verification of signatures in events is gone:
Screenshot from 2025-10-08 11-51-28

TODOS: 
- [x] test that homestead txs (chainid 0)  still work. 
- [x] larger profile


LuisPH3:
PASS: TestChainId/AcceptsLegacyTxSignedWithHomestead (0.65s)

PR #555, opened 2025-10-20
 Increase default EventsBufferLimit to avoid sync lags

LuisPH3:
This PR is based on #492, it includes documentation about the issue and sets a metric to observe the behavior of the modified configuration. 

Incomplete events are those whose parents aren't known by the node, therefore cannot be added to the dag yet. When an event cannot be processed after received, it is stored in a buffer, which maximum size is configured by the parameters modified in this PR. 
Too low values can lead to dropping of recent dag history, forcing the node to fetch the forgotten events from the network again. 

Larger buffer size mean a wider window of time to store recently gossiped events, and this creates resilience against events which propagation is slower. The extra overheads of computing a larger buffer of events have been studied, and it is acceptable, as it cannot be sustained.

### DIFF
--- a/gossip/config.go
+++ b/gossip/config.go
@@ -223,9 +223,21 @@ func DefaultConfig(scale cachescale.Func) Config {
 		StructLogLimit:  2000,
 	}
 	sessionCfg := cfg.Protocol.DagStreamLeecher.Session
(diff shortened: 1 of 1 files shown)


thaarok (approved):

LuisPH3:
image
overnight test.
Time < 16:00 belongs to a different build

PR #556, opened 2025-10-20
Fix issue preventing from running live state check on archive db 

LuisPH3:
This change replaces `carmen.NoArchive`  by `carmen.S5Archive`  when opening the carmen database for checks. 
This allows sonictool checking the live db of an archive instance. An operation which is much faster than checking the full archive.  This is used by test scripts [here](https://scala.fantom.network/job/Sonic/job/RPC-node-sync/).

Fixes: https://github.com/0xsoniclabs/sonic-admin/issues/403

### DIFF
--- a/cmd/sonictool/app/app_test.go
+++ b/cmd/sonictool/app/app_test.go
@@ -23,7 +23,6 @@ import (
 	"io"
 	"math/big"
 	"os"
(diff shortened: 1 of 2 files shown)


HerbertJordan (approved):
Nice simple fix, good test coverage 👍

PR #564, opened 2025-10-24
Update carmen dependency for import/export with config

facuMH:
This PR updates carmen dependency and makes use of the new `[Import/Export]ArchiveWithConfig` using the parameter `s.cfg.Cache.StateDbCapacity` for its value. 

The command line flag `--statedb.cache` is [stored](https://github.com/0xsoniclabs/sonic/blob/main/config/flags/flags.go#L377) in `StateDbCacheCapacityFlag`.  Then [saved](https://github.com/0xsoniclabs/sonic/blob/main/config/config.go#L419) as evm state db config.

A test can be done to verify the effect of this changes: 
- run ` SONIC_TEST_HEAP_PROFILE=1 go test ./tests/ -run TestAddressAccess -count 1` before this changes, and open the resulting pprof file with ` go tool pprof -http "localhost:8000" ./build/profile/mem_TestAddressAccess.pprof`.
- apply these changes
- re run first step with a different port and compare the two flame graphs. 

Results before:
image

Results After:
image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ module github.com/0xsoniclabs/sonic
 go 1.24.0
 
 require (
-	github.com/0xsoniclabs/carmen/go v0.0.0-20251024080435-4a0104b6f4a1
+	github.com/0xsoniclabs/carmen/go v0.0.0-20251024091706-e1c5a32d4310
 	github.com/0xsoniclabs/tosca v0.0.0-20251023131117-f542dea83f5b
 	github.com/Fantom-foundation/lachesis-base v0.0.0-20240116072301-a75735c4ef00
 	github.com/cespare/cp v1.1.1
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -19,7 +19,7 @@ module github.com/0xsoniclabs/sonic
 go 1.24.0
 
 require (
-	github.com/0xsoniclabs/carmen/go v0.0.0-20251014092906-75df60dc78cb
+	github.com/0xsoniclabs/carmen/go v0.0.0-20251024080435-4a0104b6f4a1
```
LuisPH3: please update to `e1c5a32d4310a1b8c0f5758dfd5a874d3980f6aa`
this got merged recently and includes the fix for the other memory issue.

LuisPH3 (approved):

PR #568, opened 2025-10-29
Integration testing, Add option to define validator stakes

LuisPH3:
This PR adds a mechanism to define arbitrary stakes to the validators initialized for the integration tests underlaying fake-net.

This change is useful for simulation of more representative stake distributions.


HerbertJordan (commented):
I think replacing `NumNodes` with `ValidatorStakes` reduces readability. It would be nicer if `ValidatorStakes` would be an optional extension to `NumNodes` (see comments).

facuMH (approved):
Looks good, just have an optional suggestion.

PR #569, opened 2025-10-30
Add logs block time stamp format integration test

facuMH:
This PR adds an integration test that verifies the type and formatting of the logs returned by `eth_getLogs`.

### DIFF
--- a/tests/rpc_get_logs_test.go
+++ b/tests/rpc_get_logs_test.go
@@ -17,6 +17,7 @@
 package tests
 
 import (
+	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


HerbertJordan (commented):
Looks good; added an idea to target the encoding of the block time more directly, not depending on the implementation of `hexutil.Uint64`, which may change in the future and also allow decimal values as input to be more tolerant, which would make the test ineffective.

Have a look, and maybe try it out. If this adds complexity, feel free to ignore.

HerbertJordan (approved):

PR #570, opened 2025-10-31
Fix issue breaking integration tests in CI. 

LuisPH3:
This PR fixes an issue introduced by https://github.com/0xsoniclabs/sonic/pull/559 which prevents CI from running jobs.

The issue is that with the new P2P topology, test code can start before the connections have been established. 
geth `p2p/server.go` contains an asynchronous object where connections are processed in order, `AddPeer` queues connection requests which will be processed later. Waiting under load is unpredictable. 
This patch waits for each node in the test net to have connections before progressing. This wait is moved before serving wait as it should be completed earlier by the initializing node.

### DIFF
--- a/tests/integration_test_net.go
+++ b/tests/integration_test_net.go
@@ -541,26 +541,51 @@ func (n *IntegrationTestNet) start() error {
 		}
 	}
 
-	// connect to blockchain network
-	client, err := n.GetClient()
-	if err != nil {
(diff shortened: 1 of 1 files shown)


facuMH (approved):
This PR modifies how peers are connected in an integration test net, waiting for each to acknowledge each connection before checking that each node is ready to serve.

HerbertJordan (approved):
The change makes sense, but I do not understand the problem it is fixing from the description.

PR #572, opened 2025-10-31
Make RPC estimate gas Osaka aware

facuMH:
This PR implements https://github.com/0xsoniclabs/sonic-admin/issues/365 inspired by https://github.com/ethereum/go-ethereum/commit/59405c40d394552f1915caedd5f43d971c7ba1fd.

The RPC `eth_estimateGas` call needs to be aware of the new transaction gas limit introduced by EIP-7825 for consistency.

These changes make it so, and add an integration for it.

### DIFF
--- a/ethapi/api_test.go
+++ b/ethapi/api_test.go
@@ -1616,7 +1616,7 @@ func TestIsOsaka_OverridesAreUsedForDeterminingWhetherOsakaIsEnabled(t *testing.
 	}
 }
 
-func TestCapMaxGas_IsMaxGasAware(t *testing.T) {
+func TestCapMaxGas_IsUpgradesAware(t *testing.T) {
 
 	txGas := uint64(2_000_000)
 	maxGas := uint64(500_000)
@@ -1629,7 +1629,7 @@ func TestCapMaxGas_IsMaxGasAware(t *testing.T) {
 			upgrades:       opera.GetSonicUpgrades(),
 			wantEstimation: hexutil.Uint64(txGas),
 		},
-		"transaction over max gas after brio returns error": {
+		"max gas enforced for brio": {
 			upgrades:       opera.GetBrioUpgrades(),
 			wantEstimation: hexutil.Uint64(maxGas),
 		},
@@ -1649,6 +1649,7 @@ func TestCapMaxGas_IsMaxGasAware(t *testing.T) {
 			rules := opera.Rules{
 				Upgrades: test.upgrades,
 				Economy: opera.EconomyRules{
+					// always set max event gas
 					Gas: opera.GasRules{MaxEventGas: maxGas},
 				},
 			}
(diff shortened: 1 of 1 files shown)


tests/gas_estimate_test.go
LuisPH3: Using an integration test for this feature is overkill, you do not need any system running to call a freestanding function.
please be mindful with resources.

ethapi/api.go
```diff
@@ -1274,6 +1292,46 @@ func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNr
 	return hexutil.Uint64(hi), nil
 }
 
+// getNumberAndTime returns the block number and time for the given block number or hash,
+// applying any overrides specified in blockOverrides.
+// if the number or hash is invalid or does not exist, an error is returned.
+func getNumberAndTime(ctx context.Context, b Backend, blockNrOrHash rpc.BlockNumberOrHash, overrides *BlockOverrides) (uint64, uint64, error) {
```
LuisPH3: I think that overrides does not belong to this function. This function should translate num/hash into num + timestamp. If some utility wants to override, it is out of scope for this function.

ethapi/api_test.go
```diff
@@ -1445,3 +1449,218 @@ func TestFeeHistory_BlockNotFound(t *testing.T) {
...
+				mockBackend.EXPECT().HeaderByHash(gomock.Any(), common.HexToHash("0xabcdef")).Return(expectedHeader, nil)
+			},
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
```
LuisPH3: superfluos, gomock does cleanup

LuisPH3 (dismissed):
👍

ethapi/api.go
```diff
@@ -1274,6 +1302,36 @@ func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNr
...
+// if the number or hash is invalid or does not exist, an error is returned.
+func getNumberAndTime(ctx context.Context, b Backend, blockNrOrHash rpc.BlockNumberOrHash) (uint64, uint64, error) {
+
+	var header *evmcore.EvmHeader
+	var err error
+
+	// get header
+	rpcBlockNumber, ok := blockNrOrHash.Number()
+	if ok {
+		header, err = b.HeaderByNumber(ctx, rpcBlockNumber)
```
HerbertJordan: Careful here: I think this function returns `nil` for the header and the error if no block is present with that number. This is probably unexpected, but it seems to be the convention in the RPC package.

You need to check the header for `nil` before accessing it, otherwise the code will panic.

LuisPH3 (approved):

HerbertJordan (approved):

jenikd (approved):

PR #576, opened 2025-11-04
Brio: Add eip-7934 support

simonlechner:
Ethereum's Osaka hardfork includes [EIP-7934](https://eips.ethereum.org/EIPS/eip-7934), limiting the size of a rlp encoded block. The limit is enforced both in the distributed proposer (currently no limit) and single block proposer (already limited by the maximum event size) modes.
The EIP proposes a maximum block size of 10 Mib, but geth reserves 2 Mib for beacon chain data. Although, sonic currently does not save any SCC data in blocks, the same size considerations are made for future development.

### DIFF
--- a/gossip/c_block_callbacks_test.go
+++ b/gossip/c_block_callbacks_test.go
@@ -1118,14 +1118,14 @@ func TestProcessUserTransactions_SponsoredTxSizeIsAccountedCorrectly(t *testing.
 			})
 
 			processedTx1 := []evmcore.ProcessedTransaction{{Transaction: tx1, Receipt: &types.Receipt{}}}
-			var sponsoredTransaction *types.Transaction
+			var feeChargingTransaction *types.Transaction
 			if test.gasPrice == 0 {
-				sponsoredTransaction = types.NewTx(&types.LegacyTx{
+				feeChargingTransaction = types.NewTx(&types.LegacyTx{
 					// Fill the tx to simulate the size of a fee charging tx
 					Data: make([]byte, subsidies.RlpEncodedFeeChargingTxSizeInBytes),
 				})
 				processedTx1 = append(processedTx1, evmcore.ProcessedTransaction{
-					Transaction: sponsoredTransaction,
+					Transaction: feeChargingTransaction,
 					Receipt:     &types.Receipt{},
 				})
 			}
@@ -1149,7 +1149,7 @@ func TestProcessUserTransactions_SponsoredTxSizeIsAccountedCorrectly(t *testing.
 			if test.isSponsoredTx {
 				// ensure the sponsored transaction is followed by the gas paying transaction
 				require.Equal(t, tx1, gotTxs[1])
-				require.Equal(t, sponsoredTransaction, gotTxs[2])
+				require.Equal(t, feeChargingTransaction, gotTxs[2])
 
 				require.NotContains(t, gotTxs, tx2)
 				require.Equal(t, 1, skippedCount)
--- a/tests/max_block_size/max_block_size_test.go
+++ b/tests/max_block_size/max_block_size_test.go
@@ -20,8 +20,10 @@ import (
 	"fmt"
 	"math/big"
 	"testing"
+	"time"
 
 	"github.com/0xsoniclabs/sonic/integration/makefakegenesis"
+	"github.com/0xsoniclabs/sonic/inter"
 	"github.com/0xsoniclabs/sonic/opera"
 	"github.com/0xsoniclabs/sonic/tests"
 	"github.com/ethereum/go-ethereum/common"
@@ -139,7 +141,7 @@ func increaseLimits(t *testing.T, net *tests.IntegrationTestNet) {
 	modified.Economy.ShortGasPower.AllocPerSec = 50_000_000_000
 	modified.Economy.ShortGasPower.MaxAllocPeriod = 50_000_000_000
 	modified.Economy.LongGasPower = modified.Economy.ShortGasPower
-	modified.Emitter.Interval = 1_00_000_000
+	modified.Emitter.Interval = inter.Timestamp(1 * time.Second)
 	tests.UpdateNetworkRules(t, net, modified)
 	net.AdvanceEpoch(t, 1)
 


evmcore/state_processor.go
```diff
@@ -109,11 +110,32 @@ func (p *StateProcessor) Process(
 		ProcessParentBlockHash(block.ParentHash, vmenv, statedb)
 	}
 
+	transactions := block.Transactions
```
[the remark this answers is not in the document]
HerbertJordan: Indeed, this location is not a good choice. Also, recall that this function is called 3 times -- twice for internal transactions and once for the actual payload transactions. Here, you will only see one of those calls at a time.

[runTransactions](https://github.com/0xsoniclabs/sonic/blob/7443ad56c563df99da0a96c068802ef21dd734bc/evmcore/state_processor.go#L190-L191) may be a better place, as suggested by @LuisPH3 , but you would still have to consider that this function is called 3 times - also accounting for internal transactions.

There is one more challenge to be considered: sponsored transactions. If a transaction gets sponsored, you also need to consider the introduced internal transactions for the sponsorship payments. Before running a transaction you need to make sure that the transaction fits in the remaining space + if it is a sponsored transaction, the extra internal fee-charging transaction can fit.

LuisPH3 (approved):
The new location for the fix still drops transactions, which do not produce tx-receipt pairs.
This is not too bad because they are dropped from the back of the list. 
I would consider cleaning this up though. maybe removing the tx-nil need, as originally intended, nothing for this pr.

evmcore/state_processor_test.go
```diff
@@ -298,6 +301,101 @@ func TestProcess_TracksParentBlockHashIfPragueIsEnabled(t *testing.T) {
...
+	for _, isBrio := range []bool{false, true} {
+		t.Run(fmt.Sprintf("isBrio=%v", isBrio), func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			runner := NewMock_transactionRunner(ctrl)
+
+			transactions := []*types.Transaction{
+				types.NewTx(&types.LegacyTx{Data: make([]byte, params.MaxBlockSize/2)}),
+				types.NewTx(&types.LegacyTx{Data: make([]byte, params.MaxBlockSize/2)}),
+			}
+
```
facuMH: could you add her e something like `require.greater(tx1.size+tx2.size, params.MaxBlockSize)`?

evmcore/state_processor.go
```diff
@@ -186,6 +190,42 @@ func runTransactions(
...
+	// Cheating validator transactions have a size of around 112 bytes.
+	// Epoch Seal internal transaction have a size of around 10808 bytes.
+	// Epoch Seal Validator internal transaction have a size of around 2744 bytes.
+	RlpEncodedInternalTransactionSizeInBytes = 16384
+)
+
+// filterTransactionsExceedingMaxBlockSize ensures that the size of the block does not exceed the
+// maximum allowed block size introduced by EIP-7934.
+func filterTransactionsExceedingMaxBlockSize(txs types.Transactions, upgrades opera.Upgrades) types.Transactions {
+	size := RlpEncodedMaxHeaderSizeInBytes + RlpEncodedInternalTransactionSizeInBytes
```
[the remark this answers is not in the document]
simonlechner: The same variable is used to save the sum of the header and all txs, so it is not just the `baseSize`

gossip/emitter/scheduler/scheduler_test.go
HerbertJordan: Seems like an unrelated change.

evmcore/state_processor.go
```diff
@@ -186,6 +190,42 @@ func runTransactions(
...
+
+	// RlpEncodedInternalTransactionSizeInBytes is an upper bound for the size of
+	// RLP-encoded internal transactions included in the block for block size
+	// calculations.
+	// Transactions during the setup of the chain are not considered,
+	// it is unlikely to have a lot of payload transactions during the setup phase.
+	// Cheating validator transactions have a size of around 112 bytes.
+	// Epoch Seal internal transaction have a size of around 10808 bytes.
+	// Epoch Seal Validator internal transaction have a size of around 2744 bytes.
+	RlpEncodedInternalTransactionSizeInBytes = uint64(16384)
```
HerbertJordan: From the documentation it is not clear where this number is coming from. It is not the sum of the provided values. I presume it is intended to be all cases, rounded up to 16 KiB, so maybe state this in the comment and define it like
```suggestion
	RlpEncodedInternalTransactionSizeInBytes = uint64(16 * 1024)  // 16 KiB
```

HerbertJordan (changes requested):

gossip/emitter/scheduler/scheduler.go
```diff
@@ -79,6 +79,7 @@ func (s *Scheduler) Schedule(
 	defer processor.release()
 
 	remainingGas := limits.Gas
+
```
facuMH: unnecessary ?

gossip/c_block_callbacks.go
```diff
@@ -589,6 +597,47 @@ func consensusCallbackBeginBlockFn(
...
+					remainingSize -= txSizeIncSubsidies(tx)
+				}
+			}
+		}
+	}
+}
+
+// txSizeIncSubsidies returns the size of the transaction,
+// including any overhead introduced by sponsorship requests.
+func txSizeIncSubsidies(tx *types.Transaction) uint64 {
```
facuMH: maybe this is just me, but `Inc` could be confused with `Increment` instead of `Include`. what about `txSizeWithSubsidies` ?

gossip/blockproc/subsidies/subsidies_test.go
```diff
@@ -570,6 +571,21 @@ func TestIsCovered_InvalidReturnFromEvm_ReturnsError(t *testing.T) {
 	require.ErrorContains(err, "failed to parse result of subsidies registry call")
 }
 
+func TestRlpEncodedFeeChargingTxSizeInBytes_IsUpperBound(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	nonceSource := NewMockNonceSource(ctrl)
+	nonceSource.EXPECT().GetNonce(common.Address{}).Return(uint64(math.MaxUint64))
+	id := FundId{}
```
LuisPH3: https://github.com/0xsoniclabs/sonic/blob/8b30297530ce445d6f4bfb9863e0b565509941c7/gossip/blockproc/subsidies/subsidies.go#L308

Although the function seems to copy 32 bytes into init data, this function seems to address a worst case with maximum usage of each field bytes, It would make sense to add 32 bytes of data for the fundId as well.

LuisPH3 (dismissed):

gossip/c_block_callbacks.go
```diff
@@ -589,6 +591,59 @@ func consensusCallbackBeginBlockFn(
...
+		neededSpace := txSizeIncludingSubsidies(tx)
+		if neededSpace < remainingSize && tx.Gas() < remainingGas {
+			for _, processed := range evmProcessor.Execute([]*types.Transaction{tx}, remainingGas) {
+				if processed.Receipt != nil { // < nil if skipped
+					blockBuilder.AddTransaction(
+						processed.Transaction,
+						processed.Receipt,
+					)
+					remainingGas = userTransactionGasLimit - processed.Receipt.CumulativeGasUsed
+					remainingSize -= neededSpace
```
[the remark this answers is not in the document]
simonlechner: Yes, you are correct, the calculation was wrong.
The evmprocessor is mocked in all unit tests, therefore it does not return two transactions in the case of a sponsored transaction by default. I added a unit test modeling the sponsored transaction behavior in the mock.

gossip/c_block_callbacks.go
```diff
@@ -589,6 +591,59 @@ func consensusCallbackBeginBlockFn(
...
+func processUserTransactions(evmProcessor blockproc.EVMProcessor, blockBuilder *inter.BlockBuilder, orderedTxs []*types.Transaction, userTransactionGasLimit uint64) {
+	remainingGas := userTransactionGasLimit
+	remainingSize := uint64(params.MaxBlockSize - rlpEncodedMaxHeaderSizeInBytes)
+	internalTxs := blockBuilder.GetTransactions()
+	for _, tx := range internalTxs {
+		remainingSize -= tx.Size()
+	}
+	for _, tx := range orderedTxs {
+		neededSpace := txSizeIncludingSubsidies(tx)
+		if neededSpace < remainingSize && tx.Gas() < remainingGas {
```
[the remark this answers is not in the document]
simonlechner: I tried several different options and discussed some with @LuisPH3, the current version is the simplest without adding more complexity to the `consensusCallbckBeginBlockFn` and without adding unexpected arguments to the `evmProcessor.finalize` function. Adding an `addSkipped` function to the evmProcessor [interface](https://github.com/0xsoniclabs/sonic/blob/046feadf974bc52e591eb22dafdcfa706e1471ef/gossip/blockproc/interface.go#L70) could also be an option, but is not very expected and requires more changes to the evmprocessor implementation.

gossip/c_block_callbacks_test.go
```diff
@@ -864,6 +869,273 @@ func TestIsPermissible_DetectsNonPermissibleTransactions(t *testing.T) {
...
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			size := txSizeIncludingSubsidies(test.tx)
+			require.Equal(t, test.expectedSize, size)
+		})
+	}
+}
+
+func TestProcessUserTransactions_SponsoredTxSizeIsAccountedCorrectly(t *testing.T) {
```
HerbertJordan: nit: please keep test targeting the same function grouped

gossip/c_block_callbacks_test.go
```diff
@@ -864,6 +869,273 @@ func TestIsPermissible_DetectsNonPermissibleTransactions(t *testing.T) {
...
+			}
+
+			evmProcessor.EXPECT().
+				Execute([]*types.Transaction{tx0}, gomock.Any()).
+				Return([]evmcore.ProcessedTransaction{{Transaction: tx0, Receipt: &types.Receipt{}}})
+			evmProcessor.EXPECT().
+				Execute([]*types.Transaction{tx1}, gomock.Any()).
+				Return([]evmcore.ProcessedTransaction{
+					{Transaction: tx1, Receipt: &types.Receipt{}},
+					processedTxs,
```
HerbertJordan: nit: to be authentic, this transaction should only be added it it is not `nil`.

HerbertJordan (dismissed):

tests/max_block_size/max_block_size_test.go
```diff
@@ -0,0 +1,147 @@
...
+		for upgradeName, upgrade := range upgrades {
+			t.Run(fmt.Sprintf("%s_%s", modeName, upgradeName), func(t *testing.T) {
+
+				upgrade.SingleProposerBlockFormation = singleProposer
+				net := tests.StartIntegrationTestNet(t, tests.IntegrationTestNetOptions{
+					Upgrades: &upgrade,
+					Accounts: genesisAccounts,
+					NumNodes: 3,
+				})
+				increaseLimits(t, net)
```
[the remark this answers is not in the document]
simonlechner: There is currently no simple way to do this in the integration tests. I can add this feature in a follow up, but I would like to get those changes in first.

gossip/c_block_callbacks_test.go
```diff
@@ -864,6 +869,268 @@ func TestIsPermissible_DetectsNonPermissibleTransactions(t *testing.T) {
...
+		return b
+	}
+
+	maxHash := common.Hash(setToMax(32))
+	maxAddress := common.Address(setToMax(20))
+	maxBloom := types.Bloom(setToMax(2048))
+	maxBlock := types.BlockNonce(setToMax(8))
+	maxUint64 := uint64(math.MaxUint64)
+
+	// Geth sanity check for extra defines an upper bound of 100 * 1024 bytes.
```
facuMH: this line is a bit confusing for me. `geth sanity check*s* for extra defines` or `geth sanity check defines an extra upper bound` ?

gossip/c_block_callbacks_test.go
```diff
@@ -864,6 +869,268 @@ func TestIsPermissible_DetectsNonPermissibleTransactions(t *testing.T) {
...
+
+	header := &types.Header{
+		ParentHash:       maxHash,
+		UncleHash:        maxHash,
+		Coinbase:         maxAddress,
+		Root:             maxHash,
+		TxHash:           maxHash,
+		ReceiptHash:      maxHash,
+		Bloom:            maxBloom,
+		Difficulty:       big.NewInt(math.MaxInt64),
```
facuMH: big ints could use `SetBytes` to set `setToMax(256)` since that is the byte cap rlp enforces, no?

gossip/c_block_callbacks_test.go
```diff
@@ -864,6 +871,325 @@ func TestIsPermissible_DetectsNonPermissibleTransactions(t *testing.T) {
...
+
+	// Create some dummy transactions with gas usage
+	tx1 := types.NewTx(&types.LegacyTx{Nonce: 1, Gas: 1000})
+	tx2 := types.NewTx(&types.LegacyTx{Nonce: 2, Gas: 1000})
+	tx3 := types.NewTx(&types.LegacyTx{Nonce: 3, Gas: 1000})
+
+	// Set up receipts for each transaction
+	receipt1 := &types.Receipt{CumulativeGasUsed: 1000}
+	receipt2 := &types.Receipt{CumulativeGasUsed: 2000}
+	receipt3 := &types.Receipt{CumulativeGasUsed: 3000}
```
HerbertJordan: This test should not work any more. Please fix.

PR #577, opened 2025-11-04
Use NumNodes for fakeGenesis initialization

facuMH:
This PR introduces a validation step during node initialization for fake validators. If a fake validator attempts to start a node but its ID is not present in the validator set defined by the fake genesis, the node initialization will now fail with an error. This ensures that only properly initialized fake validators can participate in the network during testing.


LuisPH3 (approved):
thanks!

PR #592, opened 2025-11-20
Add mechanism to retrieve genesis id from integration test net

LuisPH3:
This PR uses applicationControl to retrieve the genesisId of a node after it has opened its db. 
This is a private mechanism to get access to this parameter that creates no changes in the public interface of the sonic tool. 

Depends on   #591

### DIFF
--- a/cmd/sonicd/app/app.go
+++ b/cmd/sonicd/app/app.go
@@ -20,6 +20,7 @@ import (
 	"os"
 
 	"github.com/0xsoniclabs/sonic/config/flags"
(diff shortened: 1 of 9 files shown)


facuMH (approved):
This PR adds as a channel in `AppControl` to communicate the genesisId, in particular, from the initialization of the sonic client to the integration test net. 
Integration test net initialization verifies that genesisID from all nodes in the same test net are the same.
An integration test is added to verify that different options, when initializing a test net, that would modify the genesis, produce different genesisIDs.

simonlechner (approved):

PR #595, opened 2025-11-21
Block filled with internal transactions only check

simonlechner:
During the Brio test plan discussion, a possible `remainingSize` underflow has been discussed. In the very unlikely case of a block in which the internal transactions exceed the block size limit, this could happen.
This PR adds a fix by simply comparing the transaction size before subtracting it.

### DIFF
--- a/gossip/c_block_callbacks_test.go
+++ b/gossip/c_block_callbacks_test.go
@@ -1162,22 +1162,39 @@ func TestProcessUserTransactions_SponsoredTxSizeIsAccountedCorrectly(t *testing.
 }
 
 func TestProcessUserTransactions_SkipUserTransactionIfInternalTransactionsExceedBlockSizeLimit(t *testing.T) {
-	ctrl := gomock.NewController(t)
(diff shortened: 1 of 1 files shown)


gossip/c_block_callbacks_test.go
```diff
@@ -1161,6 +1161,25 @@ func TestProcessUserTransactions_SponsoredTxSizeIsAccountedCorrectly(t *testing.
 	}
 }
 
+func TestProcessUserTransactions_InternalTransactionsExceedingBlockSizeLimit(t *testing.T) {
```
facuMH: nit:
```suggestion
func TestProcessUserTransactions_SkipsUserTransactions_WhenInternalTransactionsExceedsBlockSizeLimit(t *testing.T) {
```

facuMH (commented):
This PR adds a check to ignore all user transactions when the size of an internal transaction is already over the max block size allowed. 
Looks good, I jut have a couple of test suggestions.

LuisPH3 (dismissed):

HerbertJordan (approved):

PR #596, opened 2025-11-21
TxPool: Validate Local and Remote submittions

facuMH:
This PR expands two TxPool tests which validate rejection of txs to reject also local txs.


LuisPH3 (approved):
thanks!

PR #598, opened 2025-11-21
Add test for the use of brio

LuisPH3:
This test checks usage of RPC methods capable of executing calls at different block heights. The used configuration shall be correct for the choosen block height. 

There are two families of RPC methods doing calls, the ones using `DoCall` and the ones using `StateTransaction`. These functions are the ones which create the configuration need for calling `ApplyMessage` from the RPC. One consumer from each function is used to test the correct behavior.

### DIFF
--- a/tests/revision_forward_test.go
+++ b/tests/revision_forward_test.go
@@ -17,6 +17,7 @@
 package tests
 
 import (
+	"fmt"
 	"math/big"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


tests/revision_forward_test.go
```diff
@@ -195,12 +196,130 @@ func TestRevisionIsForwardedCorrectly_BrioEnablesOsakaInBlockProcessing(t *testi
...
+}
+
+func doRpcCall(client *PooledEhtClient, contractAddress common.Address, callData []byte, blockHash common.Hash) error {
+	// eth_call servers to test functions using the DoCall RPC method
+	_, err := InvokeRpcCallMethod[any]("eth_call", contractAddress, client, blockHash, nil, nil)
+	return err
+}
+
+func InvokeRpcCallMethod[result any](
+	method string, contractAddress common.Address, client *PooledEhtClient, blockHash common.Hash, args ...any) (result, error) {
```
facuMH: style nit:
```suggestion
func InvokeRpcCallMethod[result any](
	method string,
	contractAddress common.Address,
	client *PooledEhtClient,
	blockHash common.Hash,
	args ...any,
	) (result, error) {
```

tests/revision_forward_test.go
```diff
@@ -195,12 +196,130 @@ func TestRevisionIsForwardedCorrectly_BrioEnablesOsakaInBlockProcessing(t *testi
...
+	require.NoError(t, err, "expected trace_call to succeed even if execution fails")
+	require.Contains(t, trace["error"].(string), "invalid opcode", "expected invalid opcode error in trace_call before Brio upgrade")
+}
+
+func doTraceCall(client *PooledEhtClient, contractAddress common.Address, callData []byte, blockHash common.Hash) (map[string]any, error) {
+	// debug_traceCall serves to test functions using StateTransition RPC method
+	config := map[string]any{
+		"tracer": "callTracer",
+	}
+	return InvokeRpcCallMethod[map[string]any]("debug_traceCall", contractAddress, client, blockHash, config)
```
facuMH: if you cast the result like so:
```Go
	result, err := InvokeRpcCallMethod("debug_traceCall", contractAddress, client, blockHash, config)
	return result.(map[string]any), err 
```
then you can drop the generalization from `InvokeRpcCallMethod` and just use `(any, error)` for the return type.

facuMH (dismissed):
Main structure of the test looks correct. I have a couple of suggestions for the helper functions, but nothing critical. Feel free to ignore them if this becomes a urgent.

facuMH (approved):

PR #603, opened 2025-11-25
Introduce event emision throttling mechanism.

LuisPH3:
This PR introduces the event throttler utility inside the event emitter. 
This tool will instruct the event emitter to skip emission if the emitting validator is suppressed. A validator is suppressed If it does not belong to the dominant set of validators, which aggregated stake is larger than the dominantStakeThreshold. 

The throttler tool includes mechanisms to avoid skipping if:
- slowdown of consensus is detected, due to loss of online stake or networking issues. 
- transactions are available to be contributed to the next block 
- this validator does not reach minimum emission required to be considered online, and therefore not slashed for inactivity.


gossip/emitter/emitter.go
```diff
@@ -169,6 +172,8 @@ func NewEmitter(
 		Periodic:      logger.Periodic{Instance: logger.New()},
 		baseFeeSource: baseFeeSource,
 		errorLock:     errorLock,
+		eventEmissionThrottler: *throttling.NewThrottlingState(
+			config.Validator.ID, 0.75, 3, world),
```
facuMH: could you add some documentation for these parameters?

gossip/emitter/throttling/dominant_set_test.go
```diff
@@ -0,0 +1,200 @@
...
+	tests := map[string]struct {
+		stakes            []int64
+		expectedDominated bool
+		expectedSet       []idx.ValidatorID
+	}{
+		"no validators": {
+			stakes:            nil,
+			expectedDominated: false,
+		},
+		"single validator below threshold": {
```
facuMH: nit: comment somewhere near the cases that the threshold is actually a given constant for the whole test. Perhaps define it above `tests` and in test cases declaration use `threshold/2` or `threshold-1`, etc.

gossip/emitter/throttling/dominant_set_test.go
```diff
@@ -0,0 +1,200 @@
...
+		"single validator just below threshold": {
+			stakes:            []int64{99},
+			expectedDominated: false,
+		},
+		"multiple validators below threshold": {
+			stakes:            append(slices.Repeat([]int64{10}, 9), 9),
+			expectedDominated: false,
+		},
+		"multiple validators summing threshold": {
+			stakes:            append(slices.Repeat([]int64{10}, 9), []int64{9, 1}...),
```
facuMH:
```suggestion
			stakes:            append(slices.Repeat([]int64{10}, 9), []int64{10}...), 
```
or 
?
```suggestion
			stakes:            slices.Repeat([]int64{10}, 10),
```

gossip/emitter/throttling/dominant_set_test.go
```diff
@@ -0,0 +1,200 @@
...
+			stakes:            []int64{101},
+			expectedDominated: true,
+			expectedSet:       []idx.ValidatorID{1},
+		},
+		"two validators at threshold": {
+			stakes:            []int64{50, 50},
+			expectedDominated: true,
+			expectedSet:       []idx.ValidatorID{1, 2},
+		},
+		"two validators at threshold out of three": {
```
facuMH: nit:
```suggestion
		"two validators at threshold above three": {
```
not sure which one reads better. wdyt?

facuMH (commented):
This PR adds a mechanism to reduce empty event emition. 

Implementation looks sound. I have a couple of superficial comments before approving.

gossip/emitter/throttling/dominant_set_test.go
```diff
@@ -0,0 +1,200 @@
...
+			stakes:            append(slices.Repeat([]int64{10}, 9), 9),
+			expectedDominated: false,
+		},
+		"multiple validators summing threshold": {
+			stakes:            append(slices.Repeat([]int64{10}, 9), []int64{9, 1}...),
+			expectedDominated: true,
+			// all 11 elements are included
+			expectedSet: []idx.ValidatorID{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11},
+		},
+		"multiple validators just avobe threshold": {
```
HerbertJordan:
```suggestion
		"multiple validators just above threshold": {
```

gossip/emitter/emitter.go
```diff
@@ -169,6 +172,8 @@ func NewEmitter(
 		Periodic:      logger.Periodic{Instance: logger.New()},
 		baseFeeSource: baseFeeSource,
 		errorLock:     errorLock,
+		eventEmissionThrottler: *throttling.NewThrottlingState(
```
HerbertJordan: We might want to have a `nil` option to disable it if desired by the users. Initially we would do an opt-in, then an opt-out, and eventually maybe a forced enabling. Wdyt?

gossip/emitter/emitter.go
```diff
@@ -338,6 +343,14 @@ func (em *Emitter) EmitEvent() (*inter.EventPayload, error) {
 	if e == nil || err != nil {
 		return nil, err
 	}
+
+	// Right after creating the event, check whether to skip its emission
+	// this location allows to take into account the event creation time
+	// and frame in throttling decision.
+	if em.eventEmissionThrottler.SkipEventEmission(e) {
+		return nil, nil
```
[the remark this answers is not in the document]
LuisPH3: I went into that today, there are any metrics for skipped events, only for skipped transactions in events. 
We can add this one and more for the other cases.

gossip/emitter/throttling/dominant_set.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/Fantom-foundation/lachesis-base/inter/pos"
+)
+
+// ComputeDominantSet computes the dominant set of validators whose cumulative stake
+// meets or exceeds the given thresholdStake. It returns the set of validator IDs
+// that make up the dominant set and a boolean indicating whether such a set exists.
+//
+// This function uses the [pos.Validators] object methods to have a deterministic order
+// of validators with equal stakes.
+func ComputeDominantSet(validators *pos.Validators, thresholdStake pos.Weight) (map[idx.ValidatorID]struct{}, bool) {
```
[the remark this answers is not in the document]
LuisPH3: thought quite about it: There are no types which can be forced to be in the range 0-1, that would create invariants and I would have to return errors. 
on the other hand: i return a boolean already which is kind of an anomalous input

gossip/emitter/throttling/dominant_set_test.go
LuisPH3: thanks

gossip/emitter/throttling/dominant_set_test.go
```diff
@@ -0,0 +1,200 @@
...
+	for _, test := range tests {
+		stakes := makeValidators(test.stakes...)
+		set, exist := ComputeDominantSet(stakes, 100)
+		require.Equal(t, test.expectedDominated, exist)
+
+		ids := make([]idx.ValidatorID, 0, len(set))
+		for id := range set {
+			ids = append(ids, id)
+		}
+		require.ElementsMatch(t, test.expectedSet, ids)
```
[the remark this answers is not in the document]
LuisPH3: thanks

gossip/emitter/throttling/throttling.go
```diff
@@ -0,0 +1,132 @@
...
+		return false
+	}
+
+	rules := ts.world.GetRules()
+
+	// Do not skip emission if the event is in the same frame as the last emitted event
+	// for a given period of time.
+	// This means that there no progress in the network can be observed. The stake of this
+	// validator stake may be needed to reach quorum in the current frame.
+	if ts.lastUsedFrame == event.Frame() &&
```
facuMH: could the frame have advanced without `currentSkippedEventsCounts` being reseted ?

gossip/emitter/throttling/throttling.go
```diff
@@ -0,0 +1,132 @@
...
+		ts.currentSkippedEventsCount >= ts.maxSkippedEventsWithSameFrameNumber {
+		return false
+	}
+
+	// Do not skip emission if too many blocks have been missed since the last emission.
+	// This prevents this node from being flagged as inactive, and its stake being slashed.
+	//
+	// TODO: provide some guarantees about boundary conditions here: what happens when
```
facuMH: will this be done as part of a follow-up ?

gossip/emitter/throttling/throttling.go
```diff
@@ -0,0 +1,132 @@
...
+	if currentBlockNumber > ts.lastEmissionBlockNumber &&
+		currentBlockNumber-ts.lastEmissionBlockNumber > blockMissedSlack/2 {
+		return false
+	}
+
+	// Compute dominant set and check if this validator belongs to it.
+	validators, _ := ts.world.GetEpochValidators()
+	dominantSet, dominated := ComputeDominantSet(validators, ts.dominatorsThreshold)
+	if !dominated {
+		return false
```
facuMH: in case of no dominant set possible, then all events are emitted ?

facuMH (commented):
This PR adds an event throttling mechanism that reduces the number of empty events emitted. Mostly it looks good, please see comments.

gossip/emitter/throttling/dominant_set_test.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+		validators := builder.Build()
+
+		set, exists := ComputeDominantSet(validators, 0.7)
+		require.True(t, exists, "total stake is 100, 70% threshold should be found")
+
+		ids := make([]idx.ValidatorID, 0, len(set))
+		for id := range set {
+			ids = append(ids, id)
+		}
```
HerbertJordan: nit:
```suggestion
		ids := slices.Collect(maps.Keys(set))
```

gossip/emitter/throttling/throttling.go
```diff
@@ -0,0 +1,134 @@
...
+type WorldReader interface {
+	GetRules() opera.Rules
+	GetLatestBlockIndex() idx.Block
+	GetEpochValidators() (*pos.Validators, idx.Epoch)
+}
+
+// NewThrottlingState creates a new ThrottlingState for a given validator ID
+// and dominating percentile threshold.
+//
+// dominatingPercentile parameter specifies the fraction of total stake
```
HerbertJordan: nit: about terminology - the [percentile](https://en.wikipedia.org/wiki/Percentile) is the actual value, not the percentage. In this context, it would be more accurate to refer to the `dominatingPercentile` value as a `minimumActiveValidatorFraction` or `minimumLiveValidatorFraction`.

gossip/emitter/throttling/throttling_test.go
```diff
@@ -0,0 +1,186 @@
...
+			),
+		},
+		"non-first validator belongs to dominant set": {
+			validatorID: 2,
+			validators: makeValidators(
+				750, 750, // 75% owned by first two validators
+				125, 125, 125, 125,
+			),
+		},
+	}
```
HerbertJordan: this is twice the same test case ..

gossip/emitter/throttling/throttling_test.go
```diff
@@ -0,0 +1,186 @@
...
+			world.EXPECT().GetLatestBlockIndex().Return(idx.Block(0)).AnyTimes()
+
+			state := NewThrottlingState(test.validatorID, 0.75, 3, world)
+
+			event := inter.NewMockEventPayloadI(ctrl)
+			event.EXPECT().Transactions().Return(types.Transactions{}).AnyTimes()
+			event.EXPECT().Frame().Return(idx.Frame(1)).AnyTimes()
+
+			skip := state.SkipEventEmission(event)
+			require.False(t, skip)
```
HerbertJordan: Right now, the `state` declares the given event to be skipped for the reasons intended by the test. But to verify this, I had to step through the implementation with a debugger. As discussed offline, it would be good to ensure that events are skipped for the intended reasons -- here and in other tests.

Some ideas how this could be covered:
- renaming the function to "ShouldEmit" and returning an error, explaining why not
- returning an enum-type for the made decision
- have a metric counter that tracks the rejection reasons and check those values before and after

gossip/emitter/throttling/throttling.go
```diff
@@ -0,0 +1,132 @@
...
+// as stalled in the current frame.
+func NewThrottlingState(
+	validatorID idx.ValidatorID,
+	dominatingPercentile float64,
+	maxSkippedEventsWithSameFrameNumber uint,
+	stateReader WorldReader,
+) *ThrottlingState {
+	return &ThrottlingState{
+		thisValidatorID:                     validatorID,
+		dominatorsThreshold:                 min(max(dominatingPercentile, 0.7), 1),
```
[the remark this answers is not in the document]
facuMH: could you add a comment about this ?

PR #607, opened 2025-11-26
Update gas subsidies tests to brio hardfork

simonlechner:
This PR updates all gas subsidies tests, which use a fake net for testing to enable brio features. Only tests in which it does not make sense, or it is not necessary to upgrade, have kept unchanged.

### DIFF
--- a/tests/gas_subsidies/registry_update_test.go
+++ b/tests/gas_subsidies/registry_update_test.go
@@ -35,7 +35,7 @@ import (
 
 func TestRegistryUpdate_UpdateContract_SponsoredTransactionsCanBePerformed(t *testing.T) {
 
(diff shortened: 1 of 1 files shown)


tests/gas_subsidies/registry_update_test.go
```diff
@@ -36,7 +36,7 @@ import (
 func TestRegistryUpdate_UpdateContract_SponsoredTransactionsCanBePerformed(t *testing.T) {
 
 	// Enable allegro to support setCode transactions.
```
LuisPH3: Comment is slightly confusing, 

```suggestion
	// Enable a revision with allegro to support setCode transactions.
```

Regarding whenever updating this or not. I think it makes sense to have it updated to test interaction with new hardforks.

PR #609, opened 2025-11-27
Add checks for epoch in sonictool export events.

facuMH:
This PR adds new verifications when running `sonictool events export "from" "to"`:
- Both "from" and "to" of the range given by commandline are older than the newest epoch.
- Both the "from" and "to" epochs have at least one event each. 
- "from" must be a smaller number than "to".

In case of failures, exit code 1 is returned with the appropriate error.


matejmode (dismissed):
Looks good. Thanks for implementing this.

PR #610, opened 2025-12-01
Event throttling config

facuMH:
This PR implements https://github.com/0xsoniclabs/sonic-admin/issues/513

This PR adds 3 config parameters to both config file import/export and the possibility to provide them via command line flags.

Note that a validation step is added to ensure that the Dominant Set Threshold is a float value in [0,1]. This validation is added as part of `MakeAllConfigs`.

### DIFF
--- a/config/flags/flags.go
+++ b/config/flags/flags.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/0xsoniclabs/sonic/evmcore"
 	"github.com/0xsoniclabs/sonic/gossip"
+	"github.com/0xsoniclabs/sonic/gossip/emitter"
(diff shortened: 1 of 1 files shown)


gossip/emitter/config_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+package emitter
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestEmitterConfig_ValidateConfig_ReportsError_ForInvalidDominatingThreshold(t *testing.T) {
```
LuisPH3: do not forget the happy case.

PR #614, opened 2025-12-10
event throttler integration test

facuMH:
This PR adds an integration test that verifies the percentage of events emitted.

### DIFF
--- a/gossip/ethapi_backend.go
+++ b/gossip/ethapi_backend.go
@@ -283,10 +283,10 @@ func (b *EthAPIBackend) GetHeads(ctx context.Context, epoch rpc.BlockNumber) (he
 	}
 
 	if heads == nil {
-		return nil, errors.New("heads not found")
+		heads = hash.Events{}
 	}
 
-	return heads, nil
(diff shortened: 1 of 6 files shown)


gossip/ethapi_backend.go
```diff
@@ -283,10 +283,10 @@ func (b *EthAPIBackend) GetHeads(ctx context.Context, epoch rpc.BlockNumber) (he
 	}
 
 	if heads == nil {
-		heads = hash.Events{}
+		return nil, errors.New("heads not found")
```
LuisPH3: Unfortunately, this is not a change we can do. 
This is a change in the RPC interface whit the users.

LuisPH3 (changes requested):
This is the first time that I get to need this integration test, nevertheless there is one change that we cannot roll out in a patch version.

gossip/ethapi_backend.go
facuMH: sure. I will just move the empty events check to GetEventsFromEpoch.

PR #616, opened 2025-12-16
Reduce `GetTransactionReceipt` block store accesses

facuMH:
This PR fixes https://github.com/0xsoniclabs/sonic-admin/issues/508

It was observed during performance analysis of `large_transaction_test` that a large amount of memory was allocated for `GetTransactionReceipt` which made a redundant query to the blocks store.

The changes proposed add a method to the backend, so that receipts can be obtained from a single call to the block store. 

Here is a memory profile before the changes proposed in this PR:
memory_profile_before

And here is a memory profile after the changes proposed:
memory_profile_after

### DIFF
--- a/ethapi/api.go
+++ b/ethapi/api.go
@@ -1955,7 +1955,7 @@ func (s *PublicTransactionPoolAPI) GetTransactionReceipt(ctx context.Context, ha
 	if block == nil || err != nil {
 		return nil, err
 	}
-	receipts := s.b.GetReceiptsFromBlock(block)
+	receipts := s.b.FetchReceiptsForBlock(block)
 	if receipts == nil || err != nil {
 		return nil, err
 	}
--- a/gossip/ethapi_backend.go
+++ b/gossip/ethapi_backend.go
@@ -355,10 +355,10 @@ func (b *EthAPIBackend) GetReceiptsByNumber(ctx context.Context, number rpc.Bloc
 	number = rpc.BlockNumber(blockNumber)
 
 	block := b.state.GetBlock(common.Hash{}, uint64(number))
-	return b.GetReceiptsFromBlock(block), nil
+	return b.FetchReceiptsForBlock(block), nil
 }
 
-func (b *EthAPIBackend) GetReceiptsFromBlock(block *evmcore.EvmBlock) types.Receipts {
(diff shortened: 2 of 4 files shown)


gossip/ethapi_backend.go
```diff
@@ -355,18 +355,22 @@ func (b *EthAPIBackend) GetReceiptsByNumber(ctx context.Context, number rpc.Bloc
 	number = rpc.BlockNumber(blockNumber)
 
 	block := b.state.GetBlock(common.Hash{}, uint64(number))
+	return b.MakeReceiptsFromBlock(block), nil
+}
+
+func (b *EthAPIBackend) MakeReceiptsFromBlock(block *evmcore.EvmBlock) types.Receipts {
```
LuisPH3:
```suggestion
func (b *EthAPIBackend) FetchReceiptsFromBlock(block *evmcore.EvmBlock) types.Receipts {
```

LuisPH3 (approved):
This is a nice little cleanup. It may be an insight for a larger cleanup in the RPC backend interface.

gossip/ethapi_backend.go
HerbertJordan: Why "MakeReceiptsFromBlock" instead of "GetReceiptsForBlock"?

Make kinda implies that all the data is in the block, and the function is just extracting it from the block and re-formatting it. Get, on the other hand, implies that there are data base accesses involved.

HerbertJordan (approved):

ethapi/api.go
```diff
@@ -1955,7 +1955,7 @@ func (s *PublicTransactionPoolAPI) GetTransactionReceipt(ctx context.Context, ha
 	if block == nil || err != nil {
 		return nil, err
 	}
-	receipts := s.b.MakeReceiptsFromBlock(block)
+	receipts := s.b.GetReceiptsFromBlock(block)
```
LuisPH3: nit: get would be for a getter, which  is cheap. this is a database operation, and get does not carry the weight of the cost of the operation.  Read or Fetch would be better. 

Additionally, Go good practices recommend not using Get prefix, the avoid aliasing REST interface methods. Using Get in the domain of RPC gets dangerously close to REST.

LuisPH3 (approved):

HerbertJordan (approved):

PR #618, opened 2025-12-16
Update event throttler config

LuisPH3:
This PR updates flags and config to reflect the latests changes to the event throttler module.

- introduce heartbeat flag/config setting
- aggregate config in its own struct, to avoid copy of independent parameters into the throttler module. 
- replace `throttle-skip-in-same-frame`  with `event-throttler.dominant-threshold`. Throttler will not count frames anymore, therefore name had to be updated. 
- update flags and usage

### DIFF
--- a/gossip/emitter/config/config.go
+++ b/gossip/emitter/config/config.go
@@ -132,7 +132,7 @@ func DefaultThrottlerConfig() ThrottlerConfig {
 		Enabled:                false,
 		DominantStakeThreshold: 0.75,
 		DominatingTimeout:      3,
-		NonDominatingTimeout:   10,
+		NonDominatingTimeout:   100,
 	}
 }
 
--- a/tests/flags/flags_test.go
+++ b/tests/flags/flags_test.go
@@ -50,7 +50,7 @@ func TestSonicTool_DefaultConfig_HasDefaultValues(t *testing.T) {
 Enabled = false
 DominantStakeThreshold = 7.5e-01
 DominatingTimeout = 3
-NonDominatingTimeout = 10`)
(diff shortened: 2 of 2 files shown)


config/flags/flags.go
```diff
@@ -356,24 +356,29 @@ var (
...
-		Value: emitter_config.DefaultConfig().ThrottlerDominantThreshold,
+		Name:  "event-throttler.dominant-threshold",
+		Usage: "The percentage of total stake which defines the dominating validator set. This value must be between 0.7 and 1.0.",
+		Value: emitter_config.DefaultConfig().ThrottlerConfig.DominantSetThreshold,
 	}
-	ThrottlingSkipInSameFrameFlag = cli.UintFlag{
-		Name:  "emitter.throttle-skip-in-same-frame",
-		Usage: "Maximum number of skipped events within the same frame to force event emission",
-		Value: emitter_config.DefaultConfig().ThrottlerSkipInSameFrame,
+	ThrottlingStalledFrameTimeout = cli.Uint64Flag{
```
HerbertJordan: The unit of this parameter is `Attempts`, not `Frames`. Also, the term `Stalled` may also refer to the stopping of the emission of events in case of a disconnect to the network.

How about renaming this to `ThrottlingDominatingEmissionTimeout` to match the flag name?

facuMH (approved):
This PR adds a field to the throttler config, namely `ThrottlingHeartbeatTimeout`, and does a relocation/rename of the pre-existing 3 throttler config options and their corresponding flags. 

Could you ensure this also works when importing/exporting config to a toml file?

PR #626, opened 2026-01-09
Integrate event throttler

LuisPH3:
This PR integrates the event throttler in the emitter package. This PR enables this feature for the sonic client. 

The changes include both unit tests for new and older parts of the emitter code, and integration tests:

- Unit test `TestNewEmitter_InitializesThrottler_WhenThrottlerConfigPresent` regarding configuration of the throttler object is removed, this is because a new configuration object only requires copying the config values and throttler being active or not is handled (and tested)  from within the object. 
- Add emitter Unit tests for some untested code forks: not eligible for emission because of several causes. 
- Add emitter unit test to test new adapter tool regarding reading events from other validators: Because the store API requires two calls to retrieve an event, it is technically possible that he event hash is known but the event itself is not. This is a severe runtime error and an error is logged. This error should not occur in production, having it happening indicates severe failure in the system. 
- An integration test is added `TestEventThrottler_NonDominantValidatorsProduceLessEvents_WhenEventThrottlerIsEnabled` checks that event emission is reduced when the feature is enabled. 
- The integration test `TestValidatorsStakes_AllNodesProduceBlocks_WhenStakeDistributionChanges` is updated to enable the feature, proving that validators can actually keep producing blocks even when not being part of the dominating set.

### DIFF
--- a/tests/validator_stakes/validator_stakes_test.go
+++ b/tests/validator_stakes/validator_stakes_test.go
@@ -17,6 +17,7 @@
 package many
 
 import (
+	"fmt"
 	"math/big"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


tests/single_proposer/single_proposer_protocol_test.go
```diff
@@ -252,23 +251,13 @@ func getUsedEventVersion(
...
-		Epoch hexutil.Uint64
-	}{}
-	err := client.Client().Call(&block, "eth_getBlockByNumber", rpc.BlockNumber(-1), false)
-	require.NoError(err)
-
-	// Get the head events of the current epoch.
-	heads := []hexutil.Bytes{}
-	err = client.Client().Call(&heads, "dag_getHeads", rpc.BlockNumber(block.Epoch))
-	require.NoError(err)
+	heads := tests.GetLatestEvents(t, client)
```
HerbertJordan: Not sure why this change is in this PR. Are those API functions no longer supported?

tests/validator_stakes/low_stake_validator_does_not_emit_test.go
```diff
@@ -0,0 +1,207 @@
...
+				ClientExtraArguments: clientExtraArgs,
+			})
+
+			client, err := net.GetClient()
+			require.NoError(t, err)
+			defer client.Close()
+
+			var currentRules opera.Rules
+			err = client.Client().Call(&currentRules, "eth_getRules", "latest")
+			require.NoError(t, err)
```
HerbertJordan: Why are the network rules requested here?

tests/test_utils.go
```diff
@@ -491,3 +493,30 @@ func WaitForProofOf(t *testing.T, client *PooledEhtClient, blockNumber int) {
...
+	require.NoError(t, err)
+
+	return GetEventsFromEpoch(t, client, int(block.Epoch))
+}
+
+func GetEventsFromEpoch(t *testing.T, client *PooledEhtClient, epoch int) []hash.Event {
+
+	// Get the head res of the given epoch.
+	res := []string{}
+	err := client.Client().Call(&res, "dag_getHeads", rpc.BlockNumber(epoch))
```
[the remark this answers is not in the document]
LuisPH3: https://github.com/0xsoniclabs/sonic/blob/e43b890bc4d99ca9c5015be04bf3f3e58878c52e/gossip/ethapi_backend.go#L281

nop, it will not

tests/validator_stakes/low_stake_validator_does_not_emit_test.go
```diff
@@ -0,0 +1,207 @@
...
+
+		currentEpoch := getCurrentEpoch(t, client)
+		if currentEpoch != targetEpoch {
+			// If the test fails here, is because the test time is close to the duration of an epoch
+			// and one of the nodes already moved into the next epoch, because of the rpc only serving
+			// events within the current epoch, the test cannot longer be conducted.
+			require.Fail(t, "node %d moved to the next epoch, and the test cannot longer be conducted", i)
+		}
+
+		// Get the current epoch.
```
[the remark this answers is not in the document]
LuisPH3: gone

gossip/emitter/emitter_test.go
```diff
@@ -256,45 +256,209 @@ func TestEmitter_CreateEvent_InvalidValidatorSetIsDetected(t *testing.T) {
...
+		world: World{
+			External:     world,
+			EventsSigner: signer,
+			TxPool:       txPool,
+		},
+	}
+	em.validators.Store(validators)
+
+	baseFeeSource := NewMockBaseFeeSource(ctrl)
+	baseFeeSource.EXPECT().GetCurrentBaseFee()
```
facuMH: no return value? I guess this returns then the zero value, since base fee is `*big.Int`, it would return nil, is base fee nil not checked?

facuMH (approved):
This PR adds unit tests and integration test which verify the expected behaviour of the event throttler.

PR #627, opened 2026-01-09
Archive access returns block instead of header

facuMH:
This PR modiefies `StateAndHeaderByNumberOrHash` so that it returns a block instead of the header, also changing the function name to `StateAndBlockByNumberOrHash`. 

The changes have been organised so that each commit contains a different type of changes for easier reviewing.


HerbertJordan (dismissed):
Looks good to me.

LuisPH3 (approved):
As discussed offline, it would be ingesting to see which other uses of `getHeader` exist, which could create the same patter of accessing the block twice seen before.

PR #628, opened 2026-01-12
Run `tests` folder tests first

facuMH:
This PR modifies the Makefile so that `test` and `coverage` targets execute first the tests package and the rest.
The motivation for this change being that the tests package is the one whose tests take the longest, and to take better advantage of the go tests package parallelisation, the tests package and its sub packages should start running first.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,14 +32,23 @@ sonic-image:
 		--network=host \
 		-f ./Dockerfile -t "sonic:$(TAG)" .
 
+# Define the specific path to your slow integration tests
+SLOW_PKG := ./tests/...
+
+# Find all other packages by listing everything and filtering out the slow package
+# We use 'shell' to execute the go list command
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -32,14 +32,23 @@ sonic-image:
 		--network=host \
 		-f ./Dockerfile -t "sonic:$(TAG)" .
 
+# Define the specific path to your slow integration tests
+SLOW_PKG := ./tests/...
```
LuisPH3: using := instead of = ? why?

LuisPH3 (changes requested):
How can I be sure that the tests are executed in the desired order?  Is the order of arguments in the go command defined somewhere? 

Can you add a makefile linter before complicating the script even further? 
I think there is a serious maintainability problem in adding logic to makefile, it should remain purely declarative. I am not sure if the gains justify the change.

PR #629, opened 2026-01-14
Add metrics to monitor skipping event emission.

LuisPH3:
This PR add monitor metrics for the event emitter. Reasons for skipping event emission are added to the already existing metrics set.


facuMH (approved):
This PR adds 3 metrics, each reporting a different reason why an event was skipped.

PR #635, opened 2026-01-16
State Reader interface

facuMH:
This PR adds a `StateReader` interface and refactors the exposed methods to remove one unused and a redundant one with a single use. 

This will enable easier ways of more accurately testing interactions that depend on the evm state without having to initialize all the necessary components.

Also a parameter is added to the function  `LastBlockWithArchiveState`. Since it requires a change on the interface, the change is added as part of this PR. `LastBlockWithArchiveState` is called in a single location, and now that locations forwards the `false` value as was before, to preserve current behaviour.

### DIFF
--- a/gossip/evm_state_reader.go
+++ b/gossip/evm_state_reader.go
@@ -79,7 +79,8 @@ type EvmStateReader struct {
 
 // CurrentBaseFee returns the base fee of the most recent block.
 func (r *EvmStateReader) CurrentBaseFee() *big.Int {
-	return new(big.Int).Set(r.CurrentBlock().BaseFee)
+	res := r.store.GetBlock(r.store.GetLatestBlockIndex()).BaseFee
+	return new(big.Int).Set(res)
 }
 
 // CurrentMaxGasLimit returns the maximum gas limit of the most recent epoch.
@@ -96,8 +97,7 @@ func (r *EvmStateReader) CurrentMaxGasLimit() uint64 {
 
 // CurrentConfig returns the chain config applicable to the most recent block.
 func (r *EvmStateReader) CurrentConfig() *params.ChainConfig {
-	blockNumber := idx.Block(r.CurrentBlock().Number.Uint64())
-	return r.store.GetEvmChainConfig(blockNumber)
+	return r.store.GetEvmChainConfig(r.store.GetLatestBlockIndex())
 }
 
 // CurrentRules returns the rules applicable to the most recent epoch.


gossip/evm_state_reader.go
```diff
@@ -31,6 +31,23 @@ import (
 	"github.com/0xsoniclabs/sonic/opera"
 )
 
+//go:generate mockgen -source=evm_state_reader.go -destination=mock_evm_state_reader.go -package=gossip
+
+type StateReader interface {
+	GetCurrentBaseFee() *big.Int
+	MaxGasLimit() uint64
+	Config() *params.ChainConfig
```
LuisPH3: We know that the chainconfig in sonic is not a static object. It depends on the block height it is queried for. 
Can we rename this method to `CurrentConfig`, or better: accept a `blockHeight` argument ?

gossip/evm_state_reader.go
```diff
@@ -31,6 +31,23 @@ import (
 	"github.com/0xsoniclabs/sonic/opera"
 )
 
+//go:generate mockgen -source=evm_state_reader.go -destination=mock_evm_state_reader.go -package=gossip
+
+type StateReader interface {
+	GetCurrentBaseFee() *big.Int
+	MaxGasLimit() uint64
```
LuisPH3: can we rename this to `currentMaxGasLimit` ?

gossip/evm_state_reader.go
```diff
@@ -31,6 +31,23 @@ import (
...
+	Config() *params.ChainConfig
+	GetCurrentRules() opera.Rules
+	CurrentBlock() *evmcore.EvmBlock
+	CurrentHeader() *evmcore.EvmHeader
+	LastBlockWithArchiveState(withTxs bool) (*evmcore.EvmBlock, error)
+	GetHeaderByNumber(n uint64) *evmcore.EvmHeader
+	GetHeader(h common.Hash, n uint64) *evmcore.EvmHeader
+	GetBlock(common.Hash, uint64) *evmcore.EvmBlock
+	GetTxPoolStateDB() (state.StateDB, error)
+	GetRpcStateDB(*big.Int, common.Hash) (state.StateDB, error)
```
LuisPH3: What are the arguments of this method? block height I presume...

gossip/evm_state_reader.go
```diff
@@ -31,6 +31,23 @@ import (
 	"github.com/0xsoniclabs/sonic/opera"
 )
 
+//go:generate mockgen -source=evm_state_reader.go -destination=mock_evm_state_reader.go -package=gossip
+
+type StateReader interface {
```
LuisPH3: I thought of this after reviewing the interface methods: 
The scope is to minimize the changes needed to introduce this interface, nevertheless the names and arguments are extremely confusing. 
We can refactor all the method names, or we can add exhaustive documentation in this interface about what are these names and what are supposed to do or how to be used. 

I am inclined to do the full refactor, wdyt?

facuMH: I agree, we can do a refactor to rename methods/parameters.

gossip/evm_state_reader.go
```diff
@@ -31,6 +31,23 @@ import (
...
+	GetCurrentBaseFee() *big.Int
+	MaxGasLimit() uint64
+	Config() *params.ChainConfig
+	GetCurrentRules() opera.Rules
+	CurrentBlock() *evmcore.EvmBlock
+	CurrentHeader() *evmcore.EvmHeader
+	LastBlockWithArchiveState(withTxs bool) (*evmcore.EvmBlock, error)
```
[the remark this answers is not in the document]
facuMH: "current" fetches the last block of the current epoch, while "last with archive state" fetches the last block number from the archive. These two are not necessarily the same, depending on store-livedb-archive sync state.

evmcore/tx_pool.go
```diff
@@ -165,14 +165,14 @@ const (
...
+	ReadOnlyStateDB() (state.StateDB, error)
+	CurrentBaseFee() *big.Int
+	CurrentMaxGasLimit() uint64
 	SubscribeNewBlock(ch chan<- ChainHeadNotify) notify.Subscription
-	Config() *params.ChainConfig
-	GetCurrentRules() opera.Rules
-	GetHeader(common.Hash, uint64) *EvmHeader
+	CurrentConfig() *params.ChainConfig
+	CurrentRules() opera.Rules
+	Header(common.Hash, uint64) *EvmHeader
```
LuisPH3: This method has arguments in a different order than get block. They are unammed and not documented. Can you clean this up? 
Additionally, I do not see the need of having both block and header methods.

PR #638, opened 2026-01-29
Add unique-wrapper for iterators

HerbertJordan:
This PR adds a wrapper utility to the leap-join package making an underlying iterator unique.

This is a building block for composing more complex iterators for which distinctness of listed elements can not be guaranteed.

### DIFF
--- a/utils/leap/unique.go
+++ b/utils/leap/unique.go
@@ -1,0 +1,67 @@
+// Copyright 2025 Sonic Operations Ltd
+// This file is part of the Sonic Client
+//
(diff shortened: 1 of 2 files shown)


HerbertJordan:
> Could you add a unit test for Next? Even if it is used in the test for Seek, I think the Next method has enough logic to be unit tested on its own.

@facuMH the `Next` and `Current` functions are extensively tested as part of the `TestUnique_FiltersDuplicates` test, via the `All` utility function. Do you think more is needed?

PR #642, opened 2026-02-05
Integrate sfvm starting with brio

simonlechner:
This PR integrates the sfvm and uses it as the default interpreter starting from the brio hardfork. Once brio is enabled, a sync test is going to be performed, allowing the complete removal of the lfvm.

### DIFF
--- a/opera/vm_config.go
+++ b/opera/vm_config.go
@@ -71,10 +71,6 @@ var sfvmFactory = func() vm.InterpreterFactory {
 func GetVmConfig(rules Rules) vm.Config {
 	res := sonicVmConfig
 
-	if rules.Upgrades.Brio {
-		res.Interpreter = sfvmFactory
-	}
-
 	// don't charge excess gas in single proposer mode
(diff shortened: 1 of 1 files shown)


opera/vm_config.go
```diff
@@ -55,11 +56,25 @@ var sonicVmConfig = func() vm.Config {
...
+	return geth_adapter.NewGethInterpreterFactory(interpreter)
+}()
+
 // GetVmConfig returns the VM configuration to be used for processing
 // transactions under the given network rules.
 func GetVmConfig(rules Rules) vm.Config {
 	res := sonicVmConfig
 
+	if rules.Upgrades.Brio {
+		res.Interpreter = sfvmFactory
```
LuisPH3: I am not sure about this.
The sfvm is a fully compatible evm interpreter, why would it only be enabled to execute a part of the block history? 
The parametrization will change over the block history, but the interpreter used should be orthogonal.

LuisPH3 (approved):

PR #643, opened 2026-02-06
Prototype transaction bundles

LuisPH3:
This is a prototype for transaction bundles, not fit for production.
This sketch serves as an end-to-end proof-of-concept. 
This PR serves as a discussion board for the development of the new feature. Comments are welcome. 

This PR is a prof of concept:
CHANGES IN THIS PR ARE TO BE EXPECTED AND INTERFACES WILL NOT BE PRESERVED.

Reading starts at `tests/bundle/make_bundle_test.go` where the use case is illustrated. More
use cases can be found in `tests/bundle/bundle_transaction_test.go`

The reading continues by inspecting the commits one at the time, in the order they are presented. 
The `dbState` implementation found in file `evmcore/transactional_state/dbTransaction.go` is entirely untested and serves the purpose of providing the rollback functionality. This code is not to be merged into production.

- [ ] This branch is NOT to be merged.


gossip/blockproc/bundle/bundle.go
```diff
@@ -0,0 +1,256 @@
...
+)
+
+var (
+	BundleOnly    = common.HexToAddress("0x00000000000000000000000000000000000B0D1E")
+	BundleAddress = common.HexToAddress("0x00000000000000000000000000000000B0D1EADD")
+)
+
+// MetaTransaction represents the essential information of a transaction
+// that is relevant for the execution of a bundle.
+type MetaTransaction struct {
```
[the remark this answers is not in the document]
LuisPH3: TxData has too much information about pricing: 
- It wraps the duality of GasPrice and DynamicPricing functionality, which is something I would prefer not have to deal with in this layer.
- I am not sure how much of the pricing scheme is needed to describe "what does the transaction have to do" 

I am well aware of the structure of transaction types, the question here is: what does the bundle care for?

gossip/blockproc/bundle/bundle.go
```diff
@@ -0,0 +1,256 @@
...
+// shall not be included in the chain.
+func (tb *TransactionBundle) Validate(signer types.Signer) error {
+
+	executionPlan, err := tb.ExtractExecutionPlan(signer)
+	if err != nil {
+		return fmt.Errorf("failed to get execution plan: %v", err)
+	}
+
+	// validate payment tx
+	if !BelongsToExecutionPlan(tb.Payment, executionPlan.Hash()) {
```
HerbertJordan: nit: That may be too restrictive. It would be sufficient for the payment transaction to be bundle-only transaction. There should be no need for it to re-confirm the execution plan.

gossip/blockproc/bundle/bundle.go
```diff
@@ -0,0 +1,256 @@
...
+type ExecutionFlag uint16
+
+const (
+	// FlagIgnoreInvalid indicates that invalid transactions within the bundle
+	// should be ignored, and not cause the entire bundle to revert. If the
+	// flag is present, the ignored transactions will no be part of the block
+	// and no receipt will be generated for them. The execution of the bundle
+	// will continue with the next transaction in the bundle.
+	FlagIgnoreInvalid ExecutionFlag = 1 << 0
+	// FlagIgnoreReverts indicates that failed transactions within the bundle
```
HerbertJordan: nit: just about the wording, we should try to avoid to introduce to many synonyms. A "failed" transaction and a "reverted" transaction seems to be the same here.

evmcore/state_processor.go
```diff
@@ -287,6 +307,64 @@ func (r *transactionRunner) runSponsoredTransaction(
...
+func (r *transactionRunner) runTransactionBundle(
+	ctxt *runContext,
+	tx *types.Transaction,
+	txIndex int,
+) []ProcessedTransaction {
+
+	bundle, err := bundle.UnpackTransactionBundle(tx)
+	if err != nil {
+		log.Warn("Transaction bundle execution failed:", "tx", tx.Hash().Hex(), "error", err)
+		return nil
```
HerbertJordan: As you might remember, we had a length discussion on what to return as a "ProcessedTransaction" back when we introduced gas subsidies. You insisted on returning all transactions that are attempted to be executed, not just those that are included in the end. So, to stay aligned with the other run functions, at this point it should return 
```suggestion
		return []ProcessedTransaction{{Transaction: tx}}
```
where the lack of a receipt indicates that the transaction was skipped.

Same for some other cases below.

evmcore/state_processor.go
```diff
@@ -287,6 +307,64 @@ func (r *transactionRunner) runSponsoredTransaction(
...
+				payment,
+			}
+		}
+
+		res = append(res, txResult)
+
+		if bundle.StopAfterFirstSuccessfulTransaction() && txResult.Receipt != nil && txResult.Receipt.Status == types.ReceiptStatusSuccessful {
+			log.Info("Bundled transaction succeeded, stop after first", "tx", btx.Hash().Hex())
+			break
+		}
```
HerbertJordan: That does not reflect the intended semantic of the "StopAfterFirstSuccess" flag. The flag should allow you to do something like this:
```Go
   for _, tx := range bundle.Bundle {
         receipt := run(tx)
         if success(receipt) {
                break
          }
    }
```
where `success` of a transaction is something like this (don't take this as fixed, I have not tested it):
```Go
func success(receipt) bool {
    if receipt == nil {
        return flag.IgnoreInvalid
    }
    if receipt.Status == Revert {
         return flag.IgnoreReverted
    }
    return receipt.Status == Success
}
```

The way you implemented it here makes the "StopAfterFirstSuccess" not very useful.

gossip/blockproc/bundle/bundle.go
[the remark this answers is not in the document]
LuisPH3: I have been thinking quite a lot about this. Using the same transaction is not that straight forward:

- hashing: because hashing includes the access lists, it is not the same to hash the transaction before or after it has included the hash of the execution plan into the storage entry. This would require an alternative hashing scheme that allows to ignore this field. Singing the execution plan creates a chinken-and-the-egg problem.

- Usability: using the same type for both preparing the execution plan and submitting the final transaction requires signing twice the same transaction. Once without execution plan hash, once with it. This creates a weird data flow which is counter-intuitive and error-prone on the user. The data flow should be one directional.

- Compatibility: I am seriously considering the allowance of type 1 and 2 only transactions into the bundles, it does not make any sense to provide further support to blobs in the sonic network and the inclusion of 7702 transactions into a bundle creates very complicated workflows, for which would have to do further risk analysis. An MVP does not need to support other transactions than those which execute contracts and transfer values. Everything that seems accessory should go from this release.

evmcore/state_processor.go
LuisPH3: yes, this code will have to be carefully rewritten nevertheless.

PR #644, opened 2026-02-11
Add adapter of kvdb.Iterator to leap.Iterator

HerbertJordan:
This PR introduces and implementation of a `leap.Iterator` based on the iterator and table infrastructure provided by the gossip store to iterate over log-record entries in the log index.

Its key features are:
- the ability to iterate over and parse the log-index table
- the ability to Seek in the log index
- the handling of errors during the iteration process

This is a building block required for performing log filtering using the leap-join algorithm (see #645).


topicsdb/leap_join_test.go
```diff
@@ -0,0 +1,425 @@
...
+	gomock.InOrder(
+		iter.EXPECT().Next().Return(true),
+		iter.EXPECT().Key().Return(key),
+		iter.EXPECT().Value().Return([]byte{5}),
+		iter.EXPECT().Error().Return(nil),
+	)
+
+	require.True(it.Next())
+	require.Equal(iter, it.iter)
+	require.True(ctrl.Satisfied())
```
facuMH: is this here just to verify/enforce that the 4 expects were triggered by the `Next` call in line 87 ? or is this unnecessary ?

PR #647, opened 2026-02-17
Add integration test for log filtering

HerbertJordan:
This PR introduced an integration test checking the support for log-filtering.

### DIFF
--- a/tests/rpcs/get_logs_test.go
+++ b/tests/rpcs/get_logs_test.go
@@ -47,6 +47,7 @@ func TestGetLogFilters(t *testing.T) {
 	for range N {
 		source, receipt, err := tests.DeployContract(net, indexed_logs.DeployIndexedLogs)
 		require.NoError(t, err)
+		require.Equal(t, types.ReceiptStatusSuccessful, receipt.Status)
 		sources = append(sources, source)
 		sourceAddrs = append(sourceAddrs, receipt.ContractAddress)
 	}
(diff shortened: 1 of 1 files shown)


facuMH (dismissed):
This PR adds a contract which generates a large amount of logs and an integration test that queries different combinations of those logs.

tests/rpcs/get_logs_test.go
```diff
@@ -0,0 +1,340 @@
...
+
+	net := tests.StartIntegrationTestNet(t, tests.IntegrationTestNetOptions{})
+
+	// Deploy multiple instances of the `IndexedLogs` contract, which emits a
+	// Cartesian product of logs to be sliced by filter tests.
+	sources := []*indexed_logs.IndexedLogs{}
+	sourceAddrs := []common.Address{}
+	for range N {
+		source, receipt, err := tests.DeployContract(net, indexed_logs.DeployIndexedLogs)
+		require.NoError(t, err)
```
LuisPH3: nit: missing receipt success check.

tests/rpcs/get_logs_test.go
```diff
@@ -0,0 +1,340 @@
...
+			}
+			if len(logs) == 0 {
+				numEmpty++
+			}
+		})
+	}
+
+	// Smoke-test that the reference filter implementation is not broken.
+	require.Equal(t, 1, numFull, "exactly one test case should return the full set of logs")
+	require.Less(t, numEmpty, len(tests)-1, "at least one test case should return a true subset of logs")
```
LuisPH3: adding an expectation to each case on the number of returned logs would be more robust that this check, would that be too verbose?

LuisPH3 (dismissed):
The PR adds integration testing for log filtering RPC features. The test contains an extensive list of interesting cases.
Some minor comments:

LuisPH3 (approved):

PR #649, opened 2026-02-17
Switch to leap-join based log filtering

HerbertJordan:
This PR switches from the thread-pool based log filtering algorithm to the leap-join based algorithm.
It also deletes the old implementation.

The new algorithm preserves the same performance for simple queries, using alternatives only for a single topic:
image

For non-trivial queries involving multiple topics, the new algorithm enables the processing of queries that timed out previously:
image


facuMH (approved):
This PR replaces the old thread thread-pool based log indexer for the new leap-join, which shows remarkable performance improvements for non-trivial log filters. 
Nice clean up !!

PR #656, opened 2026-02-20
Use estimateGas instead of intrinsic gas for test transactions

LuisPH3:
This PR add real gas estimation for contract calls, to enable testing of more complicated transactions. 
The old test tooling would compute intrinsic gas requirements, but any transaction doing a contract call will be underfunded. 
The new code uses the existing gas estimation tools to make test transactions' generation easier.


facuMH (approved):
This PR applies a series of refactors in integration tests regarding transactions gas and gas estimation.

PR #658, opened 2026-02-20
Always retrieve transaction from Txs table directly 

LorenzSchueler:
[NOMERGE]

Problem: Transaction indices are needed for the transaction pool API.
Open Question: **Is it necessary to be able to query skipped transactions?**

---

This PR tries to remove the `TxPositions` table of the gossip DB and all related functionality.
Transactions can be queried directly from the `Txs` table instead. However, the positions within the block/event are needed in other places. 

- The `TxPositions` table is accessed using `GetTxPosition` and `SetTxPosition` in `store_tx_position.go`.
- `GetTxPosition` is only used in `ethapi_backend.go` for `GetTxPosition` and `GetTransaction`.
    - `GetTxPosition` in `ethapi_backend.go` is obsolete if the positions are not needed elsewhere.
    - `GetTransaction` queries the transaction position and used it to index into the events to obtain the transaction. It also returns the indices.

The question is whether all transactions are also in the `Txs` table and could be queried directly, and whether the indices which are retuned from `GetTransaction` are needed on the caller side?
The former question can be answered by checking that in all places where `SetTxPosition` is called, the transaction is also stored directly using `SetTx`.

- `SetTxPosition` is only called in `c_block_callbacks.go` and `c_llr_callbacks.go`.
- In `c_llr_callbacks.go` `SetTx` is called directly before `SetTxPosition`, so transactions can be retrieved directly without needing to go through the events.
- In `c_block_callbacks.go` `SetTxPositions` is called with `evmBlock.Transactions`, and `SetTx` is called with `blockBuilder.GetTransactions()`.
    - `blockBuilder.GetTransactions()` returns all non-skipped transactions (i.e., those with receipts).
    - `evmBlock` is obtained from `evmProcessor.Finalize()`.
       The processor gets the transactions passed to `processor.Execute()` and adds them no matter whether they are skipped or not.
    - `GetTx` will therefore only return non-skipped transactions, while `GetTxPosition` also returns positions for skipped transactions, which can then be queried from the events.

As mentioned above, `GetTransaction` returns the transaction AND its indices in the block and event. `GetTransaction` is called in `api.go`. There the indices in the blocks and events are used, because they are part of some types that get returned as part of the `PublicTransactionPoolAPI`.
    
Conclusion:
By removing the `TxPositions` table and implementing `GetTransactions` by a lookup of `Txs`, skipped transactions can no longer be queried. Non-skipped transactions should not be effected by this change. Also, `GetTransaction` can no longer return the indices of the transaction in the block and event.


LorenzSchueler:
@HerbertJordan @LuisPH3 I put this as future reference here, but I'll close it for now.

PR #669, opened 2026-02-23
RPC test eth_createAccessList preserves Bundle Only marker

jenikd:
This PR adds an integration test for rpc method `eth_createAccessList` and verifies that `BundleOnly` marker, in form of defined address, preserves in the access list when included in the provided transaction arguments.


LuisPH3:
Can you place this test in `tests/bundles` ? 
The tests folder is getting too large, we can group all related tests on their own test package.

facuMH (dismissed):
This PR adds an integration test that verifies the behaviour of `eth_createAccessList` in regards of the new `bundle.BundleOnly` Access List address and a plan hash.

LuisPH3 (dismissed):
This PR adds regression protection against rpc method create Access List removing bundle-only markers.

PR #671, opened 2026-02-25
Brio: add custom code size limits

simonlechner:
This PR adds the new init and contract code size limit which are enforced starting with the brio hardfork. The sfvm config is set up accordingly.

Due to the enforcement of the limits in the transaction processor a [change](https://github.com/0xsoniclabs/go-ethereum/pull/13) in the geth dependency is required.

### DIFF
--- a/evmcore/tx_validation_test.go
+++ b/evmcore/tx_validation_test.go
@@ -531,27 +531,27 @@ func TestValidateTxForNetwork_AcceptsTransactions(t *testing.T) {
 	}
 }
 
-func TestValidateTxForNetwork_SonicCodeSizeLimitStartingFromOsaka(t *testing.T) {
+func TestValidateTxForNetwork_CustomSonicCodeSizeLimitIsEnforced(t *testing.T) {
 	tests := map[string]struct {
-		osakaEnabled bool
+		customSize   bool
 		initCodeSize uint64
 		errorMessage string
 	}{
 		"Allegro below limit": {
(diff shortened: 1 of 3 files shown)


HerbertJordan (dismissed):

evmcore/tx_pool.go
```diff
@@ -711,7 +711,8 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
 		eip7623:  pool.eip7623,
 		eip7702:  pool.eip7702,
 
-		gasSubsidies: pool.chain.CurrentRules().Upgrades.GasSubsidies,
+		gasSubsidies:            pool.chain.CurrentRules().Upgrades.GasSubsidies,
+		customInitCodeSizeLimit: pool.chain.CurrentRules().Upgrades.Brio,
```
facuMH: this is already covered by the `osaka` feature flag. Perhaps it should be documented somewhere that from osaka onwards in Sonic there is a new init code size limit. But this flag and `osaka` are redundant.

LuisPH3: We should differentiate brio from osaka.
Everything which is 1:1 osaka should be guarded via osaka flags. This facilitates future updates of this code based on changes in geth. (porting new checks they may implement) I would avoid cleaning or changing this code too much, to facilitate this kind of code migrations. 

Anything which is only implemented in Sonic should be guarded by Brio or dedicated flags. This is one of these features.

PR #695, opened 2026-03-05
Add RPC entry points to submit bundles to the network

LuisPH3:
This PR adds the methods `sonic_prepareBundle` and `sonic_submitBundle`, which facilitate the creation of bundles and entirely abstracts the need to deal with the coordinator (bundler) account. 

- `sonic_prepareBundle` takes care of binding all the transactions together, users shall sign them as returned from this function.
- `sonic_submitBundle` queues transactions for execution within a bundle. This function renders unnecessary the need for an extra party signing these transactions.


ethapi/bundle_api.go
```diff
@@ -84,3 +94,204 @@ type BundleInfo struct {
 	Block    *uint64      `json:"block,omitempty"`
 	Position *uint32      `json:"position,omitempty"`
 }
+
+// BundleArgs encapsulates the parameters required for constructing a transaction bundle.
+// It contains the prepared transactions, payment details, gas requirements, and the execution plan
+// necessary for the bundle's integrity and proper execution.
+type BundleArgs struct {
```
HerbertJordan: As a type named "<something>Args" I intuitively thought that this is the input of an RPC function. However, as I figured out further down, this is actually the result. How about renaming this to `BundleDraft` or `BundleProposal` ?

ethapi/bundle_api.go
```diff
@@ -84,3 +94,204 @@ type BundleInfo struct {
...
+// It accepts a list of unsigned transactions, constructs the corresponding execution plan,
+// and updates each transaction to include the bundler-only marker, ensuring they are executed
+// exclusively as part of the specified plan.
+//
+// The returned transactions must be signed without altering any fields; any modification may
+// invalidate the execution plan.
+func (a *PublicBundleAPI) PrepareBundle(
+	ctx context.Context,
+	transactionArgs []TransactionArgs,
+	executionFlags uint8,
```
HerbertJordan: For user convenience, I think those flags should be made explicit, with defaults. So something like
```suggestion
	tolerateInvalid bool,  // false by default
	tolerateFailed bool,   // false by default
```

ethapi/bundle_api.go
```diff
@@ -84,3 +94,204 @@ type BundleInfo struct {
...
 	Position *uint32      `json:"position,omitempty"`
 }
+
+// BundleArgs encapsulates the parameters required for constructing a transaction bundle.
+// It contains the prepared transactions, payment details, gas requirements, and the execution plan
+// necessary for the bundle's integrity and proper execution.
+type BundleArgs struct {
+	// Transactions specifies the ordered list of transactions to be included in the bundle.
+	// These must be signed exactly as provided by the bundle_prepare RPC method; any modification
+	// may invalidate the execution plan and result in an ill-formed bundle.
```
facuMH:
```suggestion
	// will invalidate the execution plan and result in an ill-formed bundle.
```

HerbertJordan (changes requested):

facuMH (dismissed):
Just some superficial comments. Otherwise looks good.

ethapi/bundle_api.go
```diff
@@ -84,3 +94,204 @@ type BundleInfo struct {
...
+	// generate a problem with nonces in the pool.
+	key, err := crypto.GenerateKey()
+	if err != nil {
+		return common.Hash{}, fmt.Errorf("failed to generate signing key: %w", err)
+	}
+
+	// 4) Sign the bundle transaction with the one-use key and send it to the network
+	signer := types.LatestSignerForChainID(a.b.ChainID())
+	tx, err := types.SignNewTx(key, signer,
+		&types.DynamicFeeTx{
```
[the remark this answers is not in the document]
LuisPH3: no good reason, it should be a legacy tx

ethapi/bundle_api.go
```diff
@@ -84,3 +94,204 @@ type BundleInfo struct {
...
+// SubmitBundle implements the `sonic_submitBundle` RPC method, which allows users to submit a prepared transaction bundle for execution.
+// This function accepts a list of transactions that have been signed using the transaction arguments returned by the `sonic_prepareBundle` method,
+// along with execution parameters such as flags and block range.
+// It validates the transactions against the execution plan and submits the bundle to the network for execution.
+func (a *PublicBundleAPI) SubmitBundle(
+	ctx context.Context,
+	rawTransactions []hexutil.Bytes,
+	executionFlags uint8,
+	earliestBlock hexutil.Uint64,
+	latestBlock hexutil.Uint64,
```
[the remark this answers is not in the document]
HerbertJordan: Forwarding the plan would be more robust, and thus developer friendly. You get the plan from function `PrepareBundle` and forward it to `SubmitBundle`. If we extend/modify the plan, the code forwarding it does not need to be touched any more.

The fact that a bit more data is passed is probably not a big deal. But if you think this is critical, the plan could be split into steps and meta-data, where only the meta-data needs to be forwarded.

ethapi/bundle_api.go
```diff
@@ -86,3 +96,217 @@ type BundleInfo struct {
...
+
+type PrepareBundleArgs struct {
+	// Transactions specifies the ordered list of transactions to be included in the bundle.
+	Transactions []TransactionArgs `json:"transactions"`
+	// EarliestBlock specifies the earliest block number at which the bundle can be executed. This allows
+	// users to set a lower bound on when their bundle should be considered for execution, ensuring it is
+	// not included in blocks before a certain point in time.
+	EarliestBlock rpc.BlockNumber `json:"earliestBlock"`
+	// LatestBlock specifies the latest block number at which the bundle can be executed. This allows users
+	// to set an upper bound on when their bundle should be considered for execution, ensuring it is not included in blocks after a certain point in time. If the bundle is not executed by this block, it will be considered expired and will not be executed.
```
facuMH: make this a multi-line comment.

HerbertJordan (dismissed):

facuMH (dismissed):

facuMH (approved):

HerbertJordan (approved):

PR #721, opened 2026-03-12
Add errors to `getBundleStatus` [prototype]

facuMH:
Since bundles are submitted with the usual `SendTransaction`, when a bundles is rejected, a comprehensive message should be returned so that the user is informed why is the bundle rejected.

This PR adds error messages to `getBundleStatus`, `checkForNonceConflicts` and `validateBundleTransactionsInternal`.

This PR renames the `permanently blocked` error to `non executables`.

### DIFF
--- a/evmcore/bundle_precheck.go
+++ b/evmcore/bundle_precheck.go
@@ -17,7 +17,6 @@
 package evmcore
 
 import (
-	"errors"
 	"fmt"
 	"maps"
 	"math"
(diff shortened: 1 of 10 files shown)


LuisPH3 (changes requested):
The presented changes use a not recommended programming practice, where both an error and a value are returned. 
Users of this interface will ignore the value in presence of the error (as it is done in this code already, twice). 
Please rewrite this in a more go-compliant fashion.

evmcore/tx_pool.go
```diff
@@ -115,8 +115,11 @@ var (
 	// ErrBundleTransactionInvalid is returned when a bundle transaction is invalid.
 	ErrBundleTransactionInvalid = errors.New("invalid bundle transaction")
 
-	// ErrBundlePermanentlyBlocked is returned when a bundle is determined to be permanently blocked.
-	ErrBundlePermanentlyBlocked = errors.New("bundle is permanently blocked")
+	// ErrBundleInexecutable is returned when a bundle is determined to be currently not executable in the current state.
+	ErrBundleNonExecutable = errors.New("bundle cannot execute")
```
LorenzSchueler: nit:
```suggestion
	ErrBundleNonExecutable = errors.New("bundle is not executable")
```

LorenzSchueler (dismissed):

evmcore/bundle_precheck.go
```diff
@@ -40,30 +41,31 @@ import (
...
 	// BundleStateTemporaryBlocked indicates that the bundle is not currently
 	// executable due to temporary conditions, but it may become executable in
 	// the future as the blockchain progresses.
 	BundleStateTemporaryBlocked
-	// BundleStatePermanentlyBlocked indicates that the bundle is not executable
-	// and will never be executable in the future due to permanent conditions.
-	BundleStatePermanentlyBlocked
 )
 
 // GetBundleState determines the state of the bundle based on the current state
```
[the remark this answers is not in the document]
facuMH: when a function returns `(value, error)`, in the case of the function returning an error, the value should accompanying the error should be the zero value of the type. For that I prefer the zero value to be "non executable" rather than "running" and someone getting a "Runnable" state because they forgot to check the error.

LorenzSchueler (dismissed):

LuisPH3 (commented):
This PR adds error reporting to the `GetBundleState`.
Everything here seems wrong to me: 
- The function name should not be state, but status or something else (evaluateExecution? ).
- it is very odd that  `GetBundleState` returns an error, specially when non-executable is a valid state. This oddness is clear when errors are used in logic to decide if anything is not executable, but the content of the error is ignored. This is a code smell.  (why would an error be constructed if it was not going to be used? ) 

What does this PR get right: 
- if the function returns an error, other values should be the zero value. 
- the explanations of why something is non-executable seem correct. 
- using errors join facilitates testing.

I am honestly not sure if this PR is mergeable. If the time allows, I would like that you explore alternatives. This PR is purely constrained by golang expressivity. 
Alternatives could be more verbose and complicated, this code may still be best we have. 

@HerbertJordan wdyt?

LorenzSchueler (dismissed):

LuisPH3 (dismissed):

evmcore/bundle_precheck.go
```diff
@@ -119,9 +130,12 @@ func getBundleState(
 	}()
 
 	if success := trialRunner(envelop, chain, stateDb); !success {
-		return BundleStatePermanentlyBlocked
+		return BundleState{
+			Executable: false,
+			Reasons:    []string{"bundle trial-run failed. Revise transactions in the plan"},
```
HerbertJordan:
```suggestion
			Reasons:    []string{"bundle trial-run failed"},
```

LuisPH3 (commented):
I like the semantics now better. We have discussed the meaning of the two booleans, I believe this interpretation is easier to understand.

LuisPH3 (dismissed):

LorenzSchueler (dismissed):

LorenzSchueler (dismissed):

HerbertJordan (changes requested):

HerbertJordan (approved):

PR #743, opened 2026-03-18
Add processed bundle plan store [main]

facuMH:
This PR introduces the storage and management logic for processed bundles in the gossip store.

These new mechanism will be used to prevent re-processing of bundles by tracking execution plan hashes, block numbers, and positions. In order to enable that, the store maintains a history hash for cross-validator alignment of the last 1024 blocks. 

The implementation includes robust error handling with critical logging, batch operations for atomic updates, and utility functions for key management. 

Comprehensive unit tests cover both normal and error scenarios, including batch and iterator failures, invalid data, and clean-up logic.


gossip/store_processed_bundles.go
```diff
@@ -0,0 +1,242 @@
...
+// cross-validate that validators remain aligned on their bundle processing
+// history.
+//
+// Bundles need to be indexed by the execution plan hash instead of the hash of
+// the bundled transactions enclosing them, since otherwise the same plan may be
+// resubmitted multiple times using different envelop transactions.
+//
+// In the underlying table, the following keys are used:
+//  - key: [] -> [uint64, hash]                        // last block and hash for which the processed bundles have been stored
+//  - key: ['e']<execPlanHash> -> [block,position,count]     // for a processed bundle
```
HerbertJordan:
```suggestion
//  - key: ['e']<execPlanHash> -> [block,position,count]     // for a recently processed bundle (auto-pruned)
```

gossip/store_processed_bundles_test.go
```diff
@@ -0,0 +1,527 @@
...
+
+	blockNum := uint64(123)
+	hash := common.Hash{1, 2, 3}
+	expectedKey := append([]byte{'i'}, make([]byte, 8)...)
+	binary.BigEndian.PutUint64(expectedKey[1:9], blockNum)
+	expectedKey = append(expectedKey, hash.Bytes()...)
+	require.Equal(expectedKey, getIndexKey(blockNum, hash))
+}
+
+func TestStore_AddProcessedBundles_LogsOnBatchPutError(t *testing.T) {
```
LuisPH3: nit: `t.Parallel()`
(in all new tests as well)

LuisPH3 (changes requested):
The Changes introduce the bundles_store in main. 
Testing has been added on top of the testing existing in the prototype, the code has good coverage.

gossip/store_processed_bundles_test.go
HerbertJordan: Please add unit tests for the `xorHash` function.

HerbertJordan (commented):
A few initial thoughts, but I still need to look through the tests in detail.

gossip/store_processed_bundles_test.go
```diff
@@ -0,0 +1,527 @@
...
+	require.True(isRecentlyProcessed(hash4))
+
+	store.AddProcessedBundles(1+2*bundle.MaxBlockRange,
+		[]bundle.ExecutionInfo{})
+
+	require.False(isRecentlyProcessed(hash1))
+	require.False(isRecentlyProcessed(hash2))
+	require.False(isRecentlyProcessed(hash3))
+	require.False(isRecentlyProcessed(hash4))
+
```
HerbertJordan: With the cleanup we have to be very careful regarding off-by-one errors. We must be sure to remove the correct set of bundles such that non get removed before `bundle.MaxBlockRange` is expired. Can you add tests for that, probing the boundary?

HerbertJordan (changes requested):

gossip/store_processed_bundles_test.go
facuMH: you mean like `TestStore_ProcessedBundles_StoredHashUsesXorForAddedAndDeletedHashes` ?

LuisPH3 (dismissed):

LuisPH3 (dismissed):

gossip/store_processed_bundles_test.go
```diff
@@ -0,0 +1,804 @@
...
+			hash := store.deleteOutdatedBundles(c.currentBlockNumber, batch)
+			if c.expectDeleted {
+				require.Equal(t, existingBundleHash, xorHash(common.Hash{}, hash))
+			}
+		})
+	}
+}
+
+func TestStore_deleteOutdatedBundles_ReturnsXorHashOfDeletedEntries(t *testing.T) {
```
HerbertJordan: Should also cover a few more edge cases, like an empty list, a single entry, and more than two entries.

HerbertJordan (changes requested):

LuisPH3 (approved):
I believe that the problem that happen with this PR was that both the original implementation and prototype testing were taken for valid. 
The perceived scope was to add new tests. The correct approach would have been to rewrite every test with the same style and clear intention to facilitate reviewer analysis of coverage. 
For the next time, migrate the function prototypes, start implementing and testing the code. Take proto code as a guideline, not as final. 

Example:
We do not write tests like `TestStore_HasBundleRecentlyBeenProcessed_CleansUpOldBundleHashes` with imperative and stateful approach. We prefer table driven declarative tests. The test was clearly a prototype.

gossip/store_processed_bundles_test.go
```diff
@@ -0,0 +1,998 @@
...
+
+func TestStore_GetIndexKey_ReturnsExpectedKey(t *testing.T) {
+	require := require.New(t)
+
+	blockNum := uint64(123)
+	hash := common.Hash{1, 2, 3}
+	expectedKey := append([]byte{'i'}, make([]byte, 8)...)
+	binary.BigEndian.PutUint64(expectedKey[1:9], blockNum)
+	expectedKey = append(expectedKey, hash.Bytes()...)
+	require.Equal(expectedKey, getIndexKey(blockNum, hash))
```
LuisPH3: Add test checking that lengh is correct. 
(the test initialized a 3 bytes test, would pass if something changes in the hash type)

LuisPH3 (changes requested):

PR #751, opened 2026-03-20
Let prepare fill in bundle block range if not specified.

LuisPH3:
This PR extends sonic_prepareBundle to fill bundles range parameters in left not specified.


facuMH (dismissed):
This PR adds the capability to assign valid default block range for the bundle being prepared to the RPC call `PreparedBundle`

HerbertJordan (dismissed):

facuMH (approved):

PR #757, opened 2026-03-23
Add calculation of default tx parameters in prepareBundle

LuisPH3:
This PR adds the option for letting `sonic_prepareBundle` infer gas costs and gas prices to be used in the bundled transactions. 
This feature makes difference between initialized transaction arguments and zeroes. Zeroes are interpreted as user choice and therefore not overwritten. 

Access list is not initialized by default because it requires a slightly larger scope. It is not strictly necessary to contribute to the success of the execution either.

### DIFF
--- a/ethapi/bundle_api.go
+++ b/ethapi/bundle_api.go
@@ -138,7 +138,6 @@ type RPCPreparedBundle struct {
 // Bundled transactions with uninitialized gas limits will have their gas estimated by this method, which will take into account
 // potential state changes from previous transactions in the bundle. However, users can also choose to set gas limits on their own;
 // in this case, the provided gas limits will be used without modification.
-// Transactions with gas limit 0 will not have this field modified.
 //
 // Bundled transactions with uninitialized gas price fields (GasPrice for access list transactions,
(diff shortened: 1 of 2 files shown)


HerbertJordan (dismissed):

ethapi/bundle_api.go
```diff
@@ -134,6 +135,17 @@ type RPCPreparedBundle struct {
 // and updates each transaction to include the bundler-only marker, ensuring they are executed
 // exclusively as part of the specified plan.
 //
+// Bundled transactions with uninitialized gas limits will have their gas estimated by this method, which will take into account
+// potential state changes from previous transactions in the bundle. However, users can also choose to set gas limits on their own;
+// in this case, the provided gas limits will be used without modification.
+// Transactions with gas limit 0 will not have this field modified.
```
simonlechner: The following code has different behavior.

simonlechner (dismissed):
Looks good except for wrong documentation.

HerbertJordan (approved):

simonlechner (approved):

PR #758, opened 2026-03-23
Concept: get bundle info only if executed. 

LuisPH3:
This PR is a little large and can be separated in several PRs. 
Nevertheless, I present here an alternative to the current monitoring of bundles.


tests/contracts/revert/revert.sol
HerbertJordan: This change simplifies the scheduling problem to be solved by the network, making the difficult case more manageable in tests, but the original problem remains.

The original code was specifically designed to provide a case where only during execution it can be decided whether a revert happens or not -- since the revert depended on the ordering of calls in the block, and non-reverts and reverts may be arbitrarily mixed. A reverted transaction may be followed by a non-reverted, followed by a reverted.

The new version, as it is used in the test, just performs one transition - from a never-revert to an always revert (since the toggle is only triggered once). So initially all bundles are accepted, and at some point all bundles are dropped. This simplifies the setup of the test, but also circumvents the original problem demonstrated by the previous version.

LuisPH3: yes, this solution circumvents the original problem because the problem is not unique to bundles but to transactions as well. 
The original concept attempted to solve a problem which is deep inside the lachesis protocol, and completely out of scope of the bundles problem. The implementation presented here provides the same guarantees as the sonic network provides for any normal transaction.

PR #762, opened 2026-03-24
Remove bundle status from getBundleInfo

LuisPH3:
Bundle status was an enum meant to differentiate a bundle which has been executed from one which is pending for execution. Deciding whenever a bundle can be executed is a non-trivial task and results may not be consistent over time or when querying different RPCs in the network. 
This PR removes this feature with the aim of simplifying the API, and opens the possibility to query pending bundles via other RPC methods. 

The changes in this PR introduce a workflow analogous to waiting for any transaction, by calling `eth_getTransactionReceipt`, where the answer is only available after execution. 

The function `sonic_getBundleInfo` will return the result of evaluating a bundle only after the block attempting the bundle is completed. Any query in between will return no error with null info.


HerbertJordan (dismissed):
Looks good to me, feel free to ignore the comments.

facuMH (dismissed):

HerbertJordan (dismissed):

HerbertJordan (approved):

PR #772, opened 2026-03-27
Refactor RCP API structure

LuisPH3:
This PR breaks the cycle that prevents adding new API namespaces in different packages than ethapi. This reduces decoupling in an extremely large namespace with extremely large files.  

The purpose is to have bundles RPC developed in their own package. 
The PR hints the location where further development can start. 
Additionally, this change will enable the migration of RPC methods into a new architecture one at the time.


HerbertJordan (dismissed):

jenikd (approved):
Looks good

PR #778, opened 2026-03-30
Do not validate bundles in state processor

LorenzSchueler:
This PR removes the validation of bundles from the state processor. The validation happens in the `c_block_callsback.go` [here](https://github.com/0xsoniclabs/sonic/blob/4cbe70042f038c7be8141a187f12e6fbcb66f7ae/gossip/c_block_callbacks.go#L949-L1051).
This brings it in line with `main`.


HerbertJordan (commented):
I think this PR should not be merged. See comment.

I would suggest to update the main branch to do the validation and block range check as well.

PR #799, opened 2026-04-01
Separating backend interface to avoid dependency cycle

jenikd:
This PR explores the way to separate dependencies to avoid dependency cycle for future tests


facuMH (dismissed):

LuisPH3 (changes requested):
As discussed offline, this PR can merged after we find a better naming for the package and type. This PR does not block development of bundle_api or testing RPC framework

PR #803, opened 2026-04-02
Refactor validator flags to use brio boolean

LuisPH3:
The reason behind this change is that code sizes can never be enabled or disabled as a feature, once they are used once (contract overflowing will remain deployed). 

The brio flag can be reused to validate bundles (before brio they behave as normal txs).


facuMH (dismissed):
This PR renames the `TxPool` rules flag `customInitCodeSizeLimit` to `brio`. I strongly agree with this kind of renames and cleanups.

simonlechner (commented):
I don't see the added test coverage by the new unit test compared to `TestValidateTxForNetwork_CustomSonicCodeSizeLimitIsEnforced` test.

simonlechner (approved):

PR #805, opened 2026-04-03
Introduce cached checker utility

LuisPH3:
This PR introduces a utility to cache expensive transaction checks done in the transaction pool. The purpose is to reuse the same heuristics developed in https://github.com/0xsoniclabs/sonic/pull/543 for the transaction bundles evaluations done during pool reorg. Benefits of doing catching bundles evaluations were explored in https://github.com/0xsoniclabs/sonic/pull/791

The PR is organized in two commits, the first introduces the tool, the second migrates use in the pool to the new infrastructure.


facuMH (dismissed):
This PR introduces a new LRU generic utility, and uses it in the txpool to store the subsidies and prevent duplicated expensive computations.
Testing of the new utility cover the necessary properties.

facuMH (dismissed):

evmcore/tx_pool.go
simonlechner: Nit: also rename the factory to `subsidiesCheckerFuncFactory`.

simonlechner (dismissed):
Feel free to ignore the picky command.

evmcore/subsidies_integration.go
```diff
@@ -20,23 +20,15 @@ import (
...
-// on state for an operation which is pure.
-//
-// This interface facilitates testing and decouples the subsidies integration
-// logic from the transaction pool.
-type subsidiesChecker interface {
-	isSponsored(tx *types.Transaction) bool
-}
+// IsSponsoredCheckFunc defines the function signature
+// for checking if a sponsorship request is funded.
+type IsSponsoredCheckFunc = utils.CheckFunc[*types.Transaction, bool]
```
HerbertJordan: Nit: type type name references a single use case, but the type is more generic. The same type might also be used for bundles or other transaction checks. How about renaming this to `TransactionCheck`?

It also looks like it would be fine to make it private.

HerbertJordan (changes requested):

utils/checker_cache.go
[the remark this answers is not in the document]
LuisPH3: naming should not be an issue now,  tried to document the timing scheme in the type documentation.

HerbertJordan (dismissed):

facuMH (approved):

PR #819, opened 2026-04-08
Add Store utilities for bundles import/export.

facuMH:
This PR adds 3 methods related to bundles to the `gossip.Store`.

**Main Contributions:**
- `readProcessedBundleHistoryEntry` an internal method which fetches the entry with the empty key.
- `SetProcessedBundlesHistoryHash` an exported method to enable the genesis import to write the processed bundle history hash. 
- `EnumerateProcessedBundles` an exported method which returns a slice containing the `bundle.ExecutionInfo` corresponding to all the entries in the `ProcessedBundles` table.

NOTE: This PR has been simplified from its original ideas after an offline talk with @LuisPH3. Encoding/Decoding will be added in follow-up unrelated to the methods in the store.

### DIFF
--- a/gossip/store_processed_bundles.go
+++ b/gossip/store_processed_bundles.go
@@ -339,7 +339,15 @@ func (s *Store) DumpProcessedBundles() [][]byte {
 	it := s.table.ProcessedBundles.NewIterator([]byte{'e'}, nil)
 	defer it.Release()
 	for it.Next() {
-		entry := BundleKV{Key: it.Key(), Value: it.Value()}
+		key := it.Key()
+		value := it.Value()
+		if len(key) != 1+32 || len(value) != 16 {
+			s.Log.Crit(
+				"invalid key or value length for processed bundle entry during dump",
+				"keyLength", len(key),
+				"valueLength", len(value))
+		}
+		entry := BundleKV{Key: key, Value: value}
 		dump = append(dump, entry.Encode())
 	}
 	if it.Error() != nil {
--- a/gossip/store_processed_bundles.go
+++ b/gossip/store_processed_bundles.go
@@ -19,9 +19,9 @@ package gossip
 import (
 	"encoding/binary"
 	"errors"
-	"fmt"
 
 	"github.com/0xsoniclabs/sonic/gossip/blockproc/bundle"
+	"github.com/Fantom-foundation/lachesis-base/common/bigendian"
 	"github.com/Fantom-foundation/lachesis-base/kvdb"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
@@ -80,7 +80,7 @@ func (s *Store) AddProcessedBundles(
 	newHash := computeNewBundleStateHash(oldHash, addedHash, deletedHash, blockNum)
 
 	err := batch.Put(nil, append(
-		uint64ToBytes(blockNum),
+		bigendian.Uint64ToBytes(blockNum),
 		newHash.Bytes()...,
 	))
 	if err != nil {
@@ -233,7 +233,7 @@ func (s *Store) GetBundleExecutionInfo(execPlanHash common.Hash) *bundle.Executi
 // GetProcessedBundleHistoryHash returns the block number of the last update
 // and the current hash of the processed bundles history.
 func (s *Store) GetProcessedBundleHistoryHash() (uint64, common.Hash) {
-	state := s.processedBundleHistoryEntry()
+	state := s.readProcessedBundleHistoryEntry()
 	if state == nil {
 		return 0, common.Hash{}
 	}
@@ -242,12 +242,12 @@ func (s *Store) GetProcessedBundleHistoryHash() (uint64, common.Hash) {
 	return blockNum, hash
 }
 
-// processedBundleHistoryEntry returns the raw value of the entry that stores the
+// readProcessedBundleHistoryEntry returns the raw value of the entry that stores the
 // block number and hash of the processed bundles history, or nil if it is not
 // set. The returned value is expected to be 40 bytes long, with the first 8
 // bytes representing the block number in big-endian format, and the next 32
 // bytes representing the hash of the processed bundles history.
-func (s *Store) processedBundleHistoryEntry() []byte {
+func (s *Store) readProcessedBundleHistoryEntry() []byte {
 	state, err := s.table.ProcessedBundles.Get(nil)
 	if err != nil {
 		s.Log.Crit("failed to get hash of processed bundles", "error", err)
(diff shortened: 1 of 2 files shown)


gossip/store_processed_bundles.go
```diff
@@ -245,9 +262,117 @@ func (s *Store) GetProcessedBundleHistoryHash() (uint64, common.Hash) {
...
+	}
+
+	batch := s.table.ProcessedBundles.NewBatch()
+
+	// write entry key
+	if err := batch.Put(kv.Key, kv.Value); err != nil {
+		return err
+	}
+	// write index key
+	indexKey := getIndexKey(binary.BigEndian.Uint64(kv.Value[:8]), execPlanHash)
```
HerbertJordan: Please give the data you are parsing here a name, something like

```suggestion
    blockNumber := binary.BigEndian.Uint64(kv.Value[:8])
	indexKey := getIndexKey(blockNumber, execPlanHash)
```

gossip/store_processed_bundles.go
```diff
@@ -245,9 +262,117 @@ func (s *Store) GetProcessedBundleHistoryHash() (uint64, common.Hash) {
...
+	if currentHistoryHash.Value == nil {
+		return [][]byte{}
+	}
+	dump = append(dump, currentHistoryHash.Encode())
+
+	// get all recently processed bundles
+	it := s.table.ProcessedBundles.NewIterator([]byte{'e'}, nil)
+	defer it.Release()
+	for it.Next() {
+		entry := BundleKV{Key: it.Key(), Value: it.Value()}
```
HerbertJordan: nit: In other places in this file we do not necessarily trust that there are no entries in the key-value store with a different formatting. If this "tradition" should be retained, the length of the key and value should be checked here.

gossip/store_processed_bundles.go
```diff
@@ -272,3 +397,10 @@ func xorHash(a, b common.Hash) common.Hash {
 	}
 	return res
 }
+
+// uint32ToBytes converts a uint32 to a byte slice in big-endian order.
+func uint32ToBytes(v uint32) []byte {
+	b := make([]byte, 4)
+	binary.BigEndian.PutUint32(b, v)
+	return b
```
HerbertJordan: nit: This should be equivalent to
```suggestion
	return binary.BigEndian.AppendUint32(nil, v)
```
which could be inlined as well.

gossip/store_processed_bundles.go
HerbertJordan, line 1: Two remarks on the API design:
- it seems odd that the exported data format is `[]byte` while the import data type is `BundleKV`. Would be good to pick one of those for both functions
- while you are exporting a list of entries, the import only accepts one entry at a time. How about exporting and importing lists of entries?

gossip/store_processed_bundles.go
```diff
@@ -128,6 +129,7 @@ func (s *Store) deleteOutdatedBundles(blockNum uint64, batch kvdb.Batch) common.
 		// enough blocks have passed to start cleaning up the store
 		highestOutdatedBlockNumber := blockNum - bundle.MaxBlockRange + 1
 		it := s.table.ProcessedBundles.NewIterator([]byte{'i'}, nil)
+		defer it.Release()
```
HerbertJordan, line 132: Nice find!

gossip/store_processed_bundles.go
```diff
@@ -245,9 +262,117 @@ func (s *Store) GetProcessedBundleHistoryHash() (uint64, common.Hash) {
...
+		return err
+	}
+
+	return batch.Write()
+}
+
+// DumpProcessedBundles returns a dump of all the entries in the processed
+// bundles table in bytes. The dump includes both the history entry and the
+// entries for recently processed bundles.
+func (s *Store) DumpProcessedBundles() [][]byte {
```
HerbertJordan: How about renaming those functions to "ExportProcessedBundles" and "ImportProcessedBundles"?

gossip/store_processed_bundles_test.go
```diff
@@ -934,6 +934,319 @@ func TestStore_ProcessedBundles_RetainsAllBundlesRequiredToCoverTheMaximumBlockR
...
+			key:      []byte{0, 1, 2},  // should be 33 bytes for 'e' + hash
+			value:    make([]byte, 16), // correct length for value
+			errorMsg: "invalid key or value for processed bundle entry",
+		},
+		"entry invalid value length": {
+			key:      append([]byte{'e'}, make([]byte, 32)...), // valid length key
+			value:    []byte{0, 1, 2},                          // short value
+			errorMsg: "invalid key or value for processed bundle entry",
+		},
+		"empty execution plan": {
```
HerbertJordan:
```suggestion
		"zero execution plan": {
```

gossip/store_processed_bundles_test.go
```diff
@@ -934,6 +934,319 @@ func TestStore_ProcessedBundles_RetainsAllBundlesRequiredToCoverTheMaximumBlockR
...
+			key:      append([]byte{'e'}, make([]byte, 32)...), // valid length key
+			value:    make([]byte, 16),                         // valid length but empty value
+			errorMsg: "invalid execution plan hash",            // we require the hash to be non-zero to avoid confusion with empty entries, but the error message is the same as for invalid lengths since we check length first
```
HerbertJordan: the comment seems to be wrong ... the error message is not the same as in another case.

gossip/store_processed_bundles.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/819/changes/ccf45e0ebffc9abca9134fae41d39a25e52a3fb2

gossip/store_processed_bundles.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/819/changes/7268fa3d2ad300aef45d2660e295483691c08c6f

gossip/store_processed_bundles.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/819/changes/5666da67a87186374ca35eb17026974e51287542

gossip/store_processed_bundles.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/819/changes/b4ed6a0858faadc54bec7f9f5b2ef2829dec9da1

gossip/store_processed_bundles_test.go
facuMH: done as part of https://github.com/0xsoniclabs/sonic/pull/819/changes/5666da67a87186374ca35eb17026974e51287542

gossip/store_processed_bundles_test.go
facuMH: comment removed.

gossip/store_processed_bundles.go
```diff
@@ -231,12 +233,27 @@ func (s *Store) GetBundleExecutionInfo(execPlanHash common.Hash) *bundle.Executi
...
+	hash := common.BytesToHash(state[8:])
+	return blockNum, hash
+}
+
+// processedBundleHistoryEntry returns the raw value of the entry that stores the
+// block number and hash of the processed bundles history, or nil if it is not
+// set. The returned value is expected to be 40 bytes long, with the first 8
+// bytes representing the block number in big-endian format, and the next 32
+// bytes representing the hash of the processed bundles history.
+func (s *Store) processedBundleHistoryEntry() []byte {
```
LuisPH3: member misses verb.

gossip/store_processed_bundles.go
```diff
@@ -245,9 +262,161 @@ func (s *Store) GetProcessedBundleHistoryHash() (uint64, common.Hash) {
...
+		}
+		if err := batch.Write(); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func decodeEntry(offset int, data []byte) ([]byte, []byte, int, error) {
```
LuisPH3: This function needs a better name, and short docstring.
The problem is that this function is in the `gossip` package, therefore almost any utility can call it, and it does not say anything about what it does or where is it really needed.

gossip/store_processed_bundles.go
```diff
@@ -272,3 +441,13 @@ func xorHash(a, b common.Hash) common.Hash {
 	}
 	return res
 }
+
+// uint64ToBytes converts a uint64 to a byte slice in big-endian order.
+func uint64ToBytes(v uint64) []byte {
```
LuisPH3: utilitly already exists: 
` bigendian.Uint64ToBytes(v)`

gossip/store_processed_bundles.go
```diff
@@ -272,3 +441,13 @@ func xorHash(a, b common.Hash) common.Hash {
...
+	return binary.BigEndian.AppendUint64(nil, v)
+}
+
+// uint32ToBytes converts a uint32 to a byte slice in big-endian order.
+func uint32ToBytes(v uint32) []byte {
```
LuisPH3: idem

gossip/store_processed_bundles.go
LuisPH3: if you do not trust lachesisbase, and you want to copy the code, please place it in utils. and test it.

gossip/store_processed_bundles.go
```diff
@@ -245,9 +262,161 @@ func (s *Store) GetProcessedBundleHistoryHash() (uint64, common.Hash) {
...
+	}
+	if it.Error() != nil {
+		s.Log.Crit("failed to export processed bundles", "error", it.Error())
+	}
+	return result
+}
+
+// Encode encodes a key-value pair into a byte slice for storage or export.
+// The format is: [uint32 key length][key][uint32 value length][value]
+func Encode(key, value []byte) []byte {
```
LuisPH3: please improve naming, this function is now `gossip.Encode`, what does it encode and into what?

LuisPH3: does not need to be exported.

LuisPH3 (commented):
I find this PR a little strange, it is a little too out of context because I cannot see how is meant to be used.
Nevertheless, I find a mixture of weak-generic names introduced into the gossip package and I get concerned. 
I do not see encode/decode facilities in other parts of the store, and in the export [command](https://github.com/0xsoniclabs/sonic/blob/332d9d56b862cac7df90345f0b4952d2509ba7fa/cmd/sonictool/genesis/export.go#L168) I see that RLP is already used to serialize blocks. Let's not re-invent the wheel.

gossip/store_processed_bundles.go
facuMH: Implementation refactored.  New methods `SetProcessedBundlesHistoryHash` takes a block number and hash, while `EnumerateProcessedBundles` returns a slice of `bundle.ExecutionInfo`

gossip/store_processed_bundles.go
facuMH: renamed to `readProcessedBundleHistoryEntry`

gossip/store_processed_bundles.go
facuMH: function removed. Decode will be handled differently.

gossip/store_processed_bundles.go
facuMH: function removed. Decode will be handled elsewhere.

PR #825, opened 2026-04-09
Add RPC function EstimateGasForTransactions

jenikd:
This PR introduces new RPC function `EstimateGasForTransactions`. It is the first RPC function related to the transaction bunles API, which allows gas estimation of more than one transaction. 

Some considerations:

- Bundle flags are not supported
- If transactions become invalid or fail during execution before an estimation for a later transaction, they are simply ignored


api/sonicapi/estimate_gas.go
LuisPH3, line 1: I would drop the _api suffix in the file names, wdyt? 
After all these functions are in the sonicapi namespace

api/sonicapi/estimate_gas.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/ethereum/go-ethereum/rpc"
+)
+
+//go:generate mockgen -source=estimate_gas_api.go -destination=estimate_gas_api_mock.go -package=sonicapi
+
+// MAX_BUNDLE_TRANSACTIONS is the maximum number of transactions
+// that can be included in a bundle for gas estimation.
+const MAX_BUNDLE_TRANSACTIONS = 16
+
+type BundleGasLimits struct {
```
facuMH, line 30: document this is supposed to be used as the return type for `EstimateGasForTransactions`

api/sonicapi/estimate_gas_api_test.go
```diff
@@ -0,0 +1,208 @@
...
+		Build()
+
+	api := NewPublicBundleAPI(be)
+
+	args := []ethapi.TransactionArgs{
+		{
+			From:  acc1.Address(),
+			To:    acc2.Address(),
+			Nonce: rpctest.ToHexUint64(0),
+			Value: rpctest.ToHexBigInt(big.NewInt(1e16)),
```
facuMH: I see this is a repeated pattern. Maybe it can be made into a helper function to improve readability. wdyt ?

api/sonicapi/estimate_gas_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+	_, err = api.EstimateGasForTransactions(t.Context(), args, nil, nil, nil)
+	require.ErrorContains(t, err, "too many transactions")
+}
+
+func Test_EstimateGasForTransactions_EmptyArgs(t *testing.T) {
+	be := rpctest.NewBackendBuilder(t).Build()
+	api := NewPublicBundleAPI(be)
+
+	result, err := api.EstimateGasForTransactions(t.Context(), []ethapi.TransactionArgs{}, nil, nil, nil)
```
facuMH, line 133: test not only with empty list, but also with `nil` list.

api/sonicapi/estimate_gas_test.go
```diff
@@ -0,0 +1,208 @@
...
+		},
+	}
+
+	result, err := api.EstimateGasForTransactions(t.Context(), args, &blockNrOrHash, nil, nil)
+	require.NoError(t, err)
+	require.Len(t, result.GasLimits, 1)
+	require.GreaterOrEqual(t, uint64(result.GasLimits[0]), uint64(params.TxGas+bundleGasOverhead))
+}
+
+func Test_EstimateGasForTransactions_WithStateOverride(t *testing.T) {
```
facuMH, line 159: missing test with block overrides. Although since both state and block overrides are forwarded to `eval.EstimateGas(arg, preArgs)` you could just mock it and verify that the overrides are correctly forwarded.

api/sonicapi/estimate_gas_api.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/0xsoniclabs/sonic/api/ethapi"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/ethereum/go-ethereum/rpc"
+)
+
+//go:generate mockgen -source=estimate_gas_api.go -destination=estimate_gas_api_mock.go -package=sonicapi
+
+// MAX_BUNDLE_TRANSACTIONS is the maximum number of transactions
+// that can be included in a bundle for gas estimation.
```
LuisPH3: can we document how did we come to this limit?

api/sonicapi/estimate_gas_api_mock.go
```diff
@@ -0,0 +1,56 @@
...
+	recorder *MockGasEstimatorMockRecorder
+}
+
+// MockGasEstimatorMockRecorder is the mock recorder for MockGasEstimator.
+type MockGasEstimatorMockRecorder struct {
+	mock *MockGasEstimator
+}
+
+// NewMockGasEstimator creates a new mock instance.
+func NewMockGasEstimator(ctrl *gomock.Controller) *MockGasEstimator {
```
LuisPH3: Mock is unused, do we need the interface then? I think it was only introduced to enable testing.
On the other hand the doEstimateGasForTransactions could be unit-tested

I would say that we use it, or we do not have neither of the interface nor mock

api/sonicapi/estimate_gas_api_test.go
```diff
@@ -0,0 +1,208 @@
...
+	result, err := api.EstimateGasForTransactions(t.Context(), args, nil, nil, nil)
+	require.NoError(t, err)
+	require.Len(t, result.GasLimits, 3)
+	for i, gas := range result.GasLimits {
+		require.GreaterOrEqual(t, uint64(gas), uint64(params.TxGas+bundleGasOverhead),
+			"gas limit for tx %d is too low", i)
+	}
+}
+
+func Test_EstimateGasForTransactions_TooManyTransactions(t *testing.T) {
```
LuisPH3:
```suggestion
func Test_EstimateGasForTransactions_TooManyTransactions_RetunrsError(t *testing.T) {
```

api/sonicapi/estimate_gas_api_test.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	result, err := api.EstimateGasForTransactions(t.Context(), args, nil, nil, nil)
+	require.NoError(t, err)
+	require.Len(t, result.GasLimits, 1)
+	// Gas must be at least TxGas + bundle overhead
+	require.GreaterOrEqual(t, uint64(result.GasLimits[0]), uint64(params.TxGas+bundleGasOverhead))
+}
+
+func Test_EstimateGasForTransactions_MultipleIndependentTransfers(t *testing.T) {
```
LuisPH3:
```suggestion
func Test_EstimateGasForTransactions_MultipleIndependentTxs_AreEstimated(t *testing.T) {
```

api/sonicapi/estimate_gas_api_test.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/holiman/uint256"
+	"github.com/stretchr/testify/require"
+)
+
+// bundleGasOverhead is the extra gas added per transaction in the bundle
+// to account for the execution plan entries in the access list.
+const bundleGasOverhead = params.TxAccessListAddressGas + params.TxAccessListStorageKeyGas
+
+func Test_EstimateGasForTransactions_SingleTransfer(t *testing.T) {
```
LuisPH3:
```suggestion
func Test_EstimateGasForTransactions_SingleTx_IsEstimated(t *testing.T) {
```

api/sonicapi/estimate_gas_api_test.go
```diff
@@ -0,0 +1,208 @@
...
+	acc1, err := rpctest.NewWallet()
+	require.NoError(t, err)
+	acc2, err := rpctest.NewWallet()
+	require.NoError(t, err)
+
+	be := rpctest.NewBackendBuilder(t).
+		WithAccount(*acc1.Address(), rpctest.AccountState{Balance: big.NewInt(1e18)}).
+		WithAccount(*acc2.Address(), rpctest.AccountState{}).
```
LuisPH3: why is this acc2 needed?  should not be a problem if it never existed before

api/sonicapi/estimate_gas_api_test.go
```diff
@@ -0,0 +1,208 @@
...
+		Build()
+
+	api := NewPublicBundleAPI(be)
+
+	args := []ethapi.TransactionArgs{
+		{
+			From:  acc1.Address(),
+			To:    acc2.Address(),
```
LuisPH3: no need to create a full wallet, just any address should work

api/sonicapi/estimate_gas_api_test.go
```diff
@@ -0,0 +1,208 @@
...
+			Nonce: rpctest.ToHexUint64(uint64(i)),
+			Value: rpctest.ToHexBigInt(big.NewInt(1)),
+		}
+	}
+
+	_, err = api.EstimateGasForTransactions(t.Context(), args, nil, nil, nil)
+	require.ErrorContains(t, err, "too many transactions")
+}
+
+func Test_EstimateGasForTransactions_EmptyArgs(t *testing.T) {
```
LuisPH3:
```suggestion
func Test_EstimateGasForTransactions_EmptyArgs_ReturnsEmptyEstimationWithNoError(t *testing.T) {
```

api/sonicapi/estimate_gas_api_test.go
```diff
@@ -0,0 +1,208 @@
...
+			From:  acc1.Address(),
+			To:    acc2.Address(),
+			Value: rpctest.ToHexBigInt(big.NewInt(1e16)),
+		},
+	}
+
+	result, err := api.EstimateGasForTransactions(t.Context(), args, nil, &overrides, nil)
+	require.NoError(t, err)
+	require.Len(t, result.GasLimits, 1)
+	require.GreaterOrEqual(t, uint64(result.GasLimits[0]), uint64(params.TxGas+bundleGasOverhead))
```
LuisPH3: Strict equal?

api/sonicapi/estimate_gas_api_test.go
```diff
@@ -0,0 +1,208 @@
...
+			From:  acc1.Address(),
+			To:    acc2.Address(),
+			Value: rpctest.ToHexBigInt(big.NewInt(1e16)),
+		},
+	}
+
+	result, err := api.EstimateGasForTransactions(t.Context(), args, &blockNrOrHash, nil, nil)
+	require.NoError(t, err)
+	require.Len(t, result.GasLimits, 1)
+	require.GreaterOrEqual(t, uint64(result.GasLimits[0]), uint64(params.TxGas+bundleGasOverhead))
```
LuisPH3: Strict equal?

api/sonicapi/estimate_gas_test.go
```diff
@@ -0,0 +1,208 @@
...
+func Test_EstimateGasForTransactions_EmptyArgs(t *testing.T) {
+	be := rpctest.NewBackendBuilder(t).Build()
+	api := NewPublicBundleAPI(be)
+
+	result, err := api.EstimateGasForTransactions(t.Context(), []ethapi.TransactionArgs{}, nil, nil, nil)
+	require.NoError(t, err)
+	require.Empty(t, result.GasLimits)
+}
+
+func Test_EstimateGasForTransactions_WithExplicitBlockNumber(t *testing.T) {
```
LuisPH3, line 138:
```suggestion
func Test_EstimateGasForTransactions_WithExplicitBlockNumber_Does....(t *testing.T) {
```
not sure what shall we expect of this test.

This test seems well-intentioned but it falls short. What could change in a gas estimation when using a different block?  I guess we could use gas cost changes between different rule sets to track the execution at a given block height. 
The problem is that we do not have tooling in the testing framework to change the rules in different block heights. We could do the following:

either one of: 
- `rpctest.Block` has a rules
-  or we have a setter in the builder where we can define rule changes at block heights. (maybe this is better, something like a `map[idx.Block]opera.Rules`) 

`GetNetworkRules` can use this info to return different rules at different block heights. 


Using a transaction with large data should then yield a different estimation in allegro or brio.

api/sonicapi/estimate_gas_api_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+func Test_EstimateGasForTransactions_WithStateOverride(t *testing.T) {
+	acc1, err := rpctest.NewWallet()
+	require.NoError(t, err)
+	acc2, err := rpctest.NewWallet()
+	require.NoError(t, err)
+
+	// acc1 starts with no balance, but a state override will fund it
+	be := rpctest.NewBackendBuilder(t).
+		WithAccount(*acc1.Address(), rpctest.AccountState{}).
```
LuisPH3: should not be needed to initialize the account, or does it?

api/sonicapi/estimate_gas_test.go
```diff
@@ -0,0 +1,208 @@
...
+			To:    acc2.Address(),
+			Value: rpctest.ToHexBigInt(big.NewInt(1e16)),
+		},
+	}
+
+	result, err := api.EstimateGasForTransactions(t.Context(), args, nil, &overrides, nil)
+	require.NoError(t, err)
+	require.Len(t, result.GasLimits, 1)
+	require.GreaterOrEqual(t, uint64(result.GasLimits[0]), uint64(params.TxGas+bundleGasOverhead))
+}
```
LuisPH3, line 196: The documentation of the rpc method describes behavior of skipping transactions when fail or skipped. This calls for a test case.

LuisPH3 (changes requested):
The new framework seems quite capable. 
This is a good example of tests which previously were pretty tedious. 
Some comments:

api/sonicapi/estimate_gas.go
```diff
@@ -0,0 +1,127 @@
...
+	// bundle, in the same order as the input transactions.
+	GasLimits []hexutil.Uint64 `json:"gasLimits"`
+}
+
+// EstimateGasForTransactions implements the `sonic_estimateGasForTransactions` RPC method.
+// It estimates the gas required for each provided transaction,
+// applying state changes from previous transactions when estimating subsequent ones.
+// Transactions that become invalid or fail during execution for later estimations are ignored.
+// This method can help getting gas estimates for mutually depending transactions in bundles.
+func (a *PublicBundleAPI) EstimateGasForTransactions(
```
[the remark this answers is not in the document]
jenikd: Testing framework is dependant on the state Copy() function which should provide deep copy. Then it would be possible to test correct state changes. Fix is prepared and dependency to new carmen commit has to be updated.

api/sonicapi/estimate_gas_api_mock.go
jenikd: Removed interface and mock as they are not needed

api/sonicapi/estimate_gas_api.go
jenikd: As discussed here https://github.com/0xsoniclabs/sonic/pull/722#discussion_r2931890476
Created a constant for this and moved to common bundle package as it can be used on more places

api/sonicapi/estimate_gas_test.go
jenikd: Changed documentation and logic, when any of the transaction gas estimation fails, error for that failure is returned. Also added test for this case.

api/ethapi/log_tracer.go
```diff
@@ -19,10 +19,10 @@ package ethapi
 import (
 	"math/big"
 
-	"github.com/0xsoniclabs/tosca/go/tosca/vm"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/tracing"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/core/vm"
```
LuisPH3, line 25: 🥇

api/sonicapi/bundle_api.go
```diff
@@ -22,6 +22,12 @@ import (
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
+// MaxBundleTransactions is the maximum number of transactions
+// that can be included in a bundle for gas estimation.
+// The algorithm to estimate bundle gas is O(n^2),
+// therefore an upper bound is introduced.
+const MaxBundleTransactions = 16
```
LuisPH3: nit: this constant becomes `sonicapi.MaxBundleTransactions` which is quite generic. 
If there are not future uses for this variable planed, I would suggest that we call it `MaxNumEstimableTransactions`

api/rpc_test/backend.go
```diff
@@ -369,6 +370,11 @@ func (b *fakeBackend) ChainConfig(blockHeight idx.Block) *params.ChainConfig {
 	return opera.CreateTransientEvmChainConfig(b.chainID, heights, blockHeight)
 }
 
+// GetBundleExecutionInfo returns nil — not used in fake backend tests.
+func (b *fakeBackend) GetBundleExecutionInfo(_ common.Hash) *bundle.ExecutionInfo {
+	return nil
```
LuisPH3: nit, better panic. as it is not implemented yet in the current test cases .

api/sonicapi/estimate_gas_test.go
jenikd: Replaced with a test for upgrade heights and block override

api/sonicapi/estimate_gas_test.go
jenikd: I added proper test for block overrides

LuisPH3 (approved):

PR #827, opened 2026-04-09
Add RPC method sonic_getBundleInfo

LuisPH3:
This method can be used for waiting on the execution of a bundle, by execution plan hash. 

This PR does not register the use of the `sonic_` methods, this change is included in PR https://github.com/0xsoniclabs/sonic/pull/825. API will not be available until this is done.


api/sonicapi/get_bundle_info.go
```diff
@@ -0,0 +1,67 @@
...
+	executionPlanHash common.Hash,
+) (*RPCBundleInfo, error) {
+
+	// Check whether the given execution plan got already executed.
+	info := a.b.GetBundleExecutionInfo(executionPlanHash)
+	if info != nil {
+		return &RPCBundleInfo{
+			Block:    rpctest.ToBlockNum(info.BlockNumber),
+			Position: rpctest.ToHexUint(uint(info.Position.Offset)),
+			Count:    rpctest.ToHexUint(uint(info.Position.Count)),
```
HerbertJordan: Just to understand - the `rpctest` package is intended to be used in production code?

api/ethapi/backend.go
```diff
@@ -113,6 +114,9 @@ type Backend interface {
 	GetOriginatedFee(ctx context.Context, vid idx.ValidatorID) (*big.Int, error)
 
 	SccApiBackend
+
+	// Bundle API
+	GetBundleExecutionInfo(common.Hash) *bundle.ExecutionInfo
```
HerbertJordan: Suggestion: we should document here what the expectation is when the hash is unknown; I'd assume it is `nil`?

api/sonicapi/get_bundle_info.go
```diff
@@ -0,0 +1,67 @@
...
+// this method provides information about bundles executed in the recent past.
+// The sonic client is not capable of tracking bundles indefinitely, and may return
+// null for bundles executed too far in the past.
+//
+// In the same fashion as `eth_getTransactionReceipt`, this method returns a
+// non-error response with null payload if the bundle hasn't been executed yet.
+//
+// If the bundle has been executed, it returns the block number, position of the
+// first transaction of the bundle in the block, and the total number of non-reverted
+// transactions.
```
HerbertJordan: There could be reverted transactions that still made it into the block.

```suggestion
// first transaction of the bundle in the block, and the total number of transactions
// that got included in the block due to the execution of the bundle.
```

api/sonicapi/get_bundle_info.go
```diff
@@ -0,0 +1,67 @@
...
+			Count:    rpctest.ToHexUint(uint(info.Position.Count)),
+		}, nil
+	}
+
+	// Otherwise, the state is unknown (default).
+	return nil, nil
+}
+
+// RPCBundleInfo is the JSON RPC message returned by the GetBundleInfo API, which
+// provides information about the status of a transaction bundle.
```
HerbertJordan:
```suggestion
// provides information about the effect of the execution of a transaction bundle.
```

HerbertJordan (dismissed):
Looks good, just a few minor documentation issues.

api/sonicapi/get_bundle_info.go
LuisPH3: nop, is not. 
arg....

api/sonicapi/get_bundle_info.go
LuisPH3: A reverted transaction canno make it into the block. you mean failed transactions.

api/sonicapi/get_bundle_info.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/827/commits/9a07a2e7b089495a0a56ac7d4fe9b158887649b0

HerbertJordan (approved):

PR #828, opened 2026-04-09
Add bundle store utilities for genesis

facuMH:

**Main Contributions**
- `SetProcessedBundlesHistoryHash` an exported method to enable the genesis import to write the processed bundle history hash. 2d12a9f9f0969045474e188c977e7a221aeced22
- `EnumerateProcessedBundles` an exported method which returns a slice containing the `bundle.ExecutionInfo` corresponding to all the entries in the `ProcessedBundles` table. bc1c8762022a90c56e8d3663acedefd24edb13bc

**Minor Contributions**
- Utilisation of existing tooling. 6907c592ba4b23eee14ae19f9b59df6379356ba8
- Missing iterator release and mock ussage. d0aa253034f7286f50e1a2dd672141c3ca333172

### DIFF
--- a/gossip/store_processed_bundles.go
+++ b/gossip/store_processed_bundles.go
@@ -250,6 +250,22 @@ func (s *Store) GetProcessedBundleHistoryHash() (uint64, common.Hash) {
 	return blockNum, hash
 }
 
+// SetProcessedBundlesHistoryHash sets the block number and hash of the
+// processed bundles history. This should be used only during genesis initialization.
+func (s *Store) SetProcessedBundlesHistoryHash(blockNum uint64, hash common.Hash) {
+	// Make sure there is only one update at any time.
+	s.processedBundleMutex.Lock()
+	defer s.processedBundleMutex.Unlock()
+
+	err := s.table.ProcessedBundles.Put(nil, append(
+		bigendian.Uint64ToBytes(blockNum),
+		hash.Bytes()...,
+	))
+	if err != nil {
+		s.Log.Crit("failed to set hash of processed bundles", "error", err)
+	}
+}
+
 // --- utility functions for processed bundles management ---
 
 // getEntryKey returns the key used to store the presence of a processed bundle
--- a/gossip/store_processed_bundles.go
+++ b/gossip/store_processed_bundles.go
@@ -129,6 +129,7 @@ func (s *Store) deleteOutdatedBundles(blockNum uint64, batch kvdb.Batch) common.
 		// enough blocks have passed to start cleaning up the store
 		highestOutdatedBlockNumber := blockNum - bundle.MaxBlockRange + 1
 		it := s.table.ProcessedBundles.NewIterator([]byte{'i'}, nil)
+		defer it.Release()
 		for it.Next() {
 			key := it.Key()
 			// size of the key used to index processed bundle is:
--- a/gossip/store_processed_bundles.go
+++ b/gossip/store_processed_bundles.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 
 	"github.com/0xsoniclabs/sonic/gossip/blockproc/bundle"
+	"github.com/Fantom-foundation/lachesis-base/common/bigendian"
 	"github.com/Fantom-foundation/lachesis-base/kvdb"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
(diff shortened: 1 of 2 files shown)


LuisPH3 (dismissed):
I think this is much better than the previous PR, hopefully it serves all your purposes for the export and import. 

Question, the hash needs to be added after the history is loaded, right? it cannot be seed and recomputed while adding the history entries.

gossip/store_processed_bundles_test.go
```diff
@@ -507,6 +506,7 @@ func TestStore_deleteOutdatedBundles_RemovesBundles_WhenOld(t *testing.T) {
 			batch := NewMockstoreBatch(ctrl)
 			table := NewMockstoreTable(ctrl)
 			it := NewMockdbIterator(ctrl)
+			it.EXPECT().Release().AnyTimes()
```
HerbertJordan: nit: consider replacing the `AnyTimes()` here with something stricter. Otherwise, removing the release in the code would not be detected. I guess it should be called exactly once?

gossip/store_processed_bundles_test.go
```diff
@@ -550,6 +550,7 @@ func TestStore_deleteOutdatedBundles_RemovesMultipleEntries_WhenNotCleanedForToo
 	store.table.ProcessedBundles = table
 
 	it := NewMockdbIterator(ctrl)
+	it.EXPECT().Release().AnyTimes()
```
HerbertJordan: Same here.

gossip/store_processed_bundles_test.go
```diff
@@ -934,6 +935,143 @@ func TestStore_ProcessedBundles_RetainsAllBundlesRequiredToCoverTheMaximumBlockR
...
+}
+
+func TestStore_EnumerateProcessedBundles_ReturnsAllAddedEntries(t *testing.T) {
+
+	require := require.New(t)
+	store, err := NewMemStore(t)
+	require.NoError(err)
+
+	expected := map[common.Hash]bundle.ExecutionInfo{}
+	// fill the store with the maximum number of bundles
```
HerbertJordan: fyi: the maximum number could be way higher; you could have thousands of bundles per block; here you only have 1

gossip/store_processed_bundles.go
```diff
@@ -250,6 +252,59 @@ func (s *Store) GetProcessedBundleHistoryHash() (uint64, common.Hash) {
...
+	result := make([]bundle.ExecutionInfo, 0)
+
+	// get all recently processed bundles
+	it := s.table.ProcessedBundles.NewIterator([]byte{'e'}, nil)
+	defer it.Release()
+	for it.Next() {
+		key := it.Key()
+		value := it.Value()
+		if len(key) != 1+32 || len(value) != 16 {
+			s.Log.Crit(
```
HerbertJordan, line 287: This log message should only be a Debug message; if such a key is encountered, it should be skipped and the iteration continued. Wdyt?

HerbertJordan (dismissed):
Looks good and nice improvement compared to the previous version.

Please make sure that the tests are strict on the iterator-release calls and, if you agree, make the export function tolerate malformed entries.

gossip/store_processed_bundles_test.go
facuMH: right, it was meant to say maximum number of blocks.

gossip/store_processed_bundles.go
facuMH: I would rather keep the log.Crit.

gossip/store_processed_bundles_test.go
```diff
@@ -506,7 +507,9 @@ func TestStore_deleteOutdatedBundles_RemovesBundles_WhenOld(t *testing.T) {
 			batch := NewMockstoreBatch(ctrl)
 			table := NewMockstoreTable(ctrl)
 			it := NewMockdbIterator(ctrl)
-			it.EXPECT().Release().AnyTimes()
+			if c.finishingBlock >= bundle.MaxBlockRange-1 {
```
HerbertJordan, line 510: What's the point of the `if` if the expectation is a `AnyTimes()` ?

HerbertJordan (dismissed):

gossip/store_processed_bundles_test.go
facuMH: Last remaining anytimes. Deleted now.

LuisPH3 (approved):

PR #833, opened 2026-04-12
Update bundle format to support composition

HerbertJordan:
This PR upgrades the data format used to model transaction bundles to facilitate nested group compositions.

The bundle execution plan format introduces by this PR follows the following structure (borrowing Rust's enum structure syntax):
```Rust
struct ExecutionPlan {
    range: BlockRange,
    root: ExecutionStep
}

enum ExecutionStep {
    Tx{ flags: ExecutionFlags,  txRef: TxReference },
    Group{ oneOf: bool, tolerateFailed: bool, steps: []ExecutionStep },
}

struct TxReference {
   sender: common.Address,
   txSigHash: common.Hash,
}
```

Thus, an execution plan provides a valid block range and a hierarchical execution plan that is a tree of
- simple transactions to be processed
- a inter node listing sub-steps to be processed in an all-of or one-of mode, determined by the `oneOf` flag
Each node furthermore can define how the execution of the respective step should be interpreted regarding the success of the step. The execution flags define whether invalid or failed transactions or groups should be considered failed bundle execution steps.

**Examples**
A transaction `A` and `B` should be executed together, or not at all. The structure would look as follows:
```Go
ExecutionPlan{
    root: Group{
        oneOf: false,
        steps: []ExecutionStep{ Tx{ txRef: A} , Tx{txRef: B } },
    },
}
```
In a pretty-print, this would become a `AllOf(A,B)`.

In another case, `A` and `B` should be executed, followed by `C`. If `C` fails, it should be ignored. You would get
```Go
ExecutionPlan{
    root: Group{
        oneOf: false,
        steps: []ExecutionStep{ Tx{ txRef: A} , Tx{txRef: B } , Tx{flags: EF_TolerateFailure, txRef: C } },
    },
}
```
which in a pretty-print should be read and understood as `AllOf(A,B,Step[TolerateFailure](C))`.

Finally, if two out of three transactions A, B, and C should be executed, but also at most two, the plan could be defined as
```Go
ExecutionPlan{
    root: Group{
        oneOf: true,
        steps: []ExecutionStep{ 
            Group{oneOf: false, steps: []ExecutionStep{ Tx{ txRef: A }, Tx{ txRef: B } },
            Group{oneOf: false, steps: []ExecutionStep{ Tx{ txRef: A }, Tx{ txRef: C } },
            Group{oneOf: false, steps: []ExecutionStep{ Tx{ txRef: B }, Tx{ txRef: C } },
        },
    },
}
```
which should be read as `OneOf(AllOf(A,B), AllOf(A,C), AllOf(B,C))`.

**For reviewers**

The main changes are (in suggested reviewing order):
- the update of the `ExecutionPlan` type to support nested types; the type's encoding had to be updated accordingly, a debug-printing utility got added, and utility functions to extract needed information from execution plans
- the `TransactionBundle` was upgraded by replacing the list of transactions with an index of transactions to be referenced by execution plans; also, the execution plan is now stored explicitly in the bundle and not implicitly; this eliminated the need for the `extractExecutionPlan` function.
- the `Builder` was updated to support new, composed bundles

The rest of the PR covers adjustments of libraries already referencing the bundles to use the new bundle library infrastructure.


gossip/blockproc/bundle/execution_plan.go
```diff
@@ -17,36 +17,228 @@
...
 }
 
-// ExecutionPlan represents the plan for executing a bundle of transactions,
-// to which every participant in the bundle shall agree on.
-// The execution plan includes the list of steps to be executed, in the order of execution
-type ExecutionPlan struct {
-	Steps []ExecutionStep // Steps to be executed in the bundle, in the order of execution
-	Flags ExecutionFlags  // Execution flags that specify the behavior of the bundle execution
-	Range BlockRange      // Block range [Earliest, Latest] in which the bundle can be included
+// ExecutionStep is a single nodes in the hierarchy of processing steps defining an
```
LorenzSchueler: nit:

```suggestion
// ExecutionStep is a single node in the hierarchy of processing steps defining an
```

gossip/blockproc/bundle/execution_plan.go
```diff
@@ -17,36 +17,228 @@
...
+// execution plan. Each ExecutionStep is either a reference to a transaction to be
+// executed or a group of sub-steps. Groups may be marked as "oneOf", meaning
+// that only one of the sub-steps needs to be successfully executed for the
+// group to be considered successful. Furthermore, for each ExecutionStep, execution
+// flags can be defined to specify the behavior of the execution (e.g. whether
+// failed or invalid transactions should be tolerated).
+type ExecutionStep struct {
+	// -- common for individual transactions and groups --
+	flags ExecutionFlags
```
LorenzSchueler: I find a little bit weird why individual transactions now also have their own `ExecutionFlags`. IMO it is more intuitive to use the flags of the parent as it was the case before.
e.g. if you have a `OneOf[TolerateFailed]` with tx1 and tx2 it would be more intuitive that the `TolerateFailed` applies to `tx1` and `tx2` and not that they have their own flags.
Are there any specific reasons thy this is required?

gossip/blockproc/bundle/execution_flags_test.go
```diff
@@ -27,16 +27,9 @@ func TestExecutionFlags_Valid_AllCombinationsPass(t *testing.T) {
 	// explicit combination tests
 	tests := []ExecutionFlags{
 		0,
```
LorenzSchueler:
```suggestion
		EF_Default,
```

gossip/blockproc/bundle/builder.go
LorenzSchueler, line 48:
```suggestion
//  	Build(signer)
```

gossip/blockproc/bundle/block_range.go
```diff
@@ -71,3 +75,21 @@ func (r BlockRange) Size() uint64 {
 func (r BlockRange) IsInRange(blockNum uint64) bool {
 	return blockNum >= r.Earliest && blockNum <= r.Latest
 }
+
+func (r BlockRange) encode(writer io.Writer) error {
```
LuisPH3, line 80: nit: use rlp.

gossip/blockproc/bundle/bundle.go
```diff
@@ -81,62 +71,29 @@ var (
...
 type TransactionBundle struct {
-	Transactions types.Transactions
-	Flags        ExecutionFlags
-	Range        BlockRange // Block range [Earliest, Latest] in which the bundle can be included
+	Transactions map[TxReference]*types.Transaction
+	Plan         ExecutionPlan
 }
 
-func (tb *TransactionBundle) Encode() []byte {
+func (tb *TransactionBundle) GetTransactionsInExecutionOrder() []*types.Transaction {
```
LuisPH3: Because of execution flags and oneOf, the order of execution is pointless. This function is used in tests and in the c_block_callbacks to filter out unsuitable. 
Any order would satisfice and would prevent developers from assuming any property from it.

gossip/blockproc/bundle/execution_plan.go
```diff
@@ -17,36 +17,228 @@
...
+			continue
+		}
+		references[tx] = string([]byte{byte('A' + len(references))})
+	}
+	var out strings.Builder
+	s.print(references, &out)
+	return out.String()
+}
+
+func (s *ExecutionStep) GetTransactionReferencesInExecutionOrder() []TxReference {
```
LuisPH3: I believe that the tree is traversed in several locations for different purposes. 
the visitor pattern would be the right tool for the job.

gossip/blockproc/bundle/bundle.go
LuisPH3: Instead use `GetTransactionsInAnyOrder`.
Nit: deduplicated since they can be repeated.

gossip/blockproc/bundle/bundle.go
```diff
@@ -176,31 +133,41 @@ const (
...
+	Plan         []byte
 }
 
 func encodeInternal(
 	version byte,
 	bundle *TransactionBundle,
-) []byte {
+) ([]byte, error) {
+
+	// Create canonical form of list of included transactions.
```
LuisPH3, line 151: is there any other reason that reproducibility in tests?

gossip/blockproc/bundle/execution_plan.go
```diff
@@ -17,36 +17,228 @@
...
+	TxRef *TxReference `rlp:"nil"`
+	OneOf bool
+	Steps []stepEncodingV1
+}
+
+// -- debug and testing utilities --
+
+// String provides a human-readable representation of the step, which can be
+// useful for debugging or creating readable unit tests. It assigns a unique
+// letter (A, B, C, etc.) to each referenced transaction.
```
LuisPH3, line 234: nit, letters are not very useful debugging because the function assigning them is not bijective. I would use the first 4 characters of the tx hash instead.

gossip/blockproc/bundle/execution_plan.go
LuisPH3: I think a visitor could even replace the bundle processor logic, making it entirely generic respect to the transaction type. wdyt? 

```go
plan.Visit(func(flags, oneOf, Steps) (continue bool, err error){
```

gossip/blockproc/bundle/builder.go
```diff
@@ -310,23 +320,129 @@ func (b *builder) BuildEnvelopeAndPlan() (*types.Transaction, ExecutionPlan) {
...
 func (b *builder) Build() *types.Transaction {
 	envelope, _ := b.BuildEnvelopeAndPlan()
 	return envelope
 }
 
-// --- Utility Wrappers ---
+// --- implementation details ---
+
+// BuilderStep is a single transaction or a nested group in a bundle to build.
+type BuilderStep struct {
```
facuMH, line 343: could you move the step builder to another file?

gossip/blockproc/bundle/bundle_processor.go
```diff
@@ -58,35 +45,64 @@ type TransactionRunner interface {
 	RevertToSnapshot(id int)
 }
 
-// runAllOfBundle executes all transactions in the bundle and returns true if
-// all transactions are considered successful, false otherwise.
-func runAllOfBundle(
-	bundle *TransactionBundle,
+func runStep(
```
facuMH, line 53: could you add a docstring for `runStep`, `runAllOfGroup` and `runOneOfGroup`, so that one does not need to finde the documentation of the flags to know what is expected of each run alternative.

gossip/blockproc/bundle/execution_plan.go
HerbertJordan: I switched to the per-step version to avoid the need to use single-step groups for some use cases.

For example, let's assume you want to have 3 transactions A, B, and C. The first two need to be successful, the last one may fail. In the old version we would have to write something like:
```
AllOf( A, B, AllOf[TolerateFailed](C))
```
but it could as well be
```
AllOf( A, B, OneOf[TolerateFailed](C))
```
 In either way, the `AllOf` and `OneOf` is only added to get a scope for the execution flags. Their all-of and one-of semantic is not relevant. The new version fixes this by allowing you to say
```
AllOf( A, B, Step[TolerateFailed](C))
```
and no odd single-element group construct is needed to wrap up `C`.

gossip/blockproc/bundle/builder.go
HerbertJordan: `Build()` is correct.

gossip/blockproc/bundle/validate_test.go
```diff
@@ -187,31 +190,35 @@ func TestValidateEnvelope_AcceptsValidBlockRanges(t *testing.T) {
 		Gas  uint64
 	}{
 		"single-block range": {
-			From: 10, To: 10, Gas: 21240,
+			From: 10, To: 10, Gas: 21580,
```
facuMH: are these changes in gas needed because the amount of data encoded has changed ?

facuMH (commented):
The scope of this PR is too large. Some of the utilities could have been their own PRs and made the review process not so overwhelming. Given the changes in existing structures and addition of new structures, it would be useful to have new files for the new struct to better see what is tested

gossip/blockproc/bundle/bundle.go
LuisPH3: I cannot find the test testing that the order is respected neither.

gossip/blockproc/bundle/validate_test.go
HerbertJordan: Yes, the encoding size increased a bit, raising the floor gas costs.

gossip/blockproc/bundle/bundle_processor.go
HerbertJordan: Done in 64a7e009fb5d329306c9a69214cca733445fa1c9

gossip/blockproc/bundle/builder.go
HerbertJordan: I think it should remain with the rest of the builder.

gossip/blockproc/bundle/block_range.go
HerbertJordan: always RLP .. 690585a9154beb61efdc1503cd08acfd4823355d

gossip/blockproc/bundle/builder.go
```diff
@@ -30,14 +31,13 @@ import (
 )
 
 // This file offers utilities to build bundles from transaction data. The most
-// generic format is the NewBundle function, enabling the creation of an
+// generic format is the NewBuilder function, enabling the creation of an
 // envelope transaction carrying a bundle as follows:
 //
 //   envelope := NewBuilder(signer).
```
LuisPH3: For symmetry, I would suggest that we pass the signer in the build method instead of constructor. 
``` go
NewBuilder().Build(signer)
````

or remove entirely `NewBuilder` and only use `AllOf`  and `AnyOff`.

File: gossip/blockproc/bundle/execution_plan.go (hunk 8 above)
Comment (LuisPH3) [reply gossip/blockproc/bundle/execution_plan.go:234]: nevermind.

File: gossip/blockproc/bundle/bundle.go (hunk 5 above)
Comment (HerbertJordan) [reply gossip/blockproc/bundle/bundle.go]: This function is a life saver in unit tests, since working with the map is difficult and leads to unreadable code.

I renamed it to `GetTransactionsInReferencedOrder`. The order is important such that in test cases bundles can be picked apart easily. If the order would not be relevant, 
```Go
slices.Collect(maps.Values(bundle.Transactions))
```
would be sufficient. But in this case the order of the transactions is random.

I renamed it and added unit tests in a80bfdf067a3de9f2b028091a9a94191f24259ec

gossip/blockproc/bundle/execution_plan.go
HerbertJordan: I do not think this is necessary or makes code more readable for those use cases. I also think the bundle processor should be as plan and simple as possible.

gossip/blockproc/bundle/bundle.go
HerbertJordan: The idea is to have a canonical encoding to eliminate a source of variability. While I have no concrete issue in mind, I could think of situations where the random order of the encoded transactions or the non-identity of a combined encode-decode pair could lead to annoyance.

gossip/blockproc/bundle/execution_plan.go
HerbertJordan: The first 4 characters of the hash would also not be bijective either. I think for the intended purpose of testing and debugging the letter encoding is more readable.

What is not in this PR to not add extra code: the bundle should be print by mapping the letters to the details for the transaction, followed by the execution plan.

gossip/blockproc/bundle/builder.go
HerbertJordan: The best way would be to remove the `signer` parameter and use the default one, unless a `WithSigner` is called. But I thought of that as a sensitive topic and preserved what was there.

Only using the `AllOf ` and `OneOf`  directly does not provide the same expressiveness. There are use cases, like when you need to specify the block range, where you need the full builder.

gossip/blockproc/bundle/execution_plan.go
LuisPH3: The reasoning is that collecting transactions, printing strings, and executing are all traverses of the tree that could be centralized.  The to-string method is intrinsically the same as executing the bundle.
executing transactions using transactionsArgs and rcp.DoCall would use the same visitor as the one using types.Transactons, as neither of them are part of the plan. 

it is fine, we can use the current code.

gossip/blockproc/bundle/builder.go
LuisPH3: Just to have consistent code. Some builders are `x(signer).Build()` and other `y():Build(signer)`. Doesn't this grind your ears?  
Having two builder patterns to do the same is wasteful, independently of expressiveness. If one of them is superfluous please spare us from having to learn it.

gossip/blockproc/bundle/execution_plan.go
HerbertJordan: In fewer words: the scope of a flags should be the scope of a step, not the scope of a group.

LuisPH3 (dismissed):

gossip/blockproc/bundle/builder.go
HerbertJordan: I made the signer optional in the builder, as it should have been from the beginning (here e4182ce482bcb7f7a2844c8454cb8fec04e6e4bd).

gossip/blockproc/bundle/execution_plan.go
HerbertJordan: If this ever is an issue I will personally add a visitor. Takes me 20 minutes tops to write it, less than to discuss it here.

But this change does not need to be blocked for a visitor that right now just adds complexity and will delay it for another day of reviews.

gossip/blockproc/bundle/execution_plan.go
LorenzSchueler: Ok, I get that part. However now we are in a situation where we can apply the flags to individual groups and individual transactions, but the `TolerateInvalid` flag is not really doing anything to a group since both `runAllOfGroup` and `runOneOfGroup` are never returning `TransactionResultInvalid`.
At that point, `TolerateInvalid` could also be a "transaction field" in `ExecutionStep`.

We can also leave it as is, since it works. The semantics are just slightly confusing to me. But maybe I'm also just overlooking something.

HerbertJordan: @LorenzSchueler Yes, the `TolerateInvalid` is irrelevant for groups. We could check for the absence of that flag in the bundle validation, to consider it being set for groups as invalid.

To "enforce" it in the data structure, we would have some options. Currently we have
```Go
type ExecutionStep struct {
	// -- common for individual transactions and groups --
	flags ExecutionFlags

	// -- single transaction fields --
	txRef *TxReference // < if nil, it is a group

	// -- group fields --
	oneOf bool
	steps []ExecutionStep
}
```
which is the union type of the two variants 
```Rust
enum ExecutionStep {
    Tx{ flags: ExecutionFlags,  txRef: TxReference },
    Group{ flags: ExecutionFlags, oneOf: bool, steps: []ExecutionStep },
}
```
which is looks like the simplest way of implementing this in Go. We could re-define it as
```Go
type ExecutionStep struct {
	// -- single transaction fields --
	flags ExecutionFlags
	txRef *TxReference // < if nil, it is a group

	// -- group fields --
	oneOf bool
	ignoreFailed bool
	steps []ExecutionStep
}
```
such that there are no shared fields any more. Would this be preferred?

Implementing it using multiple types like
```Go
type TxStep struct {
	flags ExecutionFlags
	txRef *TxReference // < if nil, it is a group
}
type Group struct {
	oneOf bool
	ignoreFailed bool
	steps []ExecutionStep
}
```
might be an option, and I had a version like this before, but it adds a lot of boiler plate code for interfaces and member functions, making everything more complicated.

Wdyt?

HerbertJordan: We could do this:
```Go
type ExecutionStep struct {
	single *TxStep  // < only one must be set
	group *Group
}
type TxStep struct {
	flags ExecutionFlags
	txRef TxReference
}
type Group struct {
	oneOf bool
	ignoreFailed bool
	steps []ExecutionStep
}
```

LorenzSchueler: I think you first suggestion is definitely better than what we have currently.
```go
type ExecutionStep struct {
	// -- single transaction fields --
	flags ExecutionFlags
	txRef *TxReference // < if nil, it is a group

	// -- group fields --
	oneOf bool
	ignoreFailed bool
	steps []ExecutionStep
}
```

Alternatively we could also do it like this:
```go
type ExecutionStep struct {
    // -- common fields --
    ignoreFailed bool
        
	// -- single transaction fields --
	ignoreInvalid bool
	txRef *TxReference // < if nil, it is a group

	// -- group fields --
	oneOf bool
	steps []ExecutionStep
}
```
Which might be even cleaner. But no strong opinions either way.

In an ideal world I would prefer having two types, but I agree with you - it adds a lot of boilerplate and is in the not quite verbose (I had it like this in the original prototype).

LorenzSchueler: I just saw you last suggestion - let's go with this one.

HerbertJordan: I applied the changes to get to that version in d97b87e9582ff60990f59b4efac340a1e706c818.

gossip/blockproc/bundle/bundle_processor.go
```diff
@@ -55,63 +55,93 @@ func runStep(
...
+	}
+	if step.single != nil {
+		return runSingle(step.single, transactions, runner)
+	}
+	return runGroup(step.group, transactions, runner)
+}
+
+// runGroup executes a group of steps, which can be either one-of or all-of,
+// based on the group's execution semantic. It returns true if the group
+// execution is considered successful, and false otherwise.
```
LorenzSchueler:
```suggestion
// execution is considered tolerated, and false otherwise.
```

gossip/blockproc/bundle/bundle_processor.go
```diff
@@ -55,63 +55,93 @@ func runStep(
...
-	return core_types.TransactionResultSuccessful
+	return true
 }
 
 // runOneOfGroup executes a group of steps where at least one step must be
 // successful for the group to be considered successful. If all steps fail, the
 // entire group is reverted to the state before the group execution began, and
-// the function returns a failed result. After the first successful step,
-// processing of the group stops and the function returns a successful result.
+// the function returns false. After the first successful step, processing of
```
LorenzSchueler, line 109:
```suggestion
// the function returns false. After the first tolerated step, processing of
```

gossip/blockproc/bundle/bundle_processor.go
```diff
@@ -55,63 +55,93 @@ func runStep(
...
+			return true
 		}
 	}
 	runner.RevertToSnapshot(snapshot)
-	return core_types.TransactionResultFailed
+	return false
+}
+
+// runSingle executes a single transaction step and returns true if the
+// execution result is considered successful based on the execution flags, and
```
LorenzSchueler:
```suggestion
// execution result is considered tolerated based on the execution flags, and
```

LorenzSchueler (dismissed):
The latest change cleared up the semantics. Unfortunately, it introduced more noise than expected, but I'm fine with it.

gossip/blockproc/bundle/execution_plan.go
```diff
@@ -17,36 +17,298 @@
...
-	Flags ExecutionFlags  // Execution flags that specify the behavior of the bundle execution
-	Range BlockRange      // Block range [Earliest, Latest] in which the bundle can be included
+// ExecutionStep is a single node in the hierarchy of processing steps defining an
+// execution plan. Each ExecutionStep is either a reference to a transaction to be
+// executed or a group of sub-steps. Groups may be marked as "oneOf", meaning
+// that only one of the sub-steps needs to be successfully executed for the
+// group to be considered successful. Furthermore, for each ExecutionStep, execution
+// flags can be defined to specify the behavior of the execution (e.g. whether
+// failed or invalid transactions should be tolerated).
+
```
LuisPH3: nit: missing // at beginning of line, docstring is splitted, first paragraph is not linked to symbol.

gossip/blockproc/bundle/execution_plan.go
```diff
@@ -17,36 +17,298 @@
...
+		res.group = &copy
+		res.group.tolerateFailed = flags&EF_TolerateFailed != 0
+	}
+	return res
+}
+
+// valid returns true if the step is valid, meaning that it is either a single
+// step or a group step, but not both or neither. This is a basic validation to
+// ensure the integrity of the execution plan structure.
+func (s *ExecutionStep) valid() bool {
```
LuisPH3, line 154: nit: explicit test would be nice.

gossip/blockproc/bundle/bundle_processor.go
HerbertJordan: Since this function only deals with the yes/no result of nested steps `successful` seems to be more accurate for me.

facuMH (dismissed):

gossip/blockproc/bundle/execution_plan.go
```diff
@@ -17,36 +17,298 @@
...
+	TolerateFailed bool
+	Steps          []stepEncodingV1
+}
+
+// -- debug and testing utilities --
+
+// String provides a human-readable representation of the step, which can be
+// useful for debugging or creating readable unit tests. It assigns a unique
+// letter (A, B, C, etc.) to each referenced transaction.
+func (s *ExecutionStep) String() string {
```
LuisPH3, line 235: think that this function agreggates names of nested groups, but i am not sure if it meant to print the txs on this step only. 
 Is this correct?

LorenzSchueler (dismissed):

gossip/blockproc/bundle/execution_plan_test.go
```diff
@@ -79,3 +114,275 @@ func TestExecutionPlan_Hash_ComputesDeterministicHash(t *testing.T) {
...
+	step := NewOneOfStep(step1, step2)
+	require := require.New(t)
+	require.Nil(step.single)
+	require.NotNil(step.group)
+	require.True(step.group.oneOf)
+	require.Equal([]ExecutionStep{step1, step2}, step.group.steps)
+}
+
+func TestExecutionStep_WithFlags_ReturnsNewExecutionStepWithUpdatedFlags(t *testing.T) {
+	flags := []ExecutionFlags{
```
LuisPH3, line 223: nit: default missing.

gossip/blockproc/bundle/bundle_processor.go
```diff
@@ -58,37 +45,118 @@ type TransactionRunner interface {
...
-// all transactions are considered successful, false otherwise.
-func runAllOfBundle(
-	bundle *TransactionBundle,
+// runStep executes a single execution step, which may be a transaction or a
+// group of steps (one-of or all-of). It returns true if the step is considered
+// successful based on the execution result and its execution flags, and false
+// otherwise. The transaction index map is required to resolve transaction
+// references for steps that execute transactions.
+func runStep(
+	step *ExecutionStep,
```
LuisPH3: any reason for this to be a pointer?

gossip/blockproc/bundle/bundle_processor.go
```diff
@@ -58,37 +45,118 @@ type TransactionRunner interface {
...
+}
+
+// runGroup executes a group of steps, which can be either one-of or all-of,
+// based on the group's execution semantic. It returns true if the group
+// execution is considered tolerated, and false otherwise.
+func runGroup(
+	group *group,
+	transactions map[TxReference]*types.Transaction,
+	runner TransactionRunner,
+) bool {
```
LuisPH3, line 74: defensive programming: if group is nil, fail

gossip/blockproc/bundle/bundle_processor_test.go
```diff
@@ -17,238 +17,421 @@
...
+	require.True(t, RunBundle(bundle, runner))
 }
 
-func Test_RunBundle_RevertsToSnapshotOnFailure(t *testing.T) {
+func Test_runStep_DispatchesToCorrectExecutionMode(t *testing.T) {
 	tests := map[string]struct {
-		flags ExecutionFlags
+		step           ExecutionStep
+		expectSnapshot bool // < distinguishes group and transaction mode
+		expectedResult bool // < distinguishes all-of and one-of mode
```
LuisPH3: nit: expectedSuccess (because of boolean value)

gossip/blockproc/bundle/bundle_processor_test.go
```diff
@@ -17,238 +17,421 @@
...
-			tx3 := types.NewTx(&types.LegacyTx{})
+	require.False(t, runOneOfGroup(steps, txs, runner))
+}
 
-			gomock.InOrder(
-				runner.EXPECT().Run(tx1).Return(test.notToleratedResult),
-				runner.EXPECT().Run(tx2).Return(core_types.TransactionResultSuccessful),
-				// tx3 should not be run
-			)
+func Test_runOneOfGroup_StopsAtFirstSuccessfulTransaction(t *testing.T) {
```
LuisPH3, line 324: think it would make sense to check boundary cases. 
 successful tx is first, midle, last.

gossip/blockproc/bundle/bundle_processor_test.go
```diff
@@ -17,238 +17,421 @@
...
+func Test_runSingleTx_MissingTransaction_ReturnsTransactionResultInvalid(t *testing.T) {
+	res := runSingleTx(TxReference{}, nil, nil)
+	require.Equal(t, core_types.TransactionResultInvalid, res)
+}
+
 func Test_isTolerated_InterpretsExecutionFlagsCorrectly(t *testing.T) {
 	tests := []struct {
 		flags     ExecutionFlags
 		result    core_types.TransactionResult
 		tolerated bool
```
LuisPH3: nit: expectedTolerated (because of result being a test input)
nit2: multiline

gossip/c_block_callbacks_test.go
LuisPH3, line 1: are we missing a test `TestIsPermissible_Accepted`?

gossip/blockproc/bundle/execution_plan.go
HerbertJordan: Fixed in cd7f4a459af116b361e1b73deb1fda73b8502c1f

gossip/blockproc/bundle/execution_plan.go
HerbertJordan: fixed in 3c13f39a4fcadd655c010607faa4608308b21784

gossip/blockproc/bundle/execution_plan.go
HerbertJordan: It is intended -- and does -- print the full hierarchy rooted by step `s`.

gossip/blockproc/bundle/execution_plan_test.go
HerbertJordan: The default is not changing anything, so the test would not pass with it.

I added a dedicated test for restoring flags in 7423da8d35cb084ef13010b60ce6901dc0bdc424

gossip/blockproc/bundle/bundle_processor.go
HerbertJordan: Not any more; changed in ecd403229f623be19e22eca22a6de8c96c46dfda

gossip/blockproc/bundle/bundle_processor.go
HerbertJordan: Switched to pass-by-value 09128b2f0f58f7ecd3d0d329bc76599554a061b3

gossip/blockproc/bundle/bundle_processor_test.go
HerbertJordan: changed in 5645b662d99d9a3fa793963ee9576083b7e377ec

gossip/blockproc/bundle/bundle_processor_test.go
HerbertJordan: Does not help with readability, but added here for OneOf and AllOf: 8386236144138e7182aa53aa8da44eceda195e39

gossip/blockproc/bundle/bundle_processor_test.go
HerbertJordan: changed in 7a78cfe87fe4d3b8446a3ff90b99ef0c22c7e89c

Multiline makes it less readable.

gossip/c_block_callbacks_test.go
HerbertJordan: There is a test `TestIsPermissible_BundlesWithInvalidContent_Rejected` - why would we have one that accepts it?

LuisPH3: TestIsPermissible_AcceptsPermissibleTransactions

LuisPH3 (approved):

PR #834, opened 2026-04-12
Factor out TxData extraction

HerbertJordan:
This PR introduces a utility to extract TxData from a transaction.

The extraction of TxData is a useful utility in any context where transactions have to be modified. It enabled the modification of otherwise read-only `types.Transaction` instances.

Currently, this is done by inspecting the transaction type and reconstructing the TxData. In the future, Sonic's go-ethereum fork may be modified to provide access to internal go-etherum functions handling the copying of TxData.

The PR is also used to generalize the support of transaction types in bundles.

### DIFF
--- a/utils/txs_test.go
+++ b/utils/txs_test.go
@@ -1,0 +1,288 @@
+// Copyright 2026 Sonic Operations Ltd
+// This file is part of the Sonic Client
+//
+// Sonic is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Sonic is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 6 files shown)


utils/txs_test.go
```diff
@@ -0,0 +1,288 @@
...
+	"math/big"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/holiman/uint256"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetTxData_ExtractsAllData(t *testing.T) {
```
LuisPH3, line 30: This test checks that txData -> Transaction -> txData works.
But there is another use case which is very interesting:  Transaction -> txData -> Transaction, using the `types.NewTx` method. 
I am not sure if this respects the signature and chaind. it should, but it worths to have a test about it.

LuisPH3: checking the code i do not see any reason why this would not work.

LuisPH3 (approved):
:+1

PR #837, opened 2026-04-13
Add bundles to genesis export/import.

facuMH:
This PR implements the necessary mechanisms for adding bundles into genesis export/import mechanisms via the sonictool.

- For the `ProcessedBundlesHistoryHash` a new `HistoryHash` struct is added. This makes the encoding of the special entry possible via `rlp.EncodeToBytes`.
- Since both the historical hash and the list of execution infos need to be encoded into the same section, special measures are taken considering that the historical will be the first encoded info in the `bundles` section of the genesis. 

Note: most testing of this mechanisms are done via integration tests. That will come in a follow-up. In this PR there is only unit testing for `exportBundles`.

### DIFF
--- a/cmd/sonictool/genesis/export.go
+++ b/cmd/sonictool/genesis/export.go
@@ -472,7 +472,7 @@ func (f dropableFile) Drop() error {
 func MustRlpEncodeToByte(value any) []byte {
 	res, err := rlp.EncodeToBytes(value)
 	if err != nil {
-		panic("insert your favorite message here")
+		panic("failed to encode value to bytes: " + err.Error())
 	}
 	return res
 }
--- a/cmd/sonictool/genesis/export.go
+++ b/cmd/sonictool/genesis/export.go
@@ -118,6 +118,15 @@ func ExportGenesis(ctx context.Context, gdb *gossip.Store, includeArchive bool, 
 		return err
 	}
 
+	// bundles hash
+	writer = newUnitWriter(out)
+	if err := writer.Start(header, "bh", tmpPath); err != nil {
+		return err
+	}
+	if err := exportBundlesHash(ctx, gdb, writer, lastBlock); err != nil {
+		return err
+	}
+
 	// bundles
 	writer = newUnitWriter(out)
 	if err := writer.Start(header, "bundles", tmpPath); err != nil {
@@ -285,8 +294,8 @@ func exportFwaSection(ctx context.Context, gdb *gossip.Store, writer *unitWriter
 	return nil
 }
 
-func exportBundles(ctx context.Context, gdb *gossip.Store, writer *unitWriter, lastBlock idx.Block) error {
-	log.Info("Exporting processed bundles")
+func exportBundlesHash(ctx context.Context, gdb *gossip.Store, writer *unitWriter, lastBlock idx.Block) error {
+	log.Info("Exporting processed bundles history hash")
 
 	// write the history hash as the first item.
 	blockNum, histHash := gdb.GetProcessedBundleHistoryHash()
@@ -297,6 +306,17 @@ func exportBundles(ctx context.Context, gdb *gossip.Store, writer *unitWriter, l
 	if _, err := writer.Write(b); err != nil {
 		return err
 	}
+	hash, err := writer.Flush()
+	if err != nil {
+		return err
+	}
+	log.Info("Exported processed bundles history hash", "blockNum", blockNum, "hash", histHash)
+	fmt.Printf("- Processed bundles history hash: %v \n", hash.String())
+	return nil
+}
+
+func exportBundles(ctx context.Context, gdb *gossip.Store, writer *unitWriter, lastBlock idx.Block) error {
+	log.Info("Exporting processed bundles")
 
 	// write all the execution info from the store.
 	count := 0
--- a/cmd/sonictool/genesis/export.go
+++ b/cmd/sonictool/genesis/export.go
@@ -24,6 +24,7 @@ import (
 	"path"
 
 	"github.com/0xsoniclabs/sonic/gossip"
+	"github.com/0xsoniclabs/sonic/gossip/blockproc/bundle"
 	"github.com/0xsoniclabs/sonic/inter/ibr"
 	"github.com/0xsoniclabs/sonic/inter/ier"
 	"github.com/0xsoniclabs/sonic/opera/genesis"
@@ -114,6 +115,15 @@ func ExportGenesis(ctx context.Context, gdb *gossip.Store, includeArchive bool, 
 		return err
 	}
 	if err := exportBlockCertificates(ctx, gdb, writer, lastBlock); err != nil {
+		return err
+	}
+
+	// bundles
+	writer = newUnitWriter(out)
+	if err := writer.Start(header, "bundles", tmpPath); err != nil {
+		return err
+	}
+	if err := exportBundles(ctx, gdb, writer, lastBlock); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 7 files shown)


cmd/sonictool/genesis/export.go
```diff
@@ -275,6 +285,41 @@ func exportFwaSection(ctx context.Context, gdb *gossip.Store, writer *unitWriter
 	return nil
 }
 
+func exportBundles(ctx context.Context, gdb *gossip.Store, writer *unitWriter, lastBlock idx.Block) error {
+	log.Info("Exporting processed bundles")
+
+	// write the history hash as the first item.
+	blockNum, histHash := gdb.GetProcessedBundleHistoryHash()
+	b, _ := rlp.EncodeToBytes(bundle.HistoryHash{
```
HerbertJordan: This ignores an error.  This should not be done. If you do this because you can not create a test case that triggers an error here, you should think about how to reformulate the code. Even it is impossible now to create an error, since the data can always successfully be encoded, this might not be the case in the future.

Possible ways to deal with it and get test coverage: create a utility function
```Go
func MustRlpEncodeToByte(value any) []byte {
   res, err := rlp.EncodeToByte(value)
   if err != nil {
        panic("insert your favorite message here")
   }
   return res
}
```
As a generic function, this can be independently tested (e.g. by serializing a negative number and see that it panics). In the code here you can express the assumption that this is never going to fail explicitly, but you would also notice it if this ever changes.

cmd/sonictool/genesis/export.go
```diff
@@ -275,6 +285,41 @@ func exportFwaSection(ctx context.Context, gdb *gossip.Store, writer *unitWriter
...
+		Hash:        histHash,
+	})
+	if _, err := writer.Write(b); err != nil {
+		return err
+	}
+
+	// write all the execution info from the store.
+	count := 0
+	for _, info := range gdb.EnumerateProcessedBundles() {
+		b, _ := rlp.EncodeToBytes(info)
```
HerbertJordan: same here

gossip/blockproc/bundle/execution_info.go
```diff
@@ -33,3 +33,10 @@ type PositionInBlock struct {
 	Offset uint32
 	Count  uint32
 }
+
+// HistoryHash represents the block number and cumulative hash of the processed
+// bundles history. It is used for genesis export/import serialization.
+type HistoryHash struct {
```
HerbertJordan, line 39: Not sure this needs to be here ... looks like an implementation detail of the genesis export and the type could be there as well.

opera/genesisstore/store_genesis.go
```diff
@@ -216,6 +221,58 @@ func (s RawBlockCertificates) ForEach(fn func(cert.Certificate[cert.BlockStateme
...
+	for i := range 1000 {
+		f, err := s.fMap(BundlesSection(i))
+		if err != nil {
+			continue
+		}
+		stream := rlp.NewStream(f, 0)
+		// Skip the history hash (first item in the stream).
+		var h bundle.HistoryHash
+		if err := stream.Decode(&h); err != nil {
+			return
```
HerbertJordan: This looks like it is swallowing the error never to be seen again. Should this error be handled or reported somehow?

opera/genesisstore/store_genesis.go
```diff
@@ -216,6 +221,58 @@ func (s RawBlockCertificates) ForEach(fn func(cert.Certificate[cert.BlockStateme
...
+}
+
+func (s RawProcessedBundles) ForEach(fn func(bundle.ExecutionInfo) bool) {
+	for i := range 1000 {
+		f, err := s.fMap(BundlesSection(i))
+		if err != nil {
+			continue
+		}
+		stream := rlp.NewStream(f, 0)
+		// Skip the history hash (first item in the stream).
```
HerbertJordan: If there are multiple sections, the history hash would only be the first element of the first section, right? Should this be considered?

Maybe there could be two new streams, one for the bundle execution info and one for the history hash. If we implement the optional https://github.com/0xsoniclabs/sonic-admin/issues/706, the usage of a dedicated stream for the hashes might be simpler.

cmd/sonictool/genesis/export_test.go
```diff
@@ -0,0 +1,307 @@
...
+		{2}: {Offset: 1, Count: 1},
+		{3}: {Offset: 2, Count: 1},
+	})
+	writer3 := newDryRunWriter(t)
+	err = exportBundles(context.Background(), store3, writer3, 10)
+	require.NoError(t, err)
+
+	require.Greater(t, writer3.uncompressedSize, writer1.uncompressedSize,
+		"3 bundles should produce more data than 1 bundle")
+}
```
HerbertJordan, line 55: These three tests seem to all mainly target the size of the written data ... but not actually what is written. Are all 3 needed? Should there be a test checking what is actually written? Like, trying to re-import it again?

cmd/sonictool/genesis/export_test.go
```diff
@@ -0,0 +1,307 @@
...
+	storeWithBundles := setupBundleStore(t)
+	storeWithBundles.AddProcessedBundles(1, map[common.Hash]bundle.PositionInBlock{
+		{1}: {Offset: 0, Count: 1},
+	})
+	writerWithBundles := newDryRunWriter(t)
+	err = exportBundles(context.Background(), storeWithBundles, writerWithBundles, 10)
+	require.NoError(t, err)
+
+	require.Greater(t, writerWithBundles.uncompressedSize, histOnlySize,
+		"bundles should add data beyond just the history hash")
```
HerbertJordan: Not sure how this verifies that "HistoryHashAlwaysWrittenFirst" ...

HerbertJordan (changes requested):
Looks good, but a few issues should be addressed. Please see comments.

gossip/blockproc/bundle/execution_info.go
facuMH: Discussed offline. Type will remain here.

cmd/sonictool/genesis/export_test.go
facuMH: Refactor is coming, this test is removed and replaced by others.

cmd/sonictool/genesis/export.go
facuMH: f3d520c0ce83534c8140cfb86668929b67ff78e4

opera/genesisstore/store_genesis.go
facuMH: refactored in dd053fbd2009854860e4098db1c4e72c19772a5b

opera/genesisstore/store_genesis.go
facuMH: ae29759293176ea1bfebc75c87d4acef446bb048

cmd/sonictool/genesis/export_test.go
facuMH: refactored in dd053fbd2009854860e4098db1c4e72c19772a5b

cmd/sonictool/genesis/export.go
```diff
@@ -403,3 +468,11 @@ type dropableFile struct {
 func (f dropableFile) Drop() error {
 	return os.Remove(f.path)
 }
+
+func MustRlpEncodeToByte(value any) []byte {
+	res, err := rlp.EncodeToBytes(value)
+	if err != nil {
+		panic("insert your favorite message here")
```
HerbertJordan: Really, you copied this unchecked from my suggestion?

HerbertJordan (dismissed):
Looks OK, integration testing should cover the rest.

LuisPH3 (approved):

PR #839, opened 2026-04-14
Add submitBundle RPC function

jenikd:
This PR introduces new RPC function `sonic_submitBundle`. Final step in the bundle workflow. Takes the signed transactions and execution plan produced by `sonic_prepareBundle` and wraps them into a single envelope transaction that is submitted to the network.

### DIFF
--- a/api/sonicapi/submit_bundle.go
+++ b/api/sonicapi/submit_bundle.go
@@ -44,6 +44,7 @@ type SubmitBundleArgs struct {
 }
 
 // SubmitBundle implements the `sonic_submitBundle` RPC method, which submits a prepared bundle for execution.
+// Returns the hash of the execution plan and an error if any.
 func (a *PublicBundleAPI) SubmitBundle(
 	ctx context.Context,
 	args SubmitBundleArgs,
@@ -53,12 +54,12 @@ func (a *PublicBundleAPI) SubmitBundle(
 		return common.Hash{}, fmt.Errorf("signedTransactions must not be empty")
 	}
 
-	earliest, err := parseRPCBlockNumber(args.ExecutionPlan.Earliest, a.b)
+	earliest, err := parseRPCBlockNumber(a.b, args.ExecutionPlan.Earliest)
 	if err != nil {
 		return common.Hash{}, fmt.Errorf("invalid earliest block number: %w", err)
 	}
 
-	latest, err := parseRPCBlockNumber(args.ExecutionPlan.Latest, a.b)
+	latest, err := parseRPCBlockNumber(a.b, args.ExecutionPlan.Latest)
 	if err != nil {
 		return common.Hash{}, fmt.Errorf("invalid latest block number: %w", err)
 	}
@@ -105,7 +106,7 @@ func (a *PublicBundleAPI) SubmitBundle(
 	// Make a one use key to sign the bundle
 	key, err := crypto.GenerateKey()
 	if err != nil {
-		return common.Hash{}, fmt.Errorf("failed to generate signing key: %w", err)
+		return common.Hash{}, fmt.Errorf("failed to generate single use signing key: %w", err)
 	}
 
 	// Sign the bundle transaction with the one-use key and send it to the network
@@ -129,12 +130,16 @@ func (a *PublicBundleAPI) SubmitBundle(
 
 	// Submit the transaction to the network
 	_, err = ethapi.SubmitTransaction(ctx, a.b, tx)
-	return plan.Hash(), err
+	if err != nil {
+		return common.Hash{}, fmt.Errorf("failed to submit bundle transaction: %w", err)
+	}
+
+	return plan.Hash(), nil
 }
 
 // parseRPCBlockNumber converts an RPC block number (which can be a specific block number
 // or a tag like "latest") into a uint64 block number.
-func parseRPCBlockNumber(num rpc.BlockNumber, b BundleApiBackend) (uint64, error) {
+func parseRPCBlockNumber(b BundleApiBackend, num rpc.BlockNumber) (uint64, error) {
 
 	if num == rpc.PendingBlockNumber ||
 		num == rpc.LatestBlockNumber ||
--- a/api/sonicapi/submit_bundle.go
+++ b/api/sonicapi/submit_bundle.go
@@ -17,7 +17,7 @@
 package sonicapi
 
 import (
-	context "context"
+	"context"
 	"fmt"
 
 	"github.com/0xsoniclabs/sonic/api/ethapi"
@@ -27,6 +27,7 @@ import (
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/rpc"
 )
 
 // SubmitBundleArgs represents the arguments for the `sonic_submitBundle` RPC method.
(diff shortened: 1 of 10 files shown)


api/sonicapi/submit_bundle.go
```diff
@@ -0,0 +1,114 @@
...
+		})
+	if err != nil {
+		return common.Hash{}, fmt.Errorf("failed to sign bundle transaction: %w", err)
+	}
+
+	// Validate generated transaction
+	_, plan, err := bundle.ValidateEnvelope(signer, tx)
+	if err != nil {
+		return common.Hash{}, fmt.Errorf("failed to validate bundle transaction: %w", err)
+	}
```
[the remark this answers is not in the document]
jenikd: @LuisPH3 This validation should be done in the `bundle.ValidateEnvelope(signer, tx)` function, right?
https://github.com/0xsoniclabs/sonic/blob/86338b8e93df98080d7fbafcc7aa9b103ed14d93/gossip/blockproc/bundle/validate.go#L74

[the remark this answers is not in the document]
LuisPH3: yes, `bundle.ValidateEnvelope(signer, tx)` should suffice. 

long story: 
A more sophisticated version could simulate the execution: I see a usability problem with that. it rejects transactions after every wallet involved  has signed the transactions. This could be easily frustrating. 
A better option is to validate execution during prepare. that comes with its own technical challenges though.

api/sonicapi/submit_bundle_test.go
```diff
@@ -0,0 +1,369 @@
...
+		{"single tx, TolerateFailed", bundle.EF_TolerateFailed, 1},
+		{"two txs, TolerateFailed", bundle.EF_TolerateFailed, 2},
+		{"single tx, TolerateInvalid", bundle.EF_TolerateInvalid, 1},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			pool := rpctest.NewMockTxPool(ctrl)
+			pool.EXPECT().AddLocal(gomock.Any()).Return(nil)
```
LuisPH3: I would constrain the expectation, it should expect the envelope for the bundle.

I wrote once some tool for this:
```go
type envelopeMatcher struct {
	signer   types.Signer
	planHash common.Hash
}

func (m envelopeMatcher) Matches(x interface{}) bool {
	tx, ok := x.(*types.Transaction)
	if !ok {
		return false
	}
	if !bundle.IsEnvelope(tx) {
		return false
	}

	plan, err := bundle.ExtractExecutionPlan(m.signer, tx)
	if err != nil {
		return false
	}

	return plan.Hash() == m.planHash
}

func (m envelopeMatcher) String() string {
	return "envelopeMatcher for plan hash " + m.planHash.Hex()
}
```

Test looks like:
```go

func Test_SonicAPI_SubmitBundle(t *testing.T) {

	ctrl := gomock.NewController(t)
	pool := rpctest.NewMockTxPool(ctrl)

	be := rpctest.NewBackendBuilder(t).
		WithPool(pool).
		Build()
	api := NewPublicBundleAPI(be)

	wallet := rpctest.NewWallet(t)
	bundle, plan := bundle.NewBuilder(be.GetSigner()).
		With(bundle.Step(wallet.PrivateKey, &types.AccessListTx{Nonce: 1})).
		With(bundle.Step(wallet.PrivateKey, &types.AccessListTx{Nonce: 2})).
		BuildBundleAndPlan()

	pool.EXPECT().AddLocal(envelopeMatcher{
		signer: be.GetSigner(), planHash: plan.Hash(),
	}).Return(nil)

	args := encodeBundleRequest(t, plan, bundle.Transactions)
	_, err := api.SubmitBundle(t.Context(), args)
	require.NoError(t, err)
```

This test uses the builder to create an envelope offline, and submit creates a second one. 
Another option is to write a checker which contains the parameters of submit bundle, and verifies that the envelope passed to the pool is correct.

api/sonicapi/submit_bundle_test.go
```diff
@@ -0,0 +1,369 @@
...
+		})
+	}
+}
+
+func Test_SubmitBundle_PoolError_ReturnsError(t *testing.T) {
+	poolErr := errors.New("pool is full")
+
+	ctrl := gomock.NewController(t)
+	pool := rpctest.NewMockTxPool(ctrl)
+	pool.EXPECT().AddLocal(gomock.Any()).Return(poolErr)
```
LuisPH3, line 140: 👍

api/sonicapi/submit_bundle_test.go
LuisPH3: I just noticed this comment is better fitting  `Test_SubmitBundle_SubmittedEnvelopeMatchesBundleContents` than this case.

jenikd: I understand. Did it little bit different, but check added.

api/sonicapi/submit_bundle_test.go
```diff
@@ -0,0 +1,369 @@
...
+				SignedTransactions: []hexutil.Bytes{tt.txData},
+				ExecutionPlan:      validArgs.ExecutionPlan,
+			}
+
+			_, err = NewPublicBundleAPI(be).SubmitBundle(t.Context(), args)
+			require.ErrorContains(t, err, "failed to decode bundled transaction 0")
+		})
+	}
+}
+
```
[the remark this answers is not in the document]
jenikd: Added checks for RPCBlockNumber tag variants

LuisPH3 (dismissed):

api/sonicapi/submit_bundle.go
```diff
@@ -0,0 +1,157 @@
...
+	floorDataGas, err := core.FloorDataGas(data)
+	if err != nil {
+		return common.Hash{}, fmt.Errorf("failed to finalize bundle: could not calculate floor data gas: %w", err)
+	}
+	totalGas = max(totalGas, minGas, floorDataGas)
+
+	// Make a one use key to sign the bundle
+	key, err := crypto.GenerateKey()
+	if err != nil {
+		return common.Hash{}, fmt.Errorf("failed to generate signing key: %w", err)
```
facuMH: nit:
```suggestion
		return common.Hash{}, fmt.Errorf("failed to generate single use key: %w", err)
```

api/sonicapi/submit_bundle.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	// Submit the transaction to the network
+	_, err = ethapi.SubmitTransaction(ctx, a.b, tx)
+	return plan.Hash(), err
+}
+
+// parseRPCBlockNumber converts an RPC block number (which can be a specific block number
+// or a tag like "latest") into a uint64 block number.
+func parseRPCBlockNumber(num rpc.BlockNumber, b BundleApiBackend) (uint64, error) {
```
facuMH: This might be a personal style opinion, but what do you think about putting the backend parameter first?

api/sonicapi/submit_bundle.go
```diff
@@ -0,0 +1,157 @@
...
+
+	// Validate generated transaction
+	_, plan, err := bundle.ValidateEnvelope(signer, tx)
+	if err != nil {
+		return common.Hash{}, fmt.Errorf("failed to validate bundle transaction: %w", err)
+	}
+
+	// Submit the transaction to the network
+	_, err = ethapi.SubmitTransaction(ctx, a.b, tx)
+	return plan.Hash(), err
```
facuMH:
```suggestion
    if err != nil {
        return common.Hash{}, fmt.Errorf("failed to submit bundle %w", err)
    }
	return plan.Hash(), nil
```

api/sonicapi/submit_bundle.go
```diff
@@ -0,0 +1,157 @@
...
+	// any modification will invalidate the execution plan and result in an ill-formed bundle.
+	SignedTransactions []hexutil.Bytes `json:"signedTransactions"`
+	// ExecutionPlan contains the execution plan that each bundled transaction references.
+	// This value must be provided as returned by the `sonic_prepareBundle` method;
+	// any modification will invalidate the execution plan and result in an ill-formed bundle.
+	ExecutionPlan RPCExecutionPlan `json:"executionPlan,omitempty"`
+}
+
+// SubmitBundle implements the `sonic_submitBundle` RPC method, which submits a prepared bundle for execution.
+func (a *PublicBundleAPI) SubmitBundle(
```
facuMH, line 48:
```suggestion
// Returns the hash of the execution plan, and an error if any. 
func (a *PublicBundleAPI) SubmitBundle(
```

facuMH (dismissed):

facuMH (approved):

PR #840, opened 2026-04-14
Add `BlockHeight` to `eth_config` return struct.

facuMH:
**The Problem**
When Sonic generates more than one block within a second, the `ActivationTime` (which as per the [EIP-7910](https://eips.ethereum.org/EIPS/eip-7910) is a Unix epoch seconds) does not have the required granularity to properly identify the first block when the new rules are active.

**The Proposes Solution**
Add an extra field to the `config` struct, namely `BlockHeight` which has the block number of the first block where the new rules are active.  Since the block height was already used to calculate the `ForkId` no further changes are needed.

### DIFF
--- a/tests/eth_rpc_api_functions_test.go
+++ b/tests/eth_rpc_api_functions_test.go
@@ -276,11 +276,6 @@ func TestEthConfig_ProducesReadableConfig(t *testing.T) {
 	UpdateNetworkRules(t, net, originalRules)
 	net.AdvanceEpoch(t, 1)
 
-	// get current block to confirm epoch advancement
-	currentBlock, err := client.BlockByNumber(t.Context(), nil)
-	require.NoError(t, err, "could not get current block after epoch advancement")
-	require.Greater(t, currentBlock.NumberU64(), block1.NumberU64(), "block number did not advance after epoch advancement")
-
 	// get new config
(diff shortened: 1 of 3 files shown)


LuisPH3 (dismissed):

tests/eth_rpc_api_functions_test.go
```diff
@@ -295,28 +293,18 @@ func TestEthConfig_ProducesReadableConfig(t *testing.T) {
...
-		if previousBlock.Header().Time == uint64(response["current"]["activationTime"].(float64)) {
-			foundActivationBlock = true
-			activationBlockNumber = idx.Block(previousBlock.NumberU64())
-			break
-		}
-	}
+	// reinterpret block height from string to uint64
+	heightString := response["current"]["blockHeight"].(string)
+	heightUint, err := hexutil.DecodeUint64(heightString)
+	require.NoError(t, err, "could not decode block height from hex string")
```
HerbertJordan, line 295: The `activationTime` is no longer checked. Should it be compared to the timestamp of the block referenced by the block number?

HerbertJordan (dismissed):
Looks good, consider the request optional.

LuisPH3 (approved):

PR #844, opened 2026-04-15
Add bundles genesis export/import test

facuMH:
This PR add an end to end test using the integration test net to verify the export/import of bundles is supported via genesis.

This PR implements https://github.com/0xsoniclabs/sonic-admin/issues/718

### DIFF
--- a/cmd/sonictool/app/app_test.go
+++ b/cmd/sonictool/app/app_test.go
@@ -18,7 +18,6 @@ package app_test
 
 import (
 	"fmt"
-	"path/filepath"
 
 	"crypto/ecdsa"
 	"crypto/rand"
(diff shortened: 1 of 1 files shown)


cmd/sonictool/app/app_test.go
```diff
@@ -231,6 +236,54 @@ func TestSonicTool_genesis_ExportsAndSigns_WithoutErrors(t *testing.T) {
...
+	// import genesis file
+	err = sonictool.RunWithArgs([]string{
+		"sonictool",
+		"--datadir", stateDir,
+		"genesis", "--experimental", genesisFile,
+	})
+	require.NoError(t, err)
+
+	// start the network again.
+	require.NoError(t, net.Restart())
```
HerbertJordan: There is a function [RestartWithExportImport](https://github.com/0xsoniclabs/sonic/blob/f06854c4525e43bb6d9465b6f547fa49d25a11e9/tests/integration_test_net.go#L753) in the integration testnet type. Could this one be used here?

cmd/sonictool/app/app_test.go
```diff
@@ -582,3 +635,44 @@ func replaceUserPrompter(newPrompt prompt.UserPrompter) (cleanup func()) {
...
+	common.Hash,
+	*sonicapi.RPCBundleInfo,
+) {
+	t.Helper()
+	client, err := net.GetClient()
+	require.NoError(t, err)
+	defer client.Close()
+
+	account := tests.MakeAccountWithBalance(t, net, big.NewInt(1e18))
+	address := account.Address()
```
HerbertJordan: nit: since the transaction defined below is actually sending funds from the session sponsor to this account you would not have to create it here and add balance. You could just use whatever target address you want, including the zero address.

HerbertJordan (dismissed):

cmd/sonictool/app/app_test.go
```diff
@@ -231,6 +235,29 @@ func TestSonicTool_genesis_ExportsAndSigns_WithoutErrors(t *testing.T) {
...
+
+	bundleHash, originalInfo := runBundle(t, net)
+
+	require.NoError(t, net.RestartWithExportImport())
+
+	client, err := net.GetClient()
+	require.NoError(t, err)
+	defer client.Close()
+
+	// check that the bundle is still there after the export-import process
```
LuisPH3, line 266: nit: if you mention "still" i would expect that this has been checked before. 
add a sanity check and verify that this is correct before restart.

LuisPH3 (dismissed):
We miss a test for the hash, but I believe that we do not have tools for this atm.

LuisPH3 (approved):

PR #842, opened 2026-04-14
Add tracking of bundle execution offsets

HerbertJordan:
This PR introduces proper "true" tracking of transaction indices in the state processor by preserving the legacy tracking.

Since the impact of modifying the existing transaction index tracking is difficult to assess and beyond the scope of the current project, this PR opts to introduce an independent "trueTxOffset" tracking counter parallel to the existing mechanism now renamed to "legacyTxOffset".

The PR applies the corresponding changes to the state processor and the Opera-EVM processor. It also updates a range of unit tests and adds additional unit tests documenting and verifying the intended handling of tx-indices by bundles.

PR #845, opened 2026-04-15
Add unit test for transaction index offset in receipts

HerbertJordan:
This PR adds a unit test for the computation of the offset of the position of a transaction in a block conducted by the Opera EVM processor.

The test captures the current behavior of the implementation to make sure that subsequent updates (e.g. in #842) do not alter this behavior.


LorenzSchueler (approved):

simonlechner (approved):
To ensure that skipped transactions are not considered, the mock could also return a summary with a nil receipt. But this will make the checking of the index more difficult.

PR #846, opened 2026-04-15
Add integration test for concurrent bundle executions

HerbertJordan:
This PR adds an integration test sending hundreds of bundles concurrently into the network, some passing, some failing, and verifying in the end that the reported offset and number of accepted transactions for each of those bundles is correct.

It also adds a few additional utilities to the Integration-test infrastructure simplifying the implementation of concurrent tests. In particular the `SendAll` method allowing to send a list of transactions into the network in parallel.


tests/contracts/revert/revert.sol
```diff
@@ -19,4 +21,15 @@ contract RevertContract {
             invalid()
         }
     }
+
+    function probabilisticRevert() public {
+        // reverts a transaction during execution, depending on previous history
+        // and the address of the sender. This revert should not be reliably
+        // statically predictable.
+        count++;
+        uint256 rand = uint256(uint160(msg.sender)) | count;
```
LorenzSchueler: If the last bit of `msg.sender` is 1, this will never revert. Should probably be an XOR.

HerbertJordan: Updated in 14d6b3ee591c06aeab9df1f998df6e325bc91a04

LorenzSchueler (approved):

facuMH (approved):

LorenzSchueler (approved):

PR #847, opened 2026-04-15
Geth dependency: Add effective gas tip unit test

simonlechner:
Several functions are used from the geth dependency, during version updates these function could change behavior without anyone noticing. This is the first unit test to ensure changes are noticed.

### DIFF
--- a/tests/geth_compatibility/geth_dependency_test.go
+++ b/tests/geth_compatibility/geth_dependency_test.go
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with Sonic. If not, see <http://www.gnu.org/licenses/>.
 
-package geth_compatibility_tests
+package geth_compatibility
 
 import (
 	"math/big"
--- a/tests/geth_compatibility/geth_dependency_test.go
+++ b/tests/geth_compatibility/geth_dependency_test.go
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with Sonic. If not, see <http://www.gnu.org/licenses/>.
 
-package tests
+package geth_compatibility_tests
 
 import (
 	"math/big"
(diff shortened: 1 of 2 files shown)


tests/geth_compatibility/geth_dependency_test.go
```diff
@@ -0,0 +1,120 @@
...
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			txData := &types.DynamicFeeTx{
+				GasTipCap: test.gasTipCap,
+				GasFeeCap: test.gasFeeCap,
+			}
+			tx := types.NewTx(txData)
+			_, err := tx.EffectiveGasTip(test.baseFee)
+			require.Error(t, err)
+		})
```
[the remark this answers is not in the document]
simonlechner: Behavior should not depend on the kind of error, we only care whether there is an error.

tests/geth_dependency_test.go
```diff
@@ -0,0 +1,120 @@
...
+			gasTipCap: big.NewInt(10),
+			gasFeeCap: big.NewInt(40),
+		},
+		"base fee overflow": {
+			baseFee: overflow,
+		},
+		"tip cap overflow": {
+			gasTipCap: overflow,
+		},
+		"fee cap overflow": {
```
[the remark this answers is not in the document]
simonlechner: Test coverage shows that all the expected checks are reached. By setting the other values, the last two test cases test the same case as the first test case.

tests/geth_compatibility/geth_dependency_test.go
LuisPH3, line 1: nit: maybe make a folder inside `tests`:  `upstream_compat` `geth_compatibility` ... .

HerbertJordan, line 1: While it is a start, I am skeptical that checking a few sample inputs can be effective in detecting future changes.

HerbertJordan (dismissed):

LuisPH3 (dismissed):

tests/geth_compatibility/geth_dependency_test.go
simonlechner: Yes, sadly the list of functions used from geth is much longer than I expected.

LuisPH3 (approved):

PR #848, opened 2026-04-15
Add unit tests for block range validation

HerbertJordan:
Add unit tests for block range validation.


facuMH (approved):
This PR adds unit tests for `ValidateRange`, verifying it reports valid and invalid ranges.

LuisPH3 (approved):

PR #850, opened 2026-04-15
Add execution step validation

HerbertJordan:
This PR introduces validation for composed `ExecutionStep`s of bundles.

The main properties checked include:
- all steps are valid, thus none defines a single and a group step, nor neither
- the execution flags provided for each step are valid
- group steps are composed of valid individual steps
- the maximum nesting depth of groups is limited

### DIFF
--- a/gossip/blockproc/bundle/validate.go
+++ b/gossip/blockproc/bundle/validate.go
@@ -26,6 +26,11 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 )
 
+const (
+	// MaxNestingDepth defines the maximum allowed nesting depth of execution steps.
+	MaxNestingDepth = 16
+)
+
 var ErrWrongEnvelopeGasLimit = errors.New("gas limit of envelope does not match gas limit of payload")
 
 // ValidateEnvelope validates an envelope and its contents.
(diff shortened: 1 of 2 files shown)


gossip/blockproc/bundle/validate.go
```diff
@@ -26,6 +26,11 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 )
 
+const (
+	// MaxNestingDepth defines the maximum allowed nesting depth of execution steps.
+	MaxNestingDepth = 16
```
facuMH, line 44: is this limited by memory or what ? if in the future someone wants to change it, maybe it would be useful for the comment to mention the restrictions to keep in mind.

gossip/blockproc/bundle/validate.go
```diff
@@ -133,6 +138,46 @@ func validateEnvelopeInternal(
 	return &txBundle, &plan, nil
 }
 
+// validateStep checks that the given execution step is valid.
+func validateStep(step ExecutionStep) error {
```
facuMH, line 155: given that none of the errors here mention bundles, just "execution steps", the callers of this function should mention bundles in the case of an error, otherwise this could be a bit out of context.

facuMH (dismissed):
Validation is consistent with the PR description. Just a couple of minor comments.

LuisPH3 (dismissed):

gossip/blockproc/bundle/validate.go
HerbertJordan: added documentation in dbac70448e224f29e69f209b8b07ebbb32a046d7

gossip/blockproc/bundle/validate.go
HerbertJordan: Alright. This function is private and to be integrated in the bundle validation to be updated accordingly in a follow-up PR (see code above).

facuMH (dismissed):

LuisPH3 (dismissed):

LuisPH3 (approved):

PR #851, opened 2026-04-15
Improve integration test utils

LorenzSchueler:
This PR improves the integration test utils.
It now provides a function to easily deploy the revert contract which will be used in an upcoming PR.
Also, `SetTransactionDefaults` no longer modifies the passed in transaction, but makes a copy on modifies this copy instead, so that this function has no side effects.


tests/bundles/bundle_test_utils.go
```diff
@@ -95,3 +100,33 @@ func WaitForBundleExecutions(
...
 }
+
+// deployContract deploys a contract using the provided deploy function and prepares the input for calling the specified method.
+//
+//nolint:unused
+func deployContract[T any](
+	t testing.TB, session tests.IntegrationTestNetSession,
+	getABI func() (*abi.ABI, error),
+	deployFunc tests.ContractDeployer[T],
+	methodName string,
```
HerbertJordan: nit: given the name, it is unexpected that this function also prepares input data for a method call. This should either be reflected in the name (something like `deployContractAndGetMethodCallParameters`) or better, kept apart in two different functions.

tests/bundles/bundle_test_utils.go
```diff
@@ -95,3 +100,33 @@ func WaitForBundleExecutions(
 	})
 	return infos, err
 }
+
+// deployContract deploys a contract using the provided deploy function and prepares the input for calling the specified method.
+//
+//nolint:unused
+func deployContract[T any](
+	t testing.TB, session tests.IntegrationTestNetSession,
+	getABI func() (*abi.ABI, error),
```
HerbertJordan: nit: could be an interface
```Go
type abiProvider interface {
    GetAbi() (*abi.ABI, error)
}
```
to which you can then pass `revert.RevertMetaData` instead of the function pointer. This simplifies the call side.

tests/bundles/bundle_test_utils.go
```diff
@@ -95,3 +100,33 @@ func WaitForBundleExecutions(
...
+
+	input, err := abi.Pack(methodName)
+	require.NoError(t, err, "failed to pack input for method %s; %v", methodName, err)
+
+	return receipt.ContractAddress, input
+}
+
+// prepareRevertContract deploys the Revert contract and prepares the input for calling the doCrash method, which always reverts.
+//
+//nolint:unused
```
LuisPH3: remove nolint, export it.

tests/bundles/bundle_test_utils.go
```diff
@@ -95,3 +100,33 @@ func WaitForBundleExecutions(
...
+func deployRevertContract(t testing.TB, session tests.IntegrationTestNetSession) (common.Address, []byte) {
```
HerbertJordan: The name does not intuitively give way what that the second result value provides you the parameters needed for the `doCrash` function. This should be reflected in then name such that the call side remains readable.

tests/bundles/bundle_test_utils.go
```diff
@@ -95,3 +100,33 @@ func WaitForBundleExecutions(
 	})
 	return infos, err
 }
+
+// deployContract deploys a contract using the provided deploy function and prepares the input for calling the specified method.
```
LuisPH3: this function name is misleading. It is also no very generic as it can only prepare calls to contract functions without arguments.
i think this function because of scope, name, intended use... only makes sense in the context of an specific test and therefore should be merged with it. (and named accordingly)

tests/test_utils.go
```diff
@@ -143,30 +143,39 @@ func SetTransactionDefaults[T types.TxData](
...
+		return any(&copied).(T)
 	case *types.DynamicFeeTx:
-		tx.Nonce = nonce
-		tx.Gas = gas
-		tx.GasFeeCap = gasPrice
+		copied := *tx
+		copied.Nonce = nonce
+		copied.Gas = gas
+		copied.GasFeeCap = gasPrice
+		return any(&copied).(T)
```
LuisPH3, line 164: would `types.TxData(&copied)` suffice?

HerbertJordan (dismissed):
The proposed `deployContract` method seems to cover two distinct tasks that should better be covered by two distinct calls. But if it makes life easier for you, it is fine with me.

tests/test_utils.go
LorenzSchueler: No, that does not work because we are not returning the interface, but T which is generic over this interface. Therefore, the return type must match the input type, and we can not return the interface type itself.

tests/bundles/bundle_test_utils.go
LorenzSchueler: But will be used only in this module.

LorenzSchueler: It will get removed in the next PR which uses these functions.

tests/bundles/bundle_test_utils.go
LorenzSchueler: I split it up into two functions so that it can be reused easier. If we need support for arguments, we can add it later.

tests/bundles/bundle_test_utils.go
LorenzSchueler: I opted to pass it as `*bind.MetaData`

tests/test_utils.go
LuisPH3: right. I implemented it this way so that you do not hide the instance behind the interface and you can still modify members after setting defaults.

tests/bundles/bundle_test_utils.go
LorenzSchueler: Done

LuisPH3 (approved):

PR #852, opened 2026-04-16
Test semantics of execution flags of single execution step in bundle

LorenzSchueler:
This PR adds an integration tests for the bundles which tests that the execution flags of single transactions follow the intended semantics.


HerbertJordan (approved):
Looks good, thanks!

PR #854, opened 2026-04-16
Add Composable JSON execution plan

LuisPH3:
This PR adds a second RPC structure to describe composable execution plans,
this type is to be returned form `prepare bundle` and accepted by `submit bundle`.

The current implementation suffixes symbol names with `Composable`, this is to be removed with the migration of submit bundle into the new structure. 

This PR builds on https://github.com/0xsoniclabs/sonic/pull/860


api/sonicapi/execution_plan.go
```diff
@@ -0,0 +1,229 @@
...
+//
+// An example of the JSON representation of an execution plan is as follows:
+//
+//		{
+//	   	"blockRange":{
+//				"earliest":"0xa",
+//				"latest":"0x15"
+//			},
+//		    "root":{
+//				"group":{
```
jenikd: Indenting is off here

api/sonicapi/execution_plan.go
```diff
@@ -0,0 +1,229 @@
...
+	return bundle.ExecutionStep{}, fmt.Errorf("invalid execution plan level: must have either single or group")
+}
+
+// toJsonExecutionPlanVisitor is an implementation of the ExecutionPlanVisitor interface
+type toJsonExecutionPlanVisitor struct {
+	levelStack []RPCExecutionPlanLevel
+	groupStack []*RPCExecutionPlanGroup
+}
+
+// newRPCExecutionPlanVisitor creates a new instance of toJsonExecutionPlanVisitor.
```
jenikd: wrong function name in comment

api/sonicapi/execution_plan.go
```diff
@@ -0,0 +1,229 @@
...
+
+	root, err := toBundleExecutionPlanLevel(rpcPlan.Root)
+	if err != nil {
+		return bundle.ExecutionPlan{}, fmt.Errorf("invalid execution plan root: %w", err)
+	}
+
+	return bundle.ExecutionPlan{
+		Range: bundle.BlockRange{
+			Earliest: uint64(rpcPlan.BlockRange.Earliest),
+			Latest:   uint64(rpcPlan.BlockRange.Latest),
```
jenikd, line 128: converting rpc.BlockNumber to uint64 is not safe as it can be negative for example for `rpc.LatestBlockNumber`. I'm preparing function for this as this will be used on more places in the code. Will fix this, once it is merged.

api/sonicapi/execution_plan.go
```diff
@@ -0,0 +1,229 @@
...
+		},
+		Root: root,
```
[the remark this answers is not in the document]
LuisPH3: This makes sense, for simplicity, rpc.BlockNumber is not to be used as tags will not be generated nor used.

api/sonicapi/execution_plan.go
LuisPH3: thanks, got deleted.

api/sonicapi/execution_plan.go
LuisPH3: because of comment https://github.com/0xsoniclabs/sonic/pull/854#discussion_r3092495088
I replaced ranges using `rpc.BlockNumber` by `hexutil.Uint64`.
The reason is that these can never be "latest" or "pending", they refer to future block numbers and the easiest thing is not to mix them up.  
 Wdyt?

jenikd: Yes, that makes sense. Thanks

api/sonicapi/execution_plan.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/854/commits/2bbf0ad67950e483a96279caae67fb1fb5474f40

jenikd (dismissed):

api/sonicapi/execution_plan.go
```diff
@@ -0,0 +1,224 @@
...
+			group = group.WithFlags(bundle.EF_TolerateFailed)
+		}
+		return group, nil
+	}
+	return bundle.ExecutionStep{}, fmt.Errorf("invalid execution plan level: must have either single or group")
+}
+
+// toJsonExecutionPlanVisitor is an implementation of the ExecutionPlanVisitor interface
+type toJsonExecutionPlanVisitor struct {
+	levelStack []RPCExecutionPlanLevel
```
HerbertJordan: nit: If I understand the visitor correctly, this list is only ever filled once with the final element of the traversal. It this thus not really a stack but a location to store the "parsing" result. It could be renamed and typed to
```Go
root RPCExecutionPlanLevel
```
or
```Go
result RPCExecutionPlanLevel
```

api/sonicapi/execution_plan.go
```diff
@@ -0,0 +1,224 @@
...
+		return bundle.ExecutionStep{},
+			fmt.Errorf("invalid execution plan level: cannot have both single and group")
+	}
+
+	if rpcLevel.Single != nil {
+		ref := bundle.NewTxStep(bundle.TxReference{
+			From: rpcLevel.Single.From,
+			Hash: rpcLevel.Single.Hash,
+		})
+		flags := bundle.ExecutionFlags(0)
```
HerbertJordan: nit:

```suggestion
		flags := bundle.EF_Default
```

HerbertJordan (dismissed):
Looks good, feel free to ignore the comments.

api/sonicapi/execution_plan.go
LuisPH3: right! thanks

jenikd (dismissed):

LuisPH3:
replaced by https://github.com/0xsoniclabs/sonic/pull/861

PR #863, opened 2026-04-17
Test nested bundles can be executed

LorenzSchueler:
This PR adds a test to make sure nested bundles can be executed.

### DIFF
--- a/tests/bundles/nested_bundle_test.go
+++ b/tests/bundles/nested_bundle_test.go
@@ -49,7 +49,7 @@ func TestBundle_NestedBundlesCanBeExecuted(t *testing.T) {
 
 	tx := tests.SetTransactionDefaults(t, net, &types.AccessListTx{}, sender)
 
-	innerEnvelope, innerBundle, _ := bundle.NewBuilder().
+	innerEnvelope, innerBundle, innerPlan := bundle.NewBuilder().
 		WithSigner(signer).
 		SetEarliest(blockNumber).
 		AllOf(bundle.Step(sender.PrivateKey, tx)).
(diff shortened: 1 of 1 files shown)


tests/bundles/nested_bundle_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// Check bundle status before submission.
+	_, err = GetBundleInfo(t.Context(), client.Client(), outerPlan.Hash())
+	require.ErrorIs(t, err, ethereum.NotFound)
+
+	// Run the bundle.
+	require.NoError(t, client.SendTransaction(t.Context(), outerEnvelope))
+
+	// Wait for the bundle to be processed.
+	info, err := WaitForBundleExecution(t.Context(), client.Client(), outerPlan.Hash())
```
HerbertJordan: There should also be a bundle info for the inner bundle. Can you test that it is available too, and that the offsets and count matches for this specific case?

HerbertJordan (changes requested):

HerbertJordan (changes requested):
Test looks good. 

Currently, there are extra changes from another PR that should disappear after a rebase. Will approve when done.

tests/bundles/nested_bundle_test.go
```diff
@@ -0,0 +1,90 @@
...
+	require.NoError(t, err)
+
+	// The outer bundle only contains the inner bundle, so they should have the
+	// same execution info.
+	require.Equal(t, outerInfo, innerInfo)
+
+	// Verify that there is no receipt for the envelopes themselves.
+	_, err = client.TransactionReceipt(t.Context(), outerEnvelope.Hash())
+	require.ErrorIs(t, err, ethereum.NotFound)
+	_, err = client.TransactionReceipt(t.Context(), innerEnvelope.Hash())
```
HerbertJordan, line 82: Note: the `innerEnvelope` was actually never send to the network. You can test for it, but the more relevant one would be `outerBundle.GetTransactionsInReferencedOrder()[0]`, which is the new envelope created by the builder while re-packaging the nested transaction.

HerbertJordan (dismissed):

tests/bundles/nested_bundle_test.go
```diff
@@ -0,0 +1,93 @@
...
+		BuildEnvelopeBundleAndPlan()
+
+	// Run the bundle.
+	require.NoError(t, client.SendTransaction(t.Context(), outerEnvelope))
+
+	// Wait for the bundle to be processed.
+	outerInfo, err := WaitForBundleExecution(t.Context(), client.Client(), outerPlan.Hash())
+	require.NoError(t, err)
+
+	innerInfo, err := WaitForBundleExecution(t.Context(), client.Client(), innerPlan.Hash())
```
LuisPH3: If you are updating this PR, I would wait on one bundle and then call `bundleGetInfo` on the other. Which should be expected to be already executed.  (wait on one, expect the other to be in immediately available in the same block)

LuisPH3 (dismissed):

LuisPH3 (approved):

PR #864, opened 2026-04-17
Add bundle validation

HerbertJordan:
Adds support for validating bundles. 

Properties that are verified include:
- bundles contain a valid execution plan
- there are no `nil` transactions
- all transactions are targeting the same chain ID
- the indexing of the transactions in the bundles is valid (correct sender and transaction signing hash)
- all transactions in the bundle accept the provided execution plan
- all transactions needed for the plan are present
- there are no extra transactions that are not needed for the plan

### DIFF
--- a/gossip/blockproc/bundle/validate.go
+++ b/gossip/blockproc/bundle/validate.go
@@ -176,7 +176,7 @@ func validateBundle(
 	for _, tx := range bundle.Transactions {
 		if chainId == nil {
 			chainId = tx.ChainId()
-		} else if id := tx.ChainId(); id == nil || chainId.Cmp(tx.ChainId()) != 0 {
+		} else if id := tx.ChainId(); id == nil || chainId.Cmp(id) != 0 {
 			return fmt.Errorf("transactions in bundle have different chain IDs")
 		}
 	}
--- a/gossip/blockproc/bundle/validate.go
+++ b/gossip/blockproc/bundle/validate.go
@@ -19,7 +19,6 @@ package bundle
 import (
 	"errors"
 	"fmt"
-	"math/big"
 	"slices"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -168,16 +167,6 @@ func validateBundle(
 	for _, tx := range bundle.Transactions {
 		if tx == nil {
 			return fmt.Errorf("invalid nil transaction in bundle")
-		}
-	}
-
-	// check that all transactions use the same Chain ID
-	var chainId *big.Int
-	for _, tx := range bundle.Transactions {
-		if chainId == nil {
-			chainId = tx.ChainId()
-		} else if id := tx.ChainId(); id == nil || chainId.Cmp(id) != 0 {
-			return fmt.Errorf("transactions in bundle have different chain IDs")
 		}
 	}
 
--- a/gossip/blockproc/bundle/validate.go
+++ b/gossip/blockproc/bundle/validate.go
@@ -19,6 +19,7 @@ package bundle
 import (
 	"errors"
 	"fmt"
+	"math/big"
 	"slices"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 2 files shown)


gossip/blockproc/bundle/validate.go
```diff
@@ -151,6 +152,83 @@ func validateEnvelopeInternal(
...
+// that it is well-formed and consistent.
+func validateBundle(
+	signer types.Signer,
+	bundle TransactionBundle,
+) error {
+
+	// check the execution plan for validity
+	if err := validatePlan(bundle.Plan); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
HerbertJordan: That is the task of future work.

gossip/blockproc/bundle/validate.go
```diff
@@ -151,6 +152,88 @@ func validateEnvelopeInternal(
...
+			return fmt.Errorf("invalid nil transaction in bundle")
+		}
+	}
+
+	// check that all transactions use the same Chain ID
+	var chainId *big.Int
+	for _, tx := range bundle.Transactions {
+		if chainId == nil {
+			chainId = tx.ChainId()
+		} else if id := tx.ChainId(); id == nil || chainId.Cmp(tx.ChainId()) != 0 {
```
facuMH:
```suggestion
		} else if id := tx.ChainId(); id == nil || chainId.Cmp(id) != 0 {
```

gossip/blockproc/bundle/validate.go
```diff
@@ -151,6 +152,88 @@ func validateEnvelopeInternal(
...
+
+	// check that there are no nil transactions in the bundle
+	for _, tx := range bundle.Transactions {
+		if tx == nil {
+			return fmt.Errorf("invalid nil transaction in bundle")
+		}
+	}
+
+	// check that all transactions use the same Chain ID
+	var chainId *big.Int
```
facuMH: shouldn't the chainID come from the `signer.ChainID()` ?

facuMH (dismissed):
bundle validation implementation and testing are consistent with the PR description. Just a small comment about source of ChainId.

gossip/blockproc/bundle/validate.go
LuisPH3: yes, this check is superfluous. Calling signer.Sender will return an error if the chainid does not match.

HerbertJordan: "Should" - in my opinion no. Could, yes.

I think it is cleaner to compare the ChainIDs given by transactions in this step, since those should be equivalent. If the signer is not consistent with this, it is detected a few lines below, when the senders are extracted from the transactions.

gossip/blockproc/bundle/validate.go
```diff
@@ -151,6 +152,88 @@ func validateEnvelopeInternal(
...
+	// check that all transactions in the bundle agree to the execution plan
+	planHash := bundle.Plan.Hash()
+	for _, tx := range bundle.Transactions {
+		if !belongsToExecutionPlan(tx, planHash) {
+			return fmt.Errorf("contains transaction not approving the execution plan")
+		}
+	}
+
+	// check that all transactions referenced by the plan are present in the bundle
+	references := map[TxReference]struct{}{}
```
LuisPH3, line 206: Something that I was thinking about while doing the RPC types, is that transactions can be referenced multiple times within one plan. 
Current implementation accepts any number of repetitions, maybe this should be keep controlled to avoid spamming data in bundles designed to fail.

gossip/blockproc/bundle/validate.go
HerbertJordan: Remove the explicit ChainID check in 249efa533c8479a9eb614636eb262697263ce6c5

gossip/blockproc/bundle/validate_test.go
```diff
@@ -535,30 +535,37 @@ func TestValidateBundle_NilTransaction_Rejected(t *testing.T) {
 }
 
 func TestValidateBundle_InconsistentChainIds_Rejected(t *testing.T) {
+	key, err := crypto.GenerateKey()
+	require.NoError(t, err)
+
+	signer12 := types.LatestSignerForChainID(big.NewInt(12))
+	signer14 := types.LatestSignerForChainID(big.NewInt(14))
+	signer16 := types.LatestSignerForChainID(big.NewInt(16))
```
LuisPH3, line 543: 👍

LuisPH3 (approved):

gossip/blockproc/bundle/validate.go
HerbertJordan: The current plan is to counter this by the envelope to require the sum of the gas as if all transactions are processed in the order of reference. This puts an indirect limit on maximum reuse.

LuisPH3: yes, the added gas must account for repeated txs. alrighty

PR #865, opened 2026-04-19
Add checked Uint64 arithmetic utility library

HerbertJordan:
Adds a utility library for simplifying handling overflows when performing calculations on `uint64` values.

Especially when dealing with gas cost computations, overflow issues need to be considered. Checking for overflows after every single step can be cumbersome. With this utility library, overflow checks can be conducted using a shared library, improving the readability of the resulting code.

See #868 for example uses of this utility.

### DIFF
--- a/utils/checked/uint64.go
+++ b/utils/checked/uint64.go
@@ -1,0 +1,93 @@
+package checked
+
+import (
+	"math"
+
+	"github.com/0xsoniclabs/carmen/go/common"
+)
+
+// ErrOverflow is returned when unwrapping a checked Uint64 that has overflowed.
+const ErrOverflow = common.ConstError("arithmetic overflow")
+
+type checkedUint64 struct {
+	value    uint64
+	overflow bool
+}
+
+// Uint64 creates a new checked Uint64 with the given value.
+func Uint64(value uint64) checkedUint64 {
+	return checkedUint64{value: value}
+}
+
+// Overflow creates a new checked Uint64 that has overflowed.
+func Overflow() checkedUint64 {
+	return checkedUint64{overflow: true}
+}
+
+// IsOverflown returns true if the checked Uint64 has overflowed, and false otherwise.
+func (s checkedUint64) IsOverflown() bool {
+	return s.overflow
+}
+
+// Unwrap returns the value of the checked Uint64 if it has not overflowed, and
+// an [ErrOverflow] otherwise.
+func (s checkedUint64) Unwrap() (uint64, error) {
+	if s.overflow {
+		return 0, ErrOverflow
+	}
+	return s.value, nil
+}
+
+// Add returns the sum of a and b as a checked Uint64 that can be used for
+// further arithmetic operations. If the sum overflows, the returned checked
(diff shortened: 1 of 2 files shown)


utils/checked/uint64.go
```diff
@@ -0,0 +1,109 @@
...
+
+// ErrOverflow is returned when unwrapping a U64 that has overflowed.
+const ErrOverflow = common.ConstError("arithmetic overflow")
+
+type U64 struct {
+	value    uint64
+	overflow bool
+}
+
+// Uint64 creates a new checked Uint64 with the given value.
```
LorenzSchueler:
```suggestion
// Uint64 creates a new checked uint64 with the given value.
```
You are still referring to `Uint64` in multiple places. I'm not sure whether you meant to write `checked uint64` or `U64`.

utils/checked/uint64.go
```diff
@@ -0,0 +1,109 @@
...
+	if x.overflow || y.overflow {
+		return Overflow()
+	}
+	if y.value != 0 && x.value > math.MaxUint64/y.value {
+		return Overflow()
+	}
+	return Uint64(x.value * y.value)
+}
+
+type arg interface {
```
LorenzSchueler: nit:
```suggestion
type unsignedInt interface {
```
?

LorenzSchueler (dismissed):

LorenzSchueler (approved):

utils/checked/uint64.go
```diff
@@ -0,0 +1,109 @@
...
+// further arithmetic operations. If the product overflows, the returned checked
+// U64 will be in an overflowed state. If any of the inputs is in an
+// overflowed state, the result will also be in an overflowed state.
+func Mul[A, B unsignedInt](a A, b B) U64 {
+	x := from(a)
+	y := from(b)
+	if x.overflow || y.overflow {
+		return Overflow()
+	}
+	if y.value != 0 && x.value > math.MaxUint64/y.value {
```
simonlechner, line 80: This check does not detect all overflows due to inaccuracy of integer division.
See an example with a prime number [here](https://goplay.tools/snippet/tMSr2_xpPkq).

simonlechner (changes requested):
The code and testing look good, but the multiplication requires a different overflow check.

utils/checked/uint64.go
HerbertJordan: Interesting ... let me check this out ...

HerbertJordan: @simonlechner in your example you define
```
f = 18_446_744_073_709_551_557
a = MaxUint64 / f = 1
b = f + 1 = 18_446_744_073_709_551_558
```
and it looks like you would expect an overflow for `a * b`?

But since `a = 1`, why should this overflow?

simonlechner: Right, my example was not chosen well and I had an error in my thought. The division is on the right side of the ">", therefore the rounding error of the integer division is not an issue.

simonlechner (approved):

PR #866, opened 2026-04-20
Geth dependency: Add list of function calls

simonlechner:
This PR adds a list of all freestanding geth functions called in sonic. Two of those are also tested in newly added unit tests.


simonlechner:
This way of testing dependencies is not feasible.

PR #867, opened 2026-04-20
Prevent bundleGetInfoFrom from returning non-existing block numbers

LuisPH3:
The changes here solve an issue manifested as flaky tests:  the reported block number where a bundle has executed may not yet be completed and therefore not found if queried. 

The solution proposed is to report unknown bundle, and let user poll again.

### DIFF
--- a/api/sonicapi/get_bundle_info.go
+++ b/api/sonicapi/get_bundle_info.go
@@ -50,20 +50,19 @@ func (a *PublicBundleAPI) GetBundleInfo(
 		return nil, nil
 	}
 
-	result := &RPCBundleInfo{
-		Block:    rpc.BlockNumber(info.BlockNumber),
-		Position: hexutil.Uint(info.Position.Offset),
-		Count:    hexutil.Uint(info.Position.Count),
-	}
-
-	if block, err := a.b.BlockByNumber(ctx, result.Block); err != nil || block == nil {
+	blockNumber := rpc.BlockNumber(info.BlockNumber)
+	if block, err := a.b.BlockByNumber(ctx, blockNumber); err != nil || block == nil {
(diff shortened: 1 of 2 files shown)


api/sonicapi/get_bundle_info.go
```diff
@@ -45,16 +45,25 @@ func (a *PublicBundleAPI) GetBundleInfo(
...
-			Count:    hexutil.Uint(info.Position.Count),
-		}, nil
+	if info == nil {
+		// bundle is unknown,
+		return nil, nil
 	}
 
-	// Otherwise, the state is unknown (default).
-	return nil, nil
+	result := &RPCBundleInfo{
```
facuMH: `result` could be declared after the block is confirmed to be finished. wdyt?

facuMH (dismissed):
just an optional comment. otherwise, fix looks good.

api/sonicapi/get_bundle_info.go
```diff
@@ -45,16 +45,25 @@ func (a *PublicBundleAPI) GetBundleInfo(
...
+		Block:    rpc.BlockNumber(info.BlockNumber),
+		Position: hexutil.Uint(info.Position.Offset),
+		Count:    hexutil.Uint(info.Position.Count),
+	}
+
+	if block, err := a.b.BlockByNumber(ctx, result.Block); err != nil || block == nil {
+		// althoug store has been notified about the bundle execution, the block is
```
HerbertJordan:
```suggestion
		// although the store has been notified about the bundle execution, the block is
```

HerbertJordan (dismissed):

facuMH (dismissed):

facuMH (approved):

PR #868, opened 2026-04-20
Factor out envelope gas limit calculation

HerbertJordan:
This PR implements a utility to compute the gas limit of envelopes.

The utility is to be used by the envelope validation and the builder to compute the expected gas limit for envelopes.

It is based on #865, which needs to be merged first.

### DIFF
--- a/gossip/blockproc/bundle/gas_test.go
+++ b/gossip/blockproc/bundle/gas_test.go
@@ -96,7 +96,7 @@ func Test_calculateEnvelopeGas_ComputesGasBasedOnProvidedParameters(t *testing.T
 				uint64(len(tc.authList)),
 			)
 			require.NoError(err)
-			require.Equal(int(tc.expected), int(want))
+			require.Equal(tc.expected, want)
 
 			// test that function produces expected value
 			have, err := calculateEnvelopeGas(
@@ -106,7 +106,7 @@ func Test_calculateEnvelopeGas_ComputesGasBasedOnProvidedParameters(t *testing.T
 				tc.authList,
 			)
 			require.NoError(err)
-			require.Equal(int(tc.expected), int(have))
+			require.Equal(tc.expected, have)
 		})
 	}
 }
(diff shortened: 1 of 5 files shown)


gossip/blockproc/bundle/gas_test.go
```diff
@@ -0,0 +1,607 @@
...
+				require.Greater(intrinsic, floorDataGas)
+				require.Greater(intrinsic, txGasSum)
+			case "floor":
+				require.Greater(floorDataGas, intrinsic)
+				require.Greater(floorDataGas, txGasSum)
+			case "transactions":
+				require.Greater(txGasSum, floorDataGas)
+				require.Greater(txGasSum, intrinsic)
+			default:
+				require.True(false, "unsupported test case spec")
```
simonlechner: This is a bit odd, consider `require.FailNow` or simply `t.Fatal`.

LuisPH3 (dismissed):
The

gossip/blockproc/bundle/gas_test.go
```diff
@@ -0,0 +1,607 @@
...
+		},
+		"big overflow from non-zero bytes": {
+			numZeroBytes:    0,
+			numNonZeroBytes: math.MaxUint64,
+		},
+		"close overflow from non-zero bytes": {
+			numZeroBytes:    0,
+			numNonZeroBytes: (math.MaxUint64-params.TxGas)/params.TxCostFloorPerToken/params.TxTokenPerNonZeroByte + 1,
+		},
+		"overflow from combined zero and non-zero bytes": {
```
simonlechner, line 512: Each of those inputs already produces an overflow on their own.

simonlechner (dismissed):
Looks good, testing covers all practical cases I can think of.

facuMH (dismissed):

gossip/blockproc/bundle/gas_test.go
HerbertJordan: Fixed in cb423a1e242c27f38e51edcbb43381db1bc0debe

gossip/blockproc/bundle/gas_test.go
HerbertJordan: Fixed in cb423a1e242c27f38e51edcbb43381db1bc0debe

simonlechner (approved):

PR #870, opened 2026-04-20
Test execution plans can be processed only once

LorenzSchueler:
This PR adds tests that ensure that the same execution plan is only processed once, whether in the same transaction as nested bundles, or in separate envelopes. It also adds a test to make sure that groups can be executed multiple times, since they belong to the same plan.

### DIFF
--- a/gossip/blockproc/bundle/builder.go
+++ b/gossip/blockproc/bundle/builder.go
@@ -497,3 +497,11 @@ func getGasLimitForEnvelope(
 
 	return max(intrinsic, floorDataGas, txGasSum)
 }
+
+func MustWrapIntoEnvelope(signer types.Signer, bundle *TransactionBundle) *types.Transaction {
+	key, err := crypto.GenerateKey()
+	if err != nil {
+		panic(fmt.Sprintf("failed to generate new key: %v", err))
+	}
+	return newEnvelope(signer, key, 0, &big.Int{}, bundle)
+}
--- a/tests/bundles/only_run_once_test.go
+++ b/tests/bundles/only_run_once_test.go
@@ -1,0 +1,239 @@
+// Copyright 2026 Sonic Operations Ltd
+// This file is part of the Sonic Client
+//
+// Sonic is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Sonic is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with Sonic. If not, see <http://www.gnu.org/licenses/>.
+
+package bundles
+
+import (
+	"math/big"
+	"testing"
+
+	"github.com/0xsoniclabs/sonic/gossip/blockproc/bundle"
+	"github.com/0xsoniclabs/sonic/opera"
+	"github.com/0xsoniclabs/sonic/tests"
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBundle_RunOnlyOnce_ExecutionPlanSubmittedMultipleTimesInDifferentEnvelopesIsOnlyProcessedOnce(t *testing.T) {
+	t.Parallel()
+	require := require.New(t)
+	upgrades := opera.GetBrioUpgrades()
+	upgrades.TransactionBundles = true
+
+	net := tests.StartIntegrationTestNet(t, tests.IntegrationTestNetOptions{
+		Upgrades: &upgrades,
+	})
+
+	client, err := net.GetClient()
+	require.NoError(err)
+	defer client.Close()
+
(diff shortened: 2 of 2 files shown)


gossip/blockproc/bundle/builder.go
```diff
@@ -490,3 +490,11 @@ func getGasLimitForEnvelope(
 
 	return max(intrinsic, floorDataGas, txGasSum)
 }
+
+func MustWrapIntoEnvelope(signer types.Signer, bundle *TransactionBundle) *types.Transaction {
```
LuisPH3, line 501: I would rather have a function `NewEnvelope(t testing.TB, signer, bundle)` in the testing namespace, this function does not serve any other purpose but testing and having `testing.TB` in the arguments makes sure that this not misused. 

Then you can fail and get nice formatted error reports, instead of a panic.

tests/bundles/only_run_once_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+	client, err := net.GetClient()
+	require.NoError(err)
+	defer client.Close()
+
+	signer := types.LatestSignerForChainID(net.GetChainId())
+
+	senders := tests.MakeAccountsWithBalance(t, net, 2, big.NewInt(1e18))
+
+	successfulTx := types.AccessListTx{}
```
LuisPH3: This transaction cannot be successful. You can spare the variable, use a literal in the builder as it will modify the transaction and make it successful.

(implicit behavior is wicked)

tests/bundles/only_run_once_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+	client, err := net.GetClient()
+	require.NoError(err)
+	defer client.Close()
+
+	signer := types.LatestSignerForChainID(net.GetChainId())
+
+	senders := tests.MakeAccountsWithBalance(t, net, 4, big.NewInt(1e18))
+
+	successfulTx := types.AccessListTx{}
```
LuisPH3: idem

tests/bundles/only_run_once_test.go
```diff
@@ -0,0 +1,239 @@
...
+	receiptA, err := net.GetReceipt(bundleTxs[0].Hash())
+	require.NoError(err)
+	require.Equal(types.ReceiptStatusSuccessful, receiptA.Status)
+
+	// Transaction B should not be executed.
+	receiptB, err := client.TransactionReceipt(t.Context(), bundleTxs[1].Hash())
+	require.ErrorIs(err, ethereum.NotFound, "Got receipt A: %+v, receipt B: %+v", receiptA, receiptB)
+}
+
+func TestBundle_RunOnlyOnce_ExecutionPlanSubmittedMultipleTimesInSameBundleIsOnlyProcessedOnce(t *testing.T) {
```
LuisPH3: please rename to include "nested bundle" so that is straight forward what is this about.

tests/bundles/only_run_once_test.go
```diff
@@ -0,0 +1,239 @@
...
+	defer client.Close()
+
+	signer := types.LatestSignerForChainID(net.GetChainId())
+
+	sender := tests.MakeAccountWithBalance(t, net, big.NewInt(1e18))
+
+	// This test creates a single envelope with a transaction bundle of
+	// the following shape:
+	//
+	//       AllOf([TolerateFailed]OneOf(A1), A0, OneOf(A1))
```
LuisPH3, line 183: A1 and A0 in this order is very difficult to follow, please use A, B, C

tests/bundles/only_run_once_test.go
```diff
@@ -0,0 +1,239 @@
...
+	require.Equal(types.ReceiptStatusSuccessful, receiptA.Status)
+	require.EqualValues(outerInfo.Block, receiptA.BlockNumber.Uint64())
+	require.EqualValues(outerInfo.Position, receiptA.TransactionIndex)
+
+	// Transaction B should not be executed.
+	receiptB, err := client.TransactionReceipt(t.Context(), bundleTxs[1].Hash())
+	require.ErrorIs(err, ethereum.NotFound, "Got receipt A: %+v, receipt B: %+v", receiptA, receiptB)
+}
+
+func TestBundle_RunOnlyOnce_FailedGroupsCanBeRetried(t *testing.T) {
```
LuisPH3, line 162: Valid test, just not sure what does it have to do with runOnlyOnce

LuisPH3 (changes requested):

HerbertJordan (dismissed):

gossip/blockproc/bundle/builder.go
LorenzSchueler: The problem is that `newEnvelope` which gets called here is private. We could make it public though.

tests/bundles/only_run_once_test.go
LorenzSchueler: Well the numbers are referring to the used nonces, which makes it IMO easier to follow.

tests/bundles/only_run_once_test.go
LorenzSchueler: Yeah, this evolved from a test in the prototype which used nested bundles but does not pass on main. The idea behind leaving it in this file is that we are testing that (nested) bundles are executed only once, but that subgroups/layers can be executed multiple times.
I can also move it somewhere else if you want.

tests/bundles/only_run_once_test.go
LuisPH3: okay, it took me quite some time to figure that out. but ok

gossip/blockproc/bundle/builder.go
LuisPH3: it's ok.

LuisPH3 (approved):

PR #871, opened 2026-04-20
Test empty bundles are rejected 

LorenzSchueler:
This PR adds tests to ensure that empty bundles are rejected.
This is necessary, because otherwise it would be possible to use up the gas limit of a block by sending in a bunch of bundles with different group structures (and therefore different hashes) which do not contain any transactions.

### DIFF
--- a/tests/bundles/empty_bundle_test.go
+++ b/tests/bundles/empty_bundle_test.go
@@ -23,21 +23,13 @@ import (
 	"time"
 
 	"github.com/0xsoniclabs/sonic/gossip/blockproc/bundle"
-	"github.com/0xsoniclabs/sonic/opera"
 	"github.com/0xsoniclabs/sonic/tests"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/stretchr/testify/require"
 )
 
 func TestBundle_BundleContainingAnyEmptyGroupIsRejected(t *testing.T) {
-	t.Parallel()
-
-	upgrades := opera.GetBrioUpgrades()
-	upgrades.TransactionBundles = true
-
-	net := tests.StartIntegrationTestNet(t, tests.IntegrationTestNetOptions{
-		Upgrades: &upgrades,
-	})
+	net := GetIntegrationTestNetWithBundlesEnabled(t)
 
 	client, err := net.GetClient()
 	require.NoError(t, err)
@@ -55,10 +47,7 @@ func TestBundle_BundleContainingAnyEmptyGroupIsRejected(t *testing.T) {
 	}{
 		"AllOf/NonEmpty": {
 			root: bundle.AllOf(
-				bundle.Step(
-					senders[0].PrivateKey,
-					tests.SetTransactionDefaults(t, net, &tx, senders[0]),
-				),
+				Step(t, net, senders[0], &tx),
 			),
 			ExpectReject: false,
 		},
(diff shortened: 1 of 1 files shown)


tests/bundles/empty_bundle_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	"github.com/0xsoniclabs/sonic/gossip/blockproc/bundle"
+	"github.com/0xsoniclabs/sonic/opera"
+	"github.com/0xsoniclabs/sonic/tests"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBundle_BundleContainingAnyEmptyGroupIsRejected(t *testing.T) {
+	t.Parallel()
```
LuisPH3: no parallel with full instances please.

tests/bundles/empty_bundle_test.go
```diff
@@ -0,0 +1,155 @@
...
+			require.NoError(t, err)
+
+			envelope, bundle, plan := bundle.NewBuilder().
+				WithSigner(signer).
+				SetEarliest(blockNumber).
+				With(c.root).
+				BuildEnvelopeBundleAndPlan()
+
+			// Send the bundle.
+			require.NoError(t, client.SendTransaction(t.Context(), envelope))
```
LuisPH3: I expect them to fail validation and not to be accepted into the pool. 

The test still makes sense though, but the tx pool validation needs to be disabled for them to be included in the pool.
A test with enabled validation (default)  needs to reject them right away.

tests/bundles/empty_bundle_test.go
```diff
@@ -0,0 +1,155 @@
...
+				require.ErrorIs(t, err, context.DeadlineExceeded)
+				return
+			}
+
+			require.NoError(t, err)
+
+			blockTxsHashes := getBlockTxsHashes(t, client, big.NewInt(info.Block.Int64()))
+			bundleTxs := bundle.GetTransactionsInReferencedOrder()
+
+			// Remove this once empty bundles are rejected.
```
LuisPH3: lets no merge this like this into main, 
option 1: add a TODO
option 2 (the better one) :  block this pr with the required changes/tickets.

LuisPH3 (changes requested):

HerbertJordan (commented):
I agree with @LuisPH3 , let's delay this test until empty-bundle filtering is implemented.

tests/bundles/empty_bundle_test.go
LuisPH3: because of tx validation coming, this stills need to be done.

tests/bundles/empty_bundle_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+			envelope, bundle, plan := bundle.NewBuilder().
+				WithSigner(signer).
+				SetEarliest(blockNumber).
+				With(c.root).
+				BuildEnvelopeBundleAndPlan()
+
+			// Send the bundle.
+			_, err = net.Send(envelope)
+			require.NoError(t, err)
```
HerbertJordan, line 115: Heads-up: When the validation at the intake point of the transaction pool is getting merged (the "validation" in the TxPool's Add function), this should start to return errors for empty bundles.

Do you want to deal with this already here, by accepting errors for those cases?

HerbertJordan (approved):
Test looks good, but it may get invalidated by an upcoming change. Do you want to make it already future proof or merge first, and depend on it to be fixed as the rejection is integrated?

tests/bundles/empty_bundle_test.go
LorenzSchueler: But pool verification is already disabled.

LuisPH3 (approved):

PR #872, opened 2026-04-20
Add validation of envelopes

HerbertJordan:
This PR finalizes the bundle validation by providing a validation step for envelopes.

### DIFF
--- a/gossip/blockproc/bundle/validate_test.go
+++ b/gossip/blockproc/bundle/validate_test.go
@@ -71,21 +71,19 @@ func TestValidateEnvelope_ValidEnvelopes_AreAcceptedAndReturnsBundleAndPlan(t *t
 		t.Run(name, func(t *testing.T) {
 			require := require.New(t)
 			signer := types.LatestSignerForChainID(big.NewInt(1))
-			envelope := tc.WithSigner(signer).Build()
+			envelope, wantBundle, wantPlan := tc.WithSigner(signer).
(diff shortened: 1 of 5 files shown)


facuMH (dismissed):

gossip/blockproc/bundle/validate_test.go
```diff
@@ -17,370 +17,309 @@
...
+			signer := types.LatestSignerForChainID(big.NewInt(1))
+			envelope := tc.WithSigner(signer).Build()
+
+			bundle, plan, err := ValidateEnvelope(signer, envelope)
 			require.NoError(err)
-			require.NotNil(bundle, "expected a bundle transaction")
-			require.NotNil(plan, "expected an execution plan")
 
-			wantedBundle, err := OpenEnvelope(signer, tx)
+			want, err := OpenEnvelope(signer, envelope)
```
LuisPH3: Since validate envelope uses OpenEnvelope internally, i think it may be better to compare with the input:
- tests are defined using a builder instead of a formed tx. the test can then create the envelope and plan and compare them

LuisPH3 (dismissed):

gossip/blockproc/bundle/validate_test.go
HerbertJordan: Fixed in 65df2748e2ec9873012486231e4b1b6a7a1ebc06

LuisPH3 (approved):

PR #873, opened 2026-04-20
Make default signer of builder accessible

HerbertJordan:
This PR adds support to the builder to retrieve the default signer through
```Go
signer := NewBuilder().GetSigner()
```
if needed. 

This helps eliminating the dependency to the global `testChainID` shared throughout the `bundle` package tests.


LuisPH3 (approved):
This is strange, I have recollection of adding this function myself.

PR #874, opened 2026-04-21
Fix baseFee in transaction tracing 

jenikd:
This PR fixes setting for a setup of an EVM when replaying transactions with transaction tracing. It must use `noBaseFee` when replaying transaction to skip gas preCheck to be able to replay internal transaction.
This was introduced with a refactor and reuse of the EVM setup for `stateDiff` in transaction tracing.

Test will be in separate PR using modification to the RPC test framework


LuisPH3 (approved):
This PR changes  the last flag to the evm initialization tool from false (with base fee ) to true (no base fee)
Additionally, adds an early return in `trace_logger.go` to prevent overruns. 
Looking forward to see the tests.

PR #875, opened 2026-04-21
Update rpc test backend

jenikd:
This PR adds transactions, receipts and basefee header into rpc test backend framework

### DIFF
--- a/api/rpc_test/backend.go
+++ b/api/rpc_test/backend.go
@@ -66,7 +66,17 @@ type Block struct {
 	// PrevRandao is the previous block's RANDAO value.
 	// If non zero, it enables Shanghai, Cancun, Prague, and Osaka features
 	// depending on the chain config.
-	PrevRandao common.Hash
+	PrevRandao   common.Hash
+	BaseFee      *big.Int
+	Transactions map[common.Hash]*Transaction
+}
+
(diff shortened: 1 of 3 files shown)


api/rpc_test/backend.go
```diff
@@ -384,6 +387,65 @@ func (b *fakeBackend) GetBundleExecutionInfo(_ common.Hash) *bundle.ExecutionInf
 	panic("not implemented")
 }
 
+// GetTransaction looks up a transaction by its hash in the fake backend's block history,
+// returning the transaction, the block number it was included in, and its index within that block.
+//
+// If the transaction is not found in any block, it returns an error.
```
LuisPH3, line 393: i think that the real backend returns non-error nil if not found:
https://github.com/0xsoniclabs/sonic/blob/dce17a7dde6d7d39ad7e613c8878fec5664fb118/gossip/ethapi_backend.go#L460

returning othewise would compromise the validity of our testing.

LuisPH3 (changes requested):

api/rpc_test/backend.go
jenikd: Fixed

LuisPH3 (approved):

PR #877, opened 2026-04-21
Add test for the trace namespace and a base fee parameter for a tx replay

jenikd:
This PR adds test for a case, where there should be no base fee checks for replying transactions inside `trace` namespace for functions like `trace_transaction` and `trace_block`. If the transaction has no fees and tips set, like internal transactions have, it was not possible to replay them as the base fee validation returned error in the state transition pre check, before EVM execution.

Test for https://github.com/0xsoniclabs/sonic/pull/874

### DIFF
--- a/api/rpc_test/trace_test.go
+++ b/api/rpc_test/trace_test.go
@@ -23,6 +23,8 @@ import (
 	"github.com/0xsoniclabs/sonic/api/ethapi"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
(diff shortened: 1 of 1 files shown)


api/rpc_test/trace_test.go
```diff
@@ -97,3 +99,70 @@ func Test_TraceSimpleTransfer(t *testing.T) {
...
+			blockNumber: 2,
+			txIndex:     0,
+			receipt:     receipt,
+		},
+	}
+
+	be := NewBackendBuilder(t).
+		WithChainID(chainId).
+		WithAccount(sender, AccountState{Balance: big.NewInt(1e18)}).
+		WithAccount(toAddr, AccountState{}).
```
LuisPH3: Do we need to initialize this account?

LuisPH3 (approved):
This PR adds a regression test for an specific issue.

LuisPH3 (approved):

PR #878, opened 2026-04-21
Add bundle stress tests

LorenzSchueler:
This PR adds two stress tests for bundles:
- submitting multiple bundles which are blocked by the nonce which only get unblocked by the very last bundle sent in
- submitting multiple bundles which have an inner group that is expensive (computationally or Db accesses) and rolls back in an outer group that does little work and succeeds, so the work during processing vs work of tx in block ratio is high.

The parameters for the size of these tests have been set to small values to speed up CI. For experimentation, these values should be set to larger values.

### DIFF
--- a/tests/bundles/stress_test.go
+++ b/tests/bundles/stress_test.go
@@ -64,12 +64,10 @@ func TestBundle_StressWithManyNonceBlockedBundles(t *testing.T) {
 		planHashes[i] = plan.Hash()
 	}
 
-	// Iteratively send in all bundles except the first one (with nonce 0)
-	// which will be blocked until the transaction with nonce 0 is executed.
-	for i := range N {
-		_, err = net.Send(envelopes[i+1])
-		require.NoError(t, err)
-	}
+	// Send in all bundles except the first one (with nonce 0) which will be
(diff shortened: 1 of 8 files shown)


LuisPH3 (dismissed):

tests/bundles/stress_test.go
```diff
@@ -0,0 +1,151 @@
...
+		bundles[i] = bundle
+		planHashes[i] = plan.Hash()
+	}
+
+	// Iteratively send in all bundles except the first one (with nonce 0)
+	// which will be blocked until the transaction with nonce 0 is executed.
+	for i := range N {
+		_, err = net.Send(envelopes[i+1])
+		require.NoError(t, err)
+	}
```
HerbertJordan: nit: maybe use the parallel submission?
```suggestion
	_, err = net.SendAll(envelopes[1:])
	require.NoError(t, err)
```

HerbertJordan (dismissed):

LorenzSchueler:
@HerbertJordan @LuisPH3 I've rewritten the second test. I passes with @LuisPH3's changes.

LuisPH3 (approved):

HerbertJordan (approved):

PR #880, opened 2026-04-21
Fix baseFee in transaction tracing  (#874)

jenikd:
* Always use noBaseFee when replaying transaction

* Fix error logging

* Add length check for actions


LuisPH3 (approved):
The patch introduces the changes merged into main

wsodsong (approved):

PR #881, opened 2026-04-21
Factor out processor setup into utility function

HerbertJordan:
This PR factors out the setup of a step-by-step transaction processor from the single proposer scheduling code to facilitate reuse in the bundle trial-running tests added by #882.

This PR does not add new features. It mainly moves code and an affiliated interface from `emitter/scheduler/processor.go` to the `NewTransactionProcessorForBlock` function in `evmcore/state_processor.go`. The rest is resolving a name conflict (the `ChainState` for the bundle eval is renamed to `ChainStateForBundleEval`).


facuMH (approved):
This PR moves decataration of `ChainState` interface in the scheduler/processor to evmcore so that it can also be used by the bundle_precheck.

LuisPH3 (approved):

PR #883, opened 2026-04-22
Refactor integration tests to account for tx pool filtering non-executable bundles. 

LuisPH3:
Some tests relied on the capability of non-executable bundles reaching the block processor. 
This PR attempts to rewrite the tests to keep testing the desired side effects, but contemplating that non-executable bundles will not reach the consensus layer.

### DIFF
--- a/tests/bundles/only_run_once_test.go
+++ b/tests/bundles/only_run_once_test.go
@@ -79,10 +79,16 @@ func TestBundle_RunOnlyOnce_ExecutionPlanSubmittedMultipleTimesInDifferentEnvelo
 	}
 
 	// Submit the same bundle multiple times using different envelopes.
-	sentHashes, _, err := net.TrySendAll(envelopes)
+	sentHashes, rejected, err := net.TrySendAll(envelopes)
 	require.NoError(err)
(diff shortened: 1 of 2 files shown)


tests/bundles/only_run_once_test.go
```diff
@@ -35,9 +35,13 @@ func TestBundle_RunOnlyOnce_ExecutionPlanSubmittedMultipleTimesInDifferentEnvelo
...
-	net := tests.StartIntegrationTestNet(t, tests.IntegrationTestNetOptions{
-		Upgrades: &upgrades,
-	})
+	net := tests.StartIntegrationTestNet(t,
+		tests.IntegrationTestNetOptions{
+			Upgrades: &upgrades,
+			ClientExtraArguments: []string{
+				"--disable-txPool-validation",
+			},
+		})
```
LorenzSchueler: As we talked about in person, this is not strictly necessary but it increases the chance of multiple envelopes making it to the state processor. Please add a comment for that. Otherwise, it looks like without disabling validation, this will never happen.

LorenzSchueler (dismissed):

LorenzSchueler (approved):

PR #884, opened 2026-04-22
Refactor tx list to reflect tx transitions regarding bundles executability. 

LuisPH3:
This PR introduces bundle envelope awareness into the transaction pool's core list management (txList), enabling correct promotion and demotion of bundle transactions between the queued and pending sets.

**Background**
The transaction pool maintains two sets per account:

**Pending**: transactions eligible for inclusion in the next block, with strictly sequential nonces.
**Queued**: transactions not yet promotable (e.g. future nonces, insufficient balance).

Two operations drive transitions between these sets:
**Ready** (queued → pending): called during promoteExecutables, pops a contiguous run of transactions starting at the expected nonce and moves them into pending.
**Filter** (pending → queued / removed, or queued → removed): called during demoteUnexecutables, inspects every transaction in the pending list and either keeps it, drops it permanently, or demotes it back to queued.

Transaction list has two variants, strict (for the pending list) and non-strict (for the queued list). The first variant allows to invalidate transactions previously pending when one tx is removed, previous implementation will demote all gapped nonces, the new semantics contemplate the possibility of bundles not being executable anymore, therefore making the transition to the pool to spare resources in the emitter.

### DIFF
--- a/evmcore/tx_list.go
+++ b/evmcore/tx_list.go
@@ -195,9 +195,9 @@ func (m *txSortedMap) Remove(nonce uint64) bool {
 	return true
 }
 
-// Ready pops a contiguous run of transactions whose nonces begin at start and
-// are sequentially increasing, returning them for promotion from the queued
-// list into the pending list. Each candidate is passed through isExecutable;
+// Ready pops a contiguous list of transactions with incremental nonces,
+// returning them for promotion from the queued list into the pending list.
+// Each candidate is passed through isExecutable;
 // the run stops at the first transaction that returns false. This gate is
 // used to keep bundle envelope transactions in the queue when a trial-run
 // against the current state shows they are not yet executable (e.g. an
@@ -426,26 +426,25 @@ func (l *txList) Filter(
 	// If the list was strict, filter anything above the lowest nonce
 	var invalids types.Transactions
 	if l.strict {
-		maxExecutableNonce := uint64(math.MaxUint64)
+		firstInvalidNonce := uint64(math.MaxUint64)
 		for _, tx := range removed {
-			maxExecutableNonce = min(maxExecutableNonce, tx.Nonce())
+			firstInvalidNonce = min(firstInvalidNonce, tx.Nonce())
 		}
 
-		// invalidate any pending bundle with nonce less than the
-		// nonce of any removed transaction which is temporarily blocked
+		// bundles with valid nonces but non-pending status are also invalidated
 		invalidBundles := l.txs.filter(func(tx *types.Transaction) bool {
 			return bundle.IsEnvelope(tx) &&
-				tx.Nonce() < maxExecutableNonce &&
+				tx.Nonce() < firstInvalidNonce &&
 				evaluateBundleStatus(tx) == bundleQueued
 		})
 		for _, tx := range invalidBundles {
-			maxExecutableNonce = min(maxExecutableNonce, tx.Nonce())
+			firstInvalidNonce = min(firstInvalidNonce, tx.Nonce())
 		}
 
 		// invalidate any transaction with nonce greater than any
 		// removed transaction
 		invalids = l.txs.filter(func(tx *types.Transaction) bool {
-			return tx.Nonce() > maxExecutableNonce
+			return tx.Nonce() > firstInvalidNonce
 		})
 
 		// invalids are to be demoted from pending to queued, and therefore
--- a/evmcore/tx_list.go
+++ b/evmcore/tx_list.go
@@ -100,11 +100,9 @@ func (m *txSortedMap) Forward(threshold uint64) types.Transactions {
 	return removed
 }
 
-// Filter iterates over the list of transactions and removes all of them for which
-// the specified function evaluates to true.
-// Filter, as opposed to 'filter', re-initialises the heap after the operation is done.
-// If you want to do several consecutive filterings, it's therefore better to first
-// do a .filter(func1) followed by .Filter(func2) or reheap()
+// Filter removes every transaction for which the predicate returns true and
+// rebuilds the nonce heap afterwards. Use the unexported filter method
+// instead when chaining multiple passes, followed by a single reheap call.
 func (m *txSortedMap) Filter(filter func(*types.Transaction) bool) types.Transactions {
 	removed := m.filter(filter)
 	// If transactions were removed, the heap and cache are ruined
@@ -197,14 +195,21 @@ func (m *txSortedMap) Remove(nonce uint64) bool {
 	return true
 }
 
-// Ready retrieves a sequentially increasing list of transactions starting at the
-// provided nonce that is ready for processing. The returned transactions will be
-// removed from the list.
+// Ready pops a contiguous run of transactions whose nonces begin at start and
+// are sequentially increasing, returning them for promotion from the queued
+// list into the pending list. Each candidate is passed through isExecutable;
+// the run stops at the first transaction that returns false. This gate is
+// used to keep bundle envelope transactions in the queue when a trial-run
+// against the current state shows they are not yet executable (e.g. an
+// inner transaction's preconditions are not met). Regular and sponsored
+// transactions always pass the check.
 //
-// Note, all transactions with nonces lower than start will also be returned to
-// prevent getting into and invalid state. This is not something that should ever
-// happen but better to be self correcting than failing!
-func (m *txSortedMap) Ready(start uint64) types.Transactions {
+// Transactions with nonces below start are also popped to self-correct any
+// stale entries that should have been forwarded already.
+func (m *txSortedMap) Ready(
+	start uint64,
+	isExecutable func(*types.Transaction) bool,
+) types.Transactions {
 	// Short circuit if no transactions are available
 	if m.index.Len() == 0 || (*m.index)[0] > start {
 		return nil
@@ -212,7 +217,12 @@ func (m *txSortedMap) Ready(start uint64) types.Transactions {
 	// Otherwise start accumulating incremental transactions
 	var ready types.Transactions
 	for next := (*m.index)[0]; m.index.Len() > 0 && (*m.index)[0] == next; next++ {
-		ready = append(ready, m.items[next])
+		tx := m.items[next]
+		if !isExecutable(tx) {
+			break
+		}
+
+		ready = append(ready, tx)
 		delete(m.items, next)
 		heap.Pop(m.index)
 	}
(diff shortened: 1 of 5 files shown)


evmcore/tx_list.go
```diff
@@ -197,22 +195,34 @@ func (m *txSortedMap) Remove(nonce uint64) bool {
...
 		return nil
 	}
 	// Otherwise start accumulating incremental transactions
 	var ready types.Transactions
 	for next := (*m.index)[0]; m.index.Len() > 0 && (*m.index)[0] == next; next++ {
-		ready = append(ready, m.items[next])
+		tx := m.items[next]
+		if !isExecutable(tx) {
+			break
+		}
```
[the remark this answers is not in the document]
LuisPH3: fixed comment,

evmcore/tx_pool.go
```diff
@@ -1532,7 +1532,13 @@ func (pool *TxPool) promoteExecutables(accounts []common.Address) []*types.Trans
...
 		log.Trace("Removed old queued transactions", "count", len(forwards))
 		// Drop all transactions that are too costly (low balance or out of gas)
-		drops, _ := list.Filter(utils.Uint256ToBigInt(pool.currentState.GetBalance(addr)), pool.currentMaxGas, subsidiesChecker)
+		drops, _ := list.Filter(
+			utils.Uint256ToBigInt(pool.currentState.GetBalance(addr)),
+			pool.currentMaxGas,
+			subsidiesChecker,
+			// FIXME: replace with real evaluation
+			func(t *types.Transaction) bundlePoolStatus { return bundlePending },
+		)
```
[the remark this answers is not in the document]
LuisPH3: coming later

evmcore/tx_pool.go
```diff
@@ -1541,7 +1547,11 @@ func (pool *TxPool) promoteExecutables(accounts []common.Address) []*types.Trans
 		queuedNofundsMeter.Mark(int64(len(drops)))
 
 		// Gather all executable transactions and promote them
-		readies := list.Ready(pool.pendingNonces.get(addr))
+		readies := list.Ready(
+			pool.pendingNonces.get(addr),
+			// FIXME: replace with real evaluation
+			func(t *types.Transaction) bool { return true },
+		)
```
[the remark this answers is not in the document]
LuisPH3: coming later

evmcore/tx_list.go
```diff
@@ -349,56 +359,103 @@ func (l *txList) Forward(threshold uint64) types.Transactions {
...
+		// Bundle transactions need to be checked before sponsored transactions
+		// since bundle envelopes may qualify as sponsored transactions.
 		if bundle.IsEnvelope(tx) {
-			// TODO: filter out permanently blocked bundles
-			return false
+			status := evaluateBundleStatus(tx)
+			if status != bundlePending {
+				bundleInvalidated = true
+			}
+			return status == bundleRejected
```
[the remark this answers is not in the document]
LuisPH3: cache coming later

evmcore/tx_list.go
```diff
@@ -349,56 +359,103 @@ func (l *txList) Forward(threshold uint64) types.Transactions {
...
+			maxExecutableNonce = min(maxExecutableNonce, tx.Nonce())
+		}
+
+		// invalidate any pending bundle with nonce less than the
+		// nonce of any removed transaction which is temporarily blocked
+		invalidBundles := l.txs.filter(func(tx *types.Transaction) bool {
+			return bundle.IsEnvelope(tx) &&
+				tx.Nonce() < maxExecutableNonce &&
+				evaluateBundleStatus(tx) == bundleQueued
+		})
```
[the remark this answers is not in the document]
LuisPH3: fixed comment

evmcore/tx_pool.go
```diff
@@ -1731,7 +1741,13 @@ func (pool *TxPool) demoteUnexecutables() {
...
 		}
 		// Drop all transactions that are too costly (low balance or out of gas), and queue any invalids back for later
-		drops, invalids := list.Filter(utils.Uint256ToBigInt(pool.currentState.GetBalance(addr)), pool.currentMaxGas, subsidiesChecker)
+		drops, invalids := list.Filter(
+			utils.Uint256ToBigInt(pool.currentState.GetBalance(addr)),
+			pool.currentMaxGas,
+			subsidiesChecker,
+			// FIXME: replace with real evaluation
+			func(t *types.Transaction) bundlePoolStatus { return bundlePending },
+		)
```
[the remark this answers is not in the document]
LuisPH3: coming later

evmcore/tx_list.go
```diff
@@ -349,56 +359,102 @@ func (l *txList) Forward(threshold uint64) types.Transactions {
...
+//   - Its cost exceeds costLimit and it is not a sponsored or bundle tx.
+//   - Its gas exceeds gasLimit and it is not a sponsored or bundle tx.
+//   - It is a sponsorship request whose sponsorship is no longer active
+//     (isSponsored returns false).
+//   - It is a bundle envelope whose status is bundleRejected (e.g. already
+//     executed or structurally invalid).
+//
+// In strict mode (pending list), after drops are removed, any remaining
+// transactions that would create a nonce gap are invalidated and returned
+// separately so the caller can demote them back to the queued list.
```
facuMH, line 374: does "return to the caller" implies they are removed from this list ?

evmcore/tx_list_test.go
```diff
@@ -109,6 +111,360 @@ func TestTxList_Filter_WithSponsoredTransactions_RetainsCovered(t *testing.T) {
...
+		return func(tx *types.Transaction) bundlePoolStatus {
+			if tx.Nonce() >= nonce {
+				return bundleRejected
+			}
+			return bundlePending
+		}
+	}
+
+	tests := map[string]struct {
+		strict          bool
```
facuMH, line 265: nit: rename `isStrict`

evmcore/tx_pool.go
```diff
@@ -1532,7 +1532,13 @@ func (pool *TxPool) promoteExecutables(accounts []common.Address) []*types.Trans
 		}
 		log.Trace("Removed old queued transactions", "count", len(forwards))
 		// Drop all transactions that are too costly (low balance or out of gas)
-		drops, _ := list.Filter(utils.Uint256ToBigInt(pool.currentState.GetBalance(addr)), pool.currentMaxGas, subsidiesChecker)
+		drops, _ := list.Filter(
+			utils.Uint256ToBigInt(pool.currentState.GetBalance(addr)),
+			pool.currentMaxGas,
+			subsidiesChecker,
+			// FIXME: replace with real evaluation
```
facuMH, line 1542: This will be integrated by a coming PR, right ?

evmcore/tx_pool_test.go
```diff
@@ -339,6 +340,18 @@ func pricedSetCodeTxWithAuth(nonce uint64, gaslimit uint64, gasFee, tip *uint256
 	})
 }
 
+// bundleTx creates a transaction that is part of a bundle with the given key and nonce.
+// the name follows the convention of other transaction creation tools in the tx pool tests.
+func bundleTx(nonce uint64, key *ecdsa.PrivateKey) *types.Transaction {
```
facuMH, line 345: is this helper function planned to be used as part of a coming PR? if not, move it to `tx_list_test.go`.

facuMH (approved):
This PR extends the promot/demote mechanisms to properly handle bundles. Additionally improves on the existing documentation and testing.

evmcore/tx_list.go
LuisPH3: // Filter removes every transaction for which the predicate returns true

HerbertJordan (commented):
Please re-add me once the changes we discussed offline have been added.

evmcore/tx_list.go
```diff
@@ -197,22 +195,34 @@ func (m *txSortedMap) Remove(nonce uint64) bool {
...
-// provided nonce that is ready for processing. The returned transactions will be
-// removed from the list.
+// Ready pops a contiguous list of transactions with incremental nonces,
+// returning them for promotion from the queued list into the pending list.
+// Each candidate is passed through isExecutable;
+// the run stops at the first transaction that returns false. This gate is
+// used to keep bundle envelope transactions in the queue when a trial-run
+// against the current state shows they are not yet executable (e.g. an
+// inner transaction's preconditions are not met). Regular and sponsored
+// transactions always pass the check.
```
HerbertJordan: I think this documentation is irrelevant for this function. This function is not aware of envelopes, trial-runs, or sponsored transactions. All of this is hidden in the `isExecutable` function for the main use case.

evmcore/tx_list.go
```diff
@@ -349,56 +359,102 @@ func (l *txList) Forward(threshold uint64) types.Transactions {
 	return l.txs.Forward(threshold)
 }
 
-// Filter removes all transactions from the list with a cost or gas limit higher
-// than the provided thresholds. Every removed transaction is returned for any
-// post-removal maintenance. Strict-mode invalidated transactions are also
-// returned.
+// Filter inspects every transaction in the list and classifies it as dropped,
+// invalidated, or kept, based on the account's balance, the current gas limit,
+// the sponsorship status, and the bundle execution state.
```
HerbertJordan: This makes it sound like the Filter function is just producing data. It does not state, like the original version of the documentation, that those changes are actually applied. Transactions are actually removed, aren't day?

evmcore/tx_list.go
```diff
@@ -349,56 +359,102 @@ func (l *txList) Forward(threshold uint64) types.Transactions {
...
 	// Filter out all the transactions above the account's funds
 	removed := l.txs.Filter(func(tx *types.Transaction) bool {
+		// Bundle transactions need to be checked before sponsored transactions
+		// since bundle envelopes may qualify as sponsored transactions.
 		if bundle.IsEnvelope(tx) {
-			// TODO: filter out permanently blocked bundles
-			return false
+			status := evaluateBundleStatus(tx)
+			if status != bundlePending {
+				bundleInvalidated = true
```
HerbertJordan: bad name: the bundle is not invalid. Please rename to "containsBundlesToBeDemoted" or something like that.

evmcore/tx_list.go
```diff
@@ -349,56 +359,102 @@ func (l *txList) Forward(threshold uint64) types.Transactions {
...
+		firstInvalidNonce := uint64(math.MaxUint64)
 		for _, tx := range removed {
-			if nonce := tx.Nonce(); lowest > nonce {
-				lowest = nonce
-			}
+			firstInvalidNonce = min(firstInvalidNonce, tx.Nonce())
+		}
+
+		// bundles with valid nonces but non-pending status are also invalidated
+		invalidBundles := l.txs.filter(func(tx *types.Transaction) bool {
```
HerbertJordan: "invalidBundles" => "blockedBundles" ?

evmcore/tx_list.go
```diff
@@ -349,56 +359,102 @@ func (l *txList) Forward(threshold uint64) types.Transactions {
...
+		invalids = l.txs.filter(func(tx *types.Transaction) bool {
+			return tx.Nonce() > firstInvalidNonce
+		})
+
+		// invalids are to be demoted from pending to queued, and therefore
+		// are re-queued by the caller, who owns both tx lists.
+		invalids = append(invalidBundles, invalids...)
+
+		if len(invalids) > 0 {
+			l.txs.reheap()
```
HerbertJordan: This used to be also called in the non-strict mode. Is this intended to be only here? After all, the filter call in line 406 may have removed elements as well, which may require a reheap.

evmcore/tx_list.go
```diff
@@ -424,15 +480,22 @@ func (l *txList) Remove(tx *types.Transaction) (bool, types.Transactions) {
...
-// provided nonce that is ready for processing. The returned transactions will be
-// removed from the list.
+// Ready returns and removes a contiguous run of transactions starting at
+// start that are eligible for promotion from queued to pending. The
+// isExecutable callback gates each transaction: regular and sponsored
+// transactions always pass, while bundle envelopes are only promoted when
+// a trial-run against the current chain state confirms they are executable.
+// The run stops at the first transaction that fails the check, leaving it
+// and all higher-nonce transactions in the queue for a future promotion
+// pass.
```
HerbertJordan: This describes one use case, which from the point of this function, may only be one of many. It is confusing to read this here, since it could also imply that all of this is done by default, no matter the `isExecutable` is doing.

evmcore/tx_list_test.go
```diff
@@ -109,6 +111,360 @@ func TestTxList_Filter_WithSponsoredTransactions_RetainsCovered(t *testing.T) {
...
+	for i := 1; i < len(txs); i++ {
+		txs[i] = transaction(uint64(i), 0, key)
+	}
+
+	list := newTxList(true)
+	for _, tx := range txs {
+		list.Add(tx, DefaultTxPoolConfig.PriceBump)
+	}
+
+	// Each bundle transaction should be checked.
```
HerbertJordan:
```suggestion
	// Bundle transactions are to be demoted.
```

evmcore/tx_list_test.go
```diff
@@ -109,6 +111,360 @@ func TestTxList_Filter_WithSponsoredTransactions_RetainsCovered(t *testing.T) {
...
+	checker := func(tx *types.Transaction) bundlePoolStatus {
+		return bundleQueued
```
HerbertJordan, line 222: If you return Queued here, the bundle is to be removed for two reasons: it is to be demoted from pending to queued and there is a nonce gap. Do I read this correctly? Should this function return `bundlePending`?

HerbertJordan (changes requested):
The changes look like they manage to get the job done. However, I have a few documentation and variable naming concerns. Feel free to ignore those.

The tests could be stricter in checking properties, but I guess for the TxPool scope, and given the difficulty of dealing with legacy code, this is fine.

The main thing that should be addressed is the change in when the `reheap` function is called.

evmcore/tx_list.go
LuisPH3: tx_list returns dropped and invalidated. this is tx_list nomenglature

evmcore/tx_list.go
LuisPH3: i am collecting invalid transactions, this is the name of this kind of transactions in this scope.

evmcore/tx_list.go
LuisPH3: this is subtle:  `l.txs.Filter(` does reheap, while `l.txs.filter(` does not. 
my current reheap location tries to reproduced existing logic.

evmcore/tx_list_test.go
LuisPH3: this test requires the first transaction (bundle) to make the transition  `pending` -> `queued` If the bundle remains `bundlePending` it will not be demoted. 
Maybe the naming of the enum values are not clear?

evmcore/tx_list.go
LuisPH3: rewriten

evmcore/tx_list.go
LuisPH3: rewriten

evmcore/tx_list.go
LuisPH3: Rewrite, notice that the old comment was also wrong, as in some scenarios the list started with older nonces that start.

evmcore/tx_list_test.go
HerbertJordan: Alright, I get it now. There is only one bundle in this example, the one with nonce 0. The rest are regular transactions. And this filter is really just targeting the first transaction with nonce 0.

HerbertJordan (dismissed):

evmcore/tx_list.go
LuisPH3: original location restored.

HerbertJordan (approved):

PR #886, opened 2026-04-22
Keep ProcessedBundleHistoryHash Zero until the first bundle is executed

facuMH:
**Problem**
The history hash formula for processed bundles would change even before bundles where ever processed, which mean 2 nodes starting from different block heights would have different `ProcessedBundleHistoryHash` even if no bundle was ever processed -defeating the purpose of the hash as a cross-validator alignment check.

**Solution**
The `ProcessedBundleHistoryHash` is now held at zero until the first bundle is executed. Empty or nil bundle maps before that point are no-ops on the hash, so validators that haven't seen any bundles yet remain in a clean initial state.

Additionally:
The deleted hash has been deleted from the formula entirely. The new formula remains deterministic regardless of when a validator joined the network, as long as both start from the zero value for `ProcessedBundleHistoryHash`, which is the default. 
This will also serve for the purpose of validating the import of this table during genesis import. But that will be a follow-up task.

This PR implements https://github.com/0xsoniclabs/sonic-admin/issues/735

### DIFF
--- a/gossip/store_processed_bundles.go
+++ b/gossip/store_processed_bundles.go
@@ -47,10 +47,10 @@ import (
 //
 // The hash of the processed bundle's history is computed as follows:
 //  - initially, the hash is zero
-//  - for every update, the hash is updated as follows:
+//  - the zero hash is kept until the first executed bundle
+//  - thereafter, for every block, the hash is updated as follows:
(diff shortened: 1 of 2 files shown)


LorenzSchueler (approved):

HerbertJordan (approved):
Nice, well done!

PR #887, opened 2026-04-22
Complex bundle structure for prepare bundle RPC function

jenikd:
This PR introduces composable bundle execution plan in `sonic_preapreBundle` RPC function. To keep the preparation of the bundle simple for the user, input can be still a list of transactions inside the json object in the `steps` list. There can also be nested bundles with transactions.

Output of the `sonic_preapreBundle` is almost the same, list of transactions to be signed and a execution plan, which is now the tree structure of the steps.

There is still easy way, how to prepare bundles with this function, where none of these transaction can fail:
```AllOf(tx A, tx B)```

``` json
{
  "jsonrpc": "2.0",
  "method": "sonic_prepareBundle",
  "params": [
    {
      "steps": [
        {
          "from": "0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
          "to": "0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
          "nonce": "0x0",
          "value": "0xde0b6b3a7640000"
        },
        {
          "from": "0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
          "to": "0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
          "nonce": "0x0",
          "value": "0xde0b6b3a7640000"
        }
      ]
    }
  ],
  "id": 1
}
```

There is also option to pass complex structure of the bundle, example below represents this:
```
AllOf(tx1, OneOf(tx2, tx3))
```
  
``` json
{
  "jsonrpc": "2.0",
  "method": "sonic_prepareBundle",
  "params": [
    {
      "steps": [
        {
          "from": "0xaaa…",
          "to": "0xbbb…",
          "nonce": "0x0",
          "gas": "0x5208",
          "gasPrice": "0x3b9aca00",
          "tolerateFailed": true
        },
        {
          "oneOf": true,
          "tolerateFailures": true,
          "steps": [
            {
              "from": "0xbbb…",
              "to": "0xccc…",
              "nonce": "0x0",
              "gas": "0x5208",
              "gasPrice": "0x3b9aca00"
            },
            {
              "from": "0xbbb…",
              "to": "0xddd…",
              "nonce": "0x0",
              "gas": "0x5208",
              "gasPrice": "0x3b9aca00"
            }
          ]
        }
      ]
    }
  ],
  "id": 1
}
```


api/sonicapi/prepare_bundle.go
```diff
@@ -31,75 +34,113 @@ import (
...
+	}
+	_, hasTx := raw["transaction"]
+	_, hasEntries := raw["entries"]
+	switch {
+	case hasTx && !hasEntries:
+		s.Tx = new(PrepareBundleTxStep)
+		return json.Unmarshal(data, s.Tx)
+	case hasEntries && !hasTx:
+		s.Group = new(PrepareBundleGroup)
+		return json.Unmarshal(data, s.Group)
```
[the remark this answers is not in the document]
jenikd: I don't think, this is the case as the object unmarshal only once and is not used anywhere else

api/sonicapi/prepare_bundle.go
```diff
@@ -31,75 +34,113 @@ import (
...
+// The returned transactions must be signed without altering any fields.
 func (a *PublicBundleAPI) PrepareBundle(
 	ctx context.Context,
 	args PrepareBundleArgs,
 ) (*RPCPreparedBundle, error) {
 
+	if len(args.Transactions) > 0 && len(args.Entries) > 0 {
+		return nil, fmt.Errorf("cannot specify both 'transactions' and 'entries' in bundle input")
+	}
+	if len(args.Transactions) > 0 {
```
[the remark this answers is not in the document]
jenikd: I think using len() is safer here, because of also checks the empty arrays, which should not be passed into the function

api/sonicapi/prepare_bundle_test.go
```diff
@@ -447,30 +510,32 @@ func Test_PrepareBundle_SingleTx_GasAndPriceEstimated(t *testing.T) {
...
 
 	args := PrepareBundleArgs{
-		Transactions: []ethapi.TransactionArgs{
-			{From: &addr1, To: &addr2, Nonce: rpctest.ToHexUint64(0), Value: rpctest.ToHexBigInt(big.NewInt(1e15))},
-			{From: &addr2, To: &addr1, Nonce: rpctest.ToHexUint64(0), Value: rpctest.ToHexBigInt(big.NewInt(1e15))},
+		PrepareBundleGroup: PrepareBundleGroup{
+			Entries: []PrepareBundleEntry{
+				txEntry(ethapi.TransactionArgs{From: &addr1, To: &addr2, Nonce: rpctest.ToHexUint64(0), Value: rpctest.ToHexBigInt(big.NewInt(1e15))}),
+				txEntry(ethapi.TransactionArgs{From: &addr2, To: &addr1, Nonce: rpctest.ToHexUint64(0), Value: rpctest.ToHexBigInt(big.NewInt(1e15))}),
+			},
```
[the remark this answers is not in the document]
jenikd: Estimate gas is using shared state between these two transactions, so it is correct as the balnce for addr2 is initialized with first transaction

api/sonicapi/prepare_bundle.go
```diff
@@ -110,64 +151,145 @@ func (a *PublicBundleAPI) PrepareBundle(
...
+	builder := &prepareBundleBuilder{
+		flatTxArgs: flatTxs,
+		gasCap:     gasCap,
+		basefee:    basefee,
+		signer:     signer,
+	}
+	root, err := builder.buildGroupStep(&args.PrepareBundleGroup)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
jenikd: Added cursor validation

api/sonicapi/prepare_bundle.go
```diff
@@ -110,64 +151,145 @@ func (a *PublicBundleAPI) PrepareBundle(
...
+	basefee    *big.Int
+	signer     types.Signer
+}
+
+// buildStep builds an ExecutionStep from the current PrepareBundleEntry, dispatching to buildTxStep or buildGroupStep as appropriate.
+func (b *prepareBundleBuilder) buildStep(step PrepareBundleEntry) (bundle.ExecutionStep, error) {
+	if step.Tx != nil {
+		return b.buildTxStep(step.Tx)
+	}
+	return b.buildGroupStep(step.Group)
```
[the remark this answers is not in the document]
jenikd: This is not needed

api/sonicapi/prepare_bundle.go
```diff
@@ -110,64 +151,145 @@ func (a *PublicBundleAPI) PrepareBundle(
...
+}
+
+// buildTxStep converts the TransactionArgs at the current cursor position into a TxStep, then advances the cursor.
+func (b *prepareBundleBuilder) buildTxStep(tx *PrepareBundleTxStep) (bundle.ExecutionStep, error) {
+	idx := b.cursor
```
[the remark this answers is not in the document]
jenikd: Added this check

api/sonicapi/prepare_bundle.go
```diff
@@ -31,75 +34,113 @@ import (
...
-	// LatestBlock specifies the latest block number at which the bundle can be executed. This allows users
-	// to set an upper bound on when their bundle should be considered for execution, ensuring it is
-	// not included in blocks after a certain point in time. If the bundle is not executed by this block,
-	// it will be considered expired and will not be executed.
-	//
-	// If left unspecified, the bundle will be eligible for execution until 1024 blocks after EarliestBlock.
-	LatestBlock *hexutil.Uint64 `json:"latestBlock"`
+	PrepareBundleGroup
+	// Transactions is a flat all-of shorthand; mutually exclusive with Entries.
+	Transactions []ethapi.TransactionArgs `json:"transactions,omitempty"`
```
LuisPH3: As discussed offline, this could be
```
type PrepareBundleArgs struct {
      RPCExecutionProposal

      // extra flags regarding enabling gas estimation or other configurable behaviours of the prepare call.
}
```

api/sonicapi/prepare_bundle.go
```diff
@@ -110,64 +151,162 @@ func (a *PublicBundleAPI) PrepareBundle(
 	}
 
 	gasPrice := a.suggestGasPrice(currentBlock)
-	fillTransactionDefaults(args.Transactions, gasLimits, gasPrice)
+	fillTransactionDefaults(flatTxs, gasLimits, gasPrice)
```
LuisPH3: Shall we fill in defaults before estimating? Could this change estimation?

api/sonicapi/prepare_bundle.go
```diff
@@ -19,203 +19,185 @@ package sonicapi
...
 
-		steps[i] = bundle.NewTxStep(bundle.TxReference{
-			From: msg.From,
-			Hash: signer.Hash(tx),
-		})
+			if txArgs.Nonce == nil {
+				return step, fmt.Errorf("transaction %d is missing nonce", cursor)
+			}
+
+			if txArgs.To == nil {
```
LuisPH3: contract creation is technically prossible.

tests/rpcs/prepare_bundle_test.go
```diff
@@ -0,0 +1,416 @@
...
+}
+
+// TestPrepareBundle tests the sonic_prepareBundle RPC method.
+func TestPrepareBundle(t *testing.T) {
+	net := tests.StartIntegrationTestNet(t, tests.IntegrationTestNetOptions{})
+	sponsor := net.GetSessionSponsor()
+	from := sponsor.Address()
+	to := common.HexToAddress("0x2222222222222222222222222222222222222222")
+
+	testCases := []struct {
```
LuisPH3: can we use a map[string] like in other tests?

tests/rpcs/prepare_bundle_test.go
LuisPH3, line 1: as discussed, we need a test which calls prepare and then signs the txs with the key of the sender used. this txs can then be checked with https://github.com/0xsoniclabs/sonic/blob/b572e2cc6dfca61540de1b4ab650c05b415869ba/gossip/blockproc/bundle/validate.go#L278

api/sonicapi/prepare_bundle.go
jenikd: removed

tests/rpcs/prepare_bundle_test.go
jenikd: done

tests/rpcs/prepare_bundle_test.go
```diff
@@ -0,0 +1,362 @@
...
+	return map[string]any{
+		"oneOf":            oneOf,
+		"tolerateFailures": tolerateFailures,
+		"steps":            steps,
+	}
+}
+
+// checkBundleOnlyMarker asserts every transaction has the BundleOnly address in its
+// access list, and that all transactions share the same plan hash storage key.
+func checkBundleOnlyMarker(t *testing.T, result prepareBundleResult) {
```
LuisPH3: This is insufficient. 
This expectation checks that every returned transaction contains the same hash, but it does not verify how is this hash created. The following properties remain untested: 

- The hash of the returned execution plan must be the same as the one included in every transaction. 
- Additionally, every transaction mentioned in the plan hash must be present in the returned transaction list.

api/sonicapi/prepare_bundle.go
```diff
@@ -26,196 +26,179 @@ import (
...
-	for i := range txs {
-		if i < len(gasLimits) && (txs[i].Gas == nil || *txs[i].Gas == 0) {
-			txs[i].Gas = &gasLimits[i]
-		}
-		if txs[i].GasPrice == nil && txs[i].MaxFeePerGas == nil {
-			if txs[i].MaxPriorityFeePerGas == nil {
-				txs[i].GasPrice = gasPrice
-			} else {
-				txs[i].MaxFeePerGas = gasPrice
+// flattenTransactions traverses the execution proposal and extracts a flat list of transactions in depth-first order,
```
LuisPH3:
```suggestion
// flattenTransactions traverses the execution proposal and extracts a map of transactions indexed by the depth-first order position in the tree,
```

api/sonicapi/prepare_bundle_test.go
```diff
@@ -25,189 +25,99 @@ import (
...
 			wantEarliest: 11,
 			wantLatest:   10 + bundle.MaxBlockRange,
 		},
-		{
-			name:         "explicit earliest",
-			currentBlock: 10,
-			earliest:     hexN(50),
-			wantEarliest: 50,
-			wantLatest:   50 + bundle.MaxBlockRange - 1,
+		"only earliest": {
```
LuisPH3, line 74: if only earliest value is given, prepare bundle should fill in the correct values for execution (latest should be defaulted to +1024 earliest)

api/sonicapi/prepare_bundle.go
```diff
@@ -26,196 +26,179 @@ import (
...
-	if latest != nil && earliest != nil {
-		if uint64(*latest) < uint64(*earliest) {
-			return bundle.BlockRange{}, fmt.Errorf("invalid block range: latest block %d is earlier than earliest block %d", *latest, *earliest)
-		}
-		if uint64(*latest)-uint64(*earliest)+1 > bundle.MaxBlockRange {
-			return bundle.BlockRange{}, fmt.Errorf("invalid block range: range %d is too large; must be at most %d blocks", uint64(*latest)-uint64(*earliest)+1, bundle.MaxBlockRange)
-		}
+// validateBlockRange checks that the provided block range is valid and within
+// allowed limits, defaulting to a sensible range if not provided.
+func validateBlockRange(currentBlock uint64, blockRange *RPCRange) (RPCRange, error) {
```
LuisPH3: nit: because this function creates values. better name would be sanitize

api/sonicapi/prepare_bundle.go
```diff
@@ -26,196 +26,179 @@ import (
...
+	if blockRange == nil {
+		maxRange := bundle.MakeMaxRangeStartingAt(currentBlock + 1)
+		return RPCRange{
+			Earliest: hexutil.Uint64(maxRange.Earliest),
+			Latest:   hexutil.Uint64(maxRange.Latest),
+		}, nil
 	}
```
LuisPH3, line 182: If latest is not defined, it shall be defaulted to the appropriate value. (use `bundle.MakeMaxRangeStartingAt` with the starting point provided by user)

api/sonicapi/prepare_bundle_test.go
```diff
@@ -25,189 +25,99 @@ import (
...
+			currentBlock:  10,
+			blockRange:    &RPCRange{Earliest: hexN(50)},
+			errorContains: "invalid block range",
 		},
-		{
-			name:         "explicit latest",
+		"explicit latest": {
```
LuisPH3, line 80: if `latest` is given but not earliest, we either deduce `earliest` or just return an error.

api/sonicapi/prepare_bundle_test.go
```diff
@@ -636,37 +616,498 @@ func Test_PrepareBundle_MultipleTxs_AllOfPlan(t *testing.T) {
...
+		},
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			result, err := api.PrepareBundle(t.Context(), tc.proposal)
+			require.NoError(t, err)
+			require.Len(t, result.Transactions, tc.wantTxCount)
+
+			// Sign each returned tx with the account's private key, then extract
```
LuisPH3: rewrite comment:

> The hash used in the execution plan shall be equal to the hash of the transaction after removing the bundle marker. All referenced transactions in the execution plan shall be pressent in the transaction list.

LuisPH3 (changes requested):
This is pretty close now, there are just some superficial comments, we shall fix the ranges generation to assist users when they provide partial ranges.

LuisPH3 (approved):

PR #888, opened 2026-04-23
Include usedGas in state processor snapshots

LorenzSchueler:
This PR captures the `usedGas` in the snapshots of the `bundleTransactionRunner`. This is needed in order for the used gas of a block to line up with the sum of the gas consumed by its transactions.
A block replay test (which will be added in an upcoming PR), also confirmed that this fix is needed.

### DIFF
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -498,6 +498,7 @@ func (b *bundleTransactionRunner) CreateSnapshot() int {
 		legacyTxOffset:                 b.legacyTxOffset,
 		trueTxOffset:                   b.trueTxOffset,
 		processedTransactionListLength: len(b.processedTransactions),
+		usedGas:                        *b.ctxt.usedGas,
 	}
 	b.snapshots = append(b.snapshots, snapshot)
 	return len(b.snapshots) - 1
@@ -516,6 +517,7 @@ func (b *bundleTransactionRunner) RevertToSnapshot(id int) {
 	b.legacyTxOffset = snapshot.legacyTxOffset
 	b.trueTxOffset = snapshot.trueTxOffset
 	b.processedTransactions = b.processedTransactions[:snapshot.processedTransactionListLength]
+	b.ctxt.usedGas = &snapshot.usedGas
 	b.snapshots = b.snapshots[:id]
 }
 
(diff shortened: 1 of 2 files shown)


LuisPH3 (dismissed):

evmcore/state_processor.go
```diff
@@ -516,6 +517,7 @@ func (b *bundleTransactionRunner) RevertToSnapshot(id int) {
 	b.legacyTxOffset = snapshot.legacyTxOffset
 	b.trueTxOffset = snapshot.trueTxOffset
 	b.processedTransactions = b.processedTransactions[:snapshot.processedTransactionListLength]
+	b.ctxt.usedGas = &snapshot.usedGas
```
HerbertJordan: I think this should be
```suggestion
	*b.ctxt.usedGas = snapshot.usedGas
```
otherwise any future effect would no longer be reflected in the original `usedGas` value, right? And this might be relevant, as the `usedGas` value is passed [here](https://github.com/0xsoniclabs/sonic/blob/bf7e9c11a786f420c45779d50c08f0cd855d2836/evmcore/state_processor.go#L103-L106) as an in/out parameter.

If so, please add a unit test making sure that the address of the `usedGas` field in the context is not altered by any runner operation.

We could consider removing the `usedGas` parameter fully, and just compute it in the end based on the receipts. After all, receipts contain a `GasUsed` field. Do you think this could be done confidently with reasonable effort?

HerbertJordan (changes requested):

evmcore/state_processor.go
LorenzSchueler: Good catch, you are right.
I added a test ensuring that the pointer is the same afterwards.

I would not want to remove the usedGas field completely, because with all the functions running a single transaction only, this could easily be lost.

evmcore/state_processor_test.go
```diff
@@ -2628,13 +2634,40 @@ func TestBundleTransactionRunner_RevertToSnapshot_CallsRevertToInterTxSnapshotOn
...
+		ctxt:                  ctxt,
+		legacyTxOffset:        50,
+		trueTxOffset:          51,
+		processedTransactions: make([]ProcessedTransaction, 100),
+	}
+
+	address := ctxt.usedGas
+
+	snapshotId = bundleTransactionRunner.CreateSnapshot()
+	bundleTransactionRunner.RevertToSnapshot(snapshotId)
```
HerbertJordan: Please also check in-between snapshots, just to make sure that the creation of snapshots also has no effect.

```suggestion
    require.Same(t, address, ctxt.usedGas)
	bundleTransactionRunner.RevertToSnapshot(snapshotId)
```

HerbertJordan (dismissed):

LuisPH3 (approved):

PR #889, opened 2026-04-23
Evaluate bundles during tx pool reorg

LuisPH3:
This PR uses `GetBundleState` to evaluate bundles during reorg runs. 
The code changes enable the following transitions: 

- `queued` -> `dropped`, when a bundle is no longer executable
- `pending` -> `dropped`, when a bundle is no longer executable
- `queued` -> `pending`, when a bundle can be executed immediately. 
- `pending` -> `queued`, when a bundle can not be executed at the moment, but maybe later can succeed. 

Notice that the bundle evaluations in this PR are explicit and will have some performance penalty. Without the changes in this PR, there is no `dropped` transition in the pool, therefore bundles would remain pooled although already executed or expired.

### DIFF
--- a/evmcore/bundle_integration_test.go
+++ b/evmcore/bundle_integration_test.go
@@ -1,0 +1,51 @@
+// Copyright 2026 Sonic Operations Ltd
+// This file is part of the Sonic Client
+//
+// Sonic is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Sonic is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with Sonic. If not, see <http://www.gnu.org/licenses/>.
+
+package evmcore
+
+import (
+	"testing"
+
+	"github.com/0xsoniclabs/sonic/opera"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/params"
+	"go.uber.org/mock/gomock"
+)
+
+func TestBundleIntegration_chainAdapter_ForwardsCallsToChainReader(t *testing.T) {
+	// Create instances of the mocks
+	ctrl := gomock.NewController(t)
+	chain := NewMockStateReader(ctrl)
+
+	// Create a chainAdapter with the mocks
+	adapter := chainAdapter{
+		chain: chain,
+	}
+
+	chain.EXPECT().CurrentRules().Return(opera.Rules{})
+	adapter.GetCurrentNetworkRules()
+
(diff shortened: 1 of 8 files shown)


evmcore/tx_pool.go
```diff
@@ -1502,24 +1519,21 @@ func (pool *TxPool) reset(oldHead, newHead *EvmHeader) {
...
 // promoteExecutables moves transactions that have become processable from the
 // future queue to the set of pending transactions. During this process, all
 // invalidated transactions (low nonce, low balance) are deleted.
 func (pool *TxPool) promoteExecutables(accounts []common.Address) []*types.Transaction {
 	// Track the promoted transactions to broadcast them at once
 	var promoted []*types.Transaction
 
-	subsidiesChecker := pool.createCachedSubsidiesChecker()
+	isSponsored := pool.getSubsidiesCheckerForReorg()
+	evaluateBundleStatus := pool.getBundlesCheckerForPromotion()
```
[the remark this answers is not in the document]
LuisPH3: this comes later, maybe here, maybe somewhow else.

evmcore/tx_pool_bundles_test.go
```diff
@@ -0,0 +1,238 @@
...
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with Sonic. If not, see <http://www.gnu.org/licenses/>.
+
+package evmcore
+
+import (
+	"crypto/ecdsa"
+	big "math/big"
```
facuMH: unnecessary.

facuMH (dismissed):

evmcore/bundle_integration_test.go
```diff
@@ -0,0 +1,51 @@
...
+	chain.EXPECT().CurrentConfig().Return(&params.ChainConfig{})
+	adapter.GetCurrentChainConfig()
+
+	chain.EXPECT().CurrentBlock().Return(&EvmBlock{})
+	adapter.GetLatestHeader()
+
+	hash := common.Hash{123}
+	number := uint64(456)
+	chain.EXPECT().Header(hash, number)
+	adapter.Header(hash, number)
```
HerbertJordan: nit: technically this test would also pass if `GetCurrentNetworkRules`, `GetCurrentChainConfig`, and `GetLatestHeader` are no-ops and the `Header` function is performing all the calls to the adapter.

evmcore/tx_pool_bundles_test.go
```diff
@@ -0,0 +1,238 @@
...
+type dummySubscription struct{}
+
+func (d *dummySubscription) Err() <-chan error {
+	return make(chan error)
+}
+
+func (d *dummySubscription) Unsubscribe() {
+}
+
+var _ event.Subscription = (*dummySubscription)(nil)
```
HerbertJordan: This seems to be used nowhere.

HerbertJordan (dismissed):
The PR is integrating the bundle pre-runner into the transaction pool using promotion/demotion/drop infrastructure established in preceding PRs.

The main changes are covered by on integration test demonstrating that bundles are moved between the different modes. Existing unit and integration tests, as well as integration and stress tests and Norma tests will add more layers of testing to this component.

evmcore/bundle_integration_test.go
LuisPH3: isolated

facuMH (approved):

PR #890, opened 2026-04-23
Preserve gas tip computation

simonlechner:
Geth's internal functions can change behavior any time, we want to preserve the current gas tip computation even if the geth dependency is updated.

### DIFF
--- a/api/ethapi/api.go
+++ b/api/ethapi/api.go
@@ -2013,8 +2013,7 @@ func (s *PublicTransactionPoolAPI) formatTxReceipt(header *evmcore.EvmHeader, tx
 	if header.BaseFee == nil {
 		fields["effectiveGasPrice"] = hexutil.Uint64(tx.GasPrice().Uint64())
 	} else {
-		// EffectiveGasTip returns an error for negative values, this is no problem here
-		gasTip, _ := gasprice.EffectiveGasTip(tx, header.BaseFee)
+		gasTip := tx.EffectiveGasTipValue(header.BaseFee)
 		gasPrice := new(big.Int).Add(header.BaseFee, gasTip)
 		fields["effectiveGasPrice"] = hexutil.Uint64(gasPrice.Uint64())
 	}
--- a/api/ethapi/api.go
+++ b/api/ethapi/api.go
@@ -2014,7 +2014,7 @@ func (s *PublicTransactionPoolAPI) formatTxReceipt(header *evmcore.EvmHeader, tx
 		fields["effectiveGasPrice"] = hexutil.Uint64(tx.GasPrice().Uint64())
 	} else {
 		// EffectiveGasTip returns an error for negative values, this is no problem here
-		gasTip, _ := tx.EffectiveGasTip(header.BaseFee)
+		gasTip, _ := gasprice.EffectiveGasTip(tx, header.BaseFee)
 		gasPrice := new(big.Int).Add(header.BaseFee, gasTip)
 		fields["effectiveGasPrice"] = hexutil.Uint64(gasPrice.Uint64())
 	}
--- a/gossip/blockproc/drivermodule/driver_txs.go
+++ b/gossip/blockproc/drivermodule/driver_txs.go
@@ -185,8 +185,8 @@ func effectiveGasPrice(tx *types.Transaction, baseFee *big.Int) *big.Int {
 
 	// To ensure backwards compatibility the gas tip calculation has to be
 	// preserved in its current form. In case of an error due to a negative
-	// result, the tip is still computed correctly and is used in the
-	// codebase, so we ignore the error here.
+	// result, the computed tip still needs to be used as reported by the
+	// function for backward compatibility with previous client versions.
 	gasTip, _ := gasprice.EffectiveGasTip(tx, baseFee)
 	return new(big.Int).Add(baseFee, gasTip)
 }
--- a/gossip/blockproc/drivermodule/driver_txs.go
+++ b/gossip/blockproc/drivermodule/driver_txs.go
@@ -27,6 +27,7 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 
 	"github.com/0xsoniclabs/sonic/gossip/blockproc"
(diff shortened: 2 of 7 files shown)


gossip/blockproc/drivermodule/driver_txs.go
```diff
@@ -182,7 +183,7 @@ func effectiveGasPrice(tx *types.Transaction, baseFee *big.Int) *big.Int {
 		return tx.GasPrice()
 	}
 	// EffectiveGasTip returns an error for negative values, this is no problem here
```
HerbertJordan: Can you update this comment here, stating that we need to preserve this specific version of how effective gas tips are computed for backward compatibility reasons, and that we need to ignore errors for the same reason?

gossip/gasprice/gas_tip.go
```diff
@@ -0,0 +1,74 @@
...
+
+// To ensure that the effective gas tip calculation is consistent across different versions
+// of the codebase, we have this helper function that can be used in all places where the
+// effective gas tip needs to be calculated.
+// Based on geth 1.16.9 which is used in sonic 2.1.6
+// https://github.com/0xsoniclabs/go-ethereum/blob/879ee1854e078ae6a6958f7cd40094377fd90b98/core/types/transaction.go#L358
+
+// EffectiveGasTip returns the effective miner gasTipCap for the given base fee.
+// Note: if the effective gasTipCap would be negative, this method
+// returns ErrGasFeeCapTooLow, and value is undefined.
```
HerbertJordan: Please extend the documentation to state that the resulting value is indeed relevant even if there is an error reported, and must thus be preserved to produce the same result as it does right now until its consensus criticality is resolved.

gossip/gasprice/reactive.go
```diff
@@ -264,7 +264,7 @@ func (gpo *Oracle) calcTxpoolStat() txpoolStat {
 	for _, tx := range sorted {
 		for p < uint64(len(s.percentiles)) && gasCounter >= p*maxGasToIndex/uint64(len(s.percentiles)) {
 			// EffectiveGasTip returns an error for negative values, this is no problem here
```
HerbertJordan: Please check where this function is used and if it is not consensus critical, errors should be handled. Wdyt?

api/ethapi/api.go
```diff
@@ -2014,7 +2014,7 @@ func (s *PublicTransactionPoolAPI) formatTxReceipt(header *evmcore.EvmHeader, tx
 		fields["effectiveGasPrice"] = hexutil.Uint64(tx.GasPrice().Uint64())
 	} else {
 		// EffectiveGasTip returns an error for negative values, this is no problem here
-		gasTip, _ := tx.EffectiveGasTip(header.BaseFee)
+		gasTip, _ := gasprice.EffectiveGasTip(tx, header.BaseFee)
```
HerbertJordan: I think errors should be handled here.

HerbertJordan (changes requested):
Where possible, we should add proper error handling. For the critical case, we should update the documentation stating why we can not do proper error handling at this point.

gossip/gasprice/reactive.go
simonlechner: This function is used to provide a tip based on the current transactions in the pool, therefore it is not consensus critical.
Prior to the update to [geth 1.16.0](https://github.com/0xsoniclabs/sonic/pull/349) `EffectiveGasTipValue` has been used which does not return an error. I changed this usage back to the original function.

api/ethapi/api.go
simonlechner: Prior to the update to [geth 1.16.0](https://github.com/0xsoniclabs/sonic/pull/349) `EffectiveGasTipValue` has been used which does not return an error. I changed this usage back to the original function.

gossip/blockproc/drivermodule/driver_txs.go
```diff
@@ -181,8 +182,12 @@ func effectiveGasPrice(tx *types.Transaction, baseFee *big.Int) *big.Int {
 	if baseFee == nil {
 		return tx.GasPrice()
 	}
-	// EffectiveGasTip returns an error for negative values, this is no problem here
-	gasTip, _ := tx.EffectiveGasTip(baseFee)
+
+	// To ensure backwards compatibility the gas tip calculation has to be
+	// preserved in its current form. In case of an error due to a negative
+	// result, the tip is still computed correctly and is used in the
+	// codebase, so we ignore the error here.
```
HerbertJordan:
```suggestion
	// result, the computed tip still needs to be used as reported by the function
	// for backward compatibility with previous client versions.
```

gossip/gasprice/gas_tip.go
```diff
@@ -0,0 +1,78 @@
...
+// Due to usages of this function in consensus critical code,
+// the current tip computation has to be preserved.
+// Even if an error is returned due to a negative result,
+// the tip is still computed correctly and is used in the codebase.
```
HerbertJordan:
```suggestion
// EffectiveGasTip returns the effective miner gasTipCap for the given base fee.
// Note: if the effective gasTipCap would be negative, this method
// returns ErrGasFeeCapTooLow, but the resulting value is still
// consensus critical. Thus, the result produced by this function,
// even in the error case needs to be preserved as is.
```

gossip/gasprice/reactive.go
```diff
@@ -264,7 +264,7 @@ func (gpo *Oracle) calcTxpoolStat() txpoolStat {
...
-			gasTip, _ := tx.EffectiveGasTip(minGasPrice)
+			gasTip := tx.EffectiveGasTipValue(minGasPrice)
```
HerbertJordan, line 266: Outdated comment.

HerbertJordan (changes requested):

HerbertJordan (dismissed):

LuisPH3 (dismissed):

LuisPH3 (approved):

simonlechner:
The `TestRegistryUpdate_UpdatesAreEffectiveImmediately` test had to be updated, due to change back to gas tip computations before the geth 1.16. Update in the RPC.

PR #891, opened 2026-04-23
Enable full bundle validation when pool receives bundles.

LuisPH3:
This PR connects pool validation, executed when txs are received into the pool. With the bundle state evaluation.
The TX pool will accept any bundle which is not permanently blocked.

### DIFF
--- a/evmcore/tx_validation.go
+++ b/evmcore/tx_validation.go
@@ -28,6 +28,8 @@ import (
 	"github.com/0xsoniclabs/sonic/inter/state"
 	"github.com/0xsoniclabs/sonic/opera"
 	"github.com/0xsoniclabs/sonic/utils"
+	"github.com/Fantom-foundation/lachesis-base/inter/idx"
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/log"
@@ -397,6 +399,18 @@ func validateBundleTransactions(
 	stateDb state.StateDB,
 	signer types.Signer,
 ) error {
+	return validateBundleTransactionsInternal(tx, netRules, chainState, stateDb, signer, GetBundleState)
+}
+
+func validateBundleTransactionsInternal(
+	tx *types.Transaction,
+	netRules NetworkRules,
+	chainState StateReader,
+	stateDb state.StateDB,
+	signer types.Signer,
+	getBundleState func(ChainStateForBundleEval, state.StateDB, *types.Transaction) BundleState,
+) error {
+
 	// This check only covers bundle transactions, ignore the rest.
 	if !bundle.IsEnvelope(tx) {
 		return nil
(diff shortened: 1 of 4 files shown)


evmcore/tx_validation.go
```diff
@@ -423,10 +438,61 @@ func validateBundleTransactions(
...
+	state state.StateDB
+}
+
+// newGetBundleStateAdaptor creates a new getBundleStateAdaptor with the given StateReader.
+//
+// GetBundleState requires a new stateDb instance which gets released.
+// the shared stateDb instance used by validation and other parts of the pool
+// cannot be used because it will be released and invalidated for further use.
+func newGetBundleStateAdaptor(chain StateReader) (getBundleStateAdaptor, error) {
+	state, err := chain.CurrentStateDB()
```
HerbertJordan: I think this `state` instance needs to be released somewhere.

evmcore/tx_validation.go
```diff
@@ -423,10 +438,61 @@ func validateBundleTransactions(
...
+		StateReader: chain,
+		state:       state,
+	}, nil
+}
+
+func (f getBundleStateAdaptor) GetCurrentNetworkRules() opera.Rules {
+	return f.CurrentRules()
+}
+
+func (f getBundleStateAdaptor) GetEvmChainConfig(idx.Block) *params.ChainConfig {
```
HerbertJordan, line 462: As discussed earlier, instead of ignoring the block parameter we should see whether we can change this to a `GetCurrentEvmChainConfig` in the respective interface.

evmcore/tx_validation.go
```diff
@@ -423,10 +438,61 @@ func validateBundleTransactions(
...
+
+func (f getBundleStateAdaptor) GetLatestHeader() *EvmHeader {
+	return f.StateReader.CurrentBlock().Header()
+}
+
+func (f getBundleStateAdaptor) Header(hash common.Hash, number uint64) *EvmHeader {
+	return f.StateReader.Block(hash, number).Header()
+}
+
+func (f getBundleStateAdaptor) StateDB() state.StateDB {
```
HerbertJordan: After #894 this should no longer be needed.

evmcore/tx_validation.go
```diff
@@ -423,10 +438,61 @@ func validateBundleTransactions(
...
+// cannot be used because it will be released and invalidated for further use.
+func newGetBundleStateAdaptor(chain StateReader) (getBundleStateAdaptor, error) {
+	state, err := chain.CurrentStateDB()
+	if err != nil {
+		return getBundleStateAdaptor{}, fmt.Errorf("failed to get current state DB: %w", err)
+	}
+	return getBundleStateAdaptor{
+		StateReader: chain,
+		state:       state,
+	}, nil
```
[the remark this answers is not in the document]
LuisPH3: this is gone, no need to create new state

evmcore/tx_validation_test.go
```diff
@@ -1865,6 +1980,33 @@ func TestValidateTx_Success(t *testing.T) {
...
+	reader.EXPECT().CurrentConfig()
+	adaptor.GetEvmChainConfig(1)
+
+	reader.EXPECT().CurrentBlock()
+	adaptor.GetLatestHeader()
+
+	hash := common.HexToHash("0x1234")
+	number := uint64(100)
+	reader.EXPECT().Block(hash, number)
+	adaptor.Header(hash, number)
```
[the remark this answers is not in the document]
LuisPH3: this is checked now

evmcore/tx_validation.go
LuisPH3: I added a fixme, I will work it in the following hours. 
The reason is that this PR may break ongoing integration test development and would be nice to have the final semantics online

evmcore/tx_validation.go
LuisPH3: no longer there.

evmcore/tx_validation_test.go
```diff
@@ -1865,6 +1951,46 @@ func TestValidateTx_Success(t *testing.T) {
...
+		})
+		header := adaptor.Header(hash, number)
+		require.EqualValues(t, header.Number.Uint64(), number)
+	})
+
+	t.Run("missing block returns nil", func(t *testing.T) {
+		reader.EXPECT().Block(gomock.Any(), gomock.Any()).Return(nil)
+		heder := adaptor.Header(common.Hash{}, 1)
+		require.Nil(t, heder)
+	})
```
HerbertJordan: While this likely works as intended here, be advised that this would also pass if all `GetCurrentNetworkRule`, `GetEvmChainConfig`, and `GetLatestHeader` would be no-ops and all expected functions are called when calling `Header`.

Also, you should not reuse a mock created with a `t` of a parent test in a nested test. The failure will get attributed to the parent test instead of the sub-test.

HerbertJordan (dismissed):

evmcore/tx_validation_test.go
LuisPH3: yes, shortcuts.

facuMH (approved):

evmcore/tx_validation_test.go
```diff
@@ -1865,6 +1951,68 @@ func TestValidateTx_Success(t *testing.T) {
...
 }
 
+func Test_ValidateBundle_GetBundleStateAdaptor(t *testing.T) {
+	hash := common.HexToHash("0x1234")
+	number := uint64(100)
+
+	tests := map[string]struct {
+		setup  func(reader *MockStateReader)
+		action func(adaptor *getBundleStateAdaptor)
+		check  func(t *testing.T)
```
LorenzSchueler, line 1961:
```suggestion
```

LorenzSchueler (approved):

PR #892, opened 2026-04-23
Add support for bundles to single proposer mode

HerbertJordan:
With this PR support for bundles is added in the single-proposer mode.

The only change requires is to accept transactions in the scheduler even if their execution does not lead to an inclusion of those transactions in the resulting block. For bundles, this is always the case, since the envelopes are discarded.


LuisPH3 (approved):

facuMH (approved):
Looking forward to the follow-up about minimum efficiency.

PR #893, opened 2026-04-23
Add bundle history hash record

facuMH:
This PR adds a record of history hash per block, under the prefix `'h'` to the `ProcessesBundles` store.

- New `getBlockHistoryHashKey` is introduced to generate the history hash entry key (using '`h`' prefix)
- New `GetOldestRetainedBundleHistoryHash` is introduced to get the oldest entry. 
- New block history hash is added as part of `AddProcessedBundles` when the most recent history has is written (nil key), and deleted as part of the pruning done by `deleteOutdatedBundles`.

This PR contributes to https://github.com/0xsoniclabs/sonic-admin/issues/706


gossip/store_processed_bundles.go
```diff
@@ -129,33 +133,52 @@ func (s *Store) addNewBundles(
...
+		}
+		oldBundleBlockNumber := binary.BigEndian.Uint64(key[1 : 1+8])
+		if oldBundleBlockNumber > highestOutdatedBlockNumber {
+			break
+		}
+		hash := common.BytesToHash(key[1+8:])
+		err := errors.Join(
+			batch.Delete(getIndexKey(oldBundleBlockNumber, hash)),
+			batch.Delete(getEntryKey(hash)),
+			batch.Delete(getBlockHistoryHashKey(oldBundleBlockNumber)),
```
HerbertJordan: You are deleting the hash entry for the block here - once, for each entry in the DB, and then again in the loop below. The loop below should be enough, no?

gossip/store_processed_bundles_test.go
```diff
@@ -238,6 +253,32 @@ func TestStore_AddProcessedBundles_LogsOnBatchWriteError(t *testing.T) {
...
+		}
+	}
+
+	// as block 1 is the first bundle, block 2 should have had a history hash
+	// entry, but it should now be gone.
+	oldestBlock, oldestHash, ok := store.GetOldestRetainedBundleHistoryHash()
+	require.True(t, ok)
+	// oldest should be 3, as blocks 0, 1, and 2 are pruned.
+	require.Equal(t, uint64(3), oldestBlock)
+	require.NotZero(t, oldestHash)
```
HerbertJordan: This looks a lot like it was implemented after the code, and the numbers of the blocks that should be present or gone are just fitted to match the implementation. Can you compute those offsets from first principles -- meaning, based on the definitions in the `bundle` package like [MakeMaxRangeStartingAt](https://github.com/0xsoniclabs/sonic/blob/1ed03f9f7d7b6bac01558595f674bfc595273313/gossip/blockproc/bundle/block_range.go#L45)? You should not do an independent interpretation of the `MaxBlockRange` constant here.

The rest of the tests also seem to have the same problem. You hard-code that if you add blocks up to bundle.MaxBlockRange + 1 block 3 is the oldest. This is in your store implementation like that, but is this the intended relation according to the bundle package definition? This needs to be verified in at least one of those tests, if not implicit in all.

If there is such a check in some of the tests not shown in this diff, please point me to that.

gossip/store_processed_bundles_test.go
HerbertJordan, line 1: There is a test missing that is checking that the hashes stored in the new table are indeed equivalent to the history hash at the respective block.

This test should:
- insert  MaxBlockRange * 2 blocks
- after each step 
   - record the current history hash
   - for the past valid history, query the hash from the `h` table and check that the match the recorded hashes

HerbertJordan (changes requested):

gossip/store_processed_bundles_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/893/commits/b81408abede81b22622a9e0a913beeaa049e2d5f - refactors the tests introduces.

gossip/store_processed_bundles.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/893/commits/b5c9ad21ebac3d5d93c626d48d3d8b9f62e9b958

gossip/store_processed_bundles_test.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/893/commits/60b781e2f54a605162f8ab69738f95b2895fcd40

gossip/store_processed_bundles_test.go
facuMH: In particular `TestStore_AddProcessedBundles_RemovesOldHistoryHash_EvenForBlockNumberWithoutBundles`, enforces that everything before the "oldest bundle history hash" does not have an entry (aka is empty), and should not verify whether the "oldest bundle history hash" is within the range defined by the bundles package.

`TestStore_GetOldestRetainedBundleHistoryHash_AdvancesAfterPruning` is the only test than needs to verify that the `GetOldestRetainedBundleHistoryHash` effectively returns what the bundles package defines for `MaxBlockRange` (and what is thin range or not).

wdyt?

gossip/store_processed_bundles_test.go
```diff
@@ -238,6 +253,72 @@ func TestStore_AddProcessedBundles_LogsOnBatchWriteError(t *testing.T) {
...
+
+		_, currentHistoryHash := store.GetProcessedBundleHistoryHash()
+		historyHashForBlock, err := store.table.ProcessedBundles.Get(getBlockHistoryHashKey(block))
+		require.NoError(t, err)
+		if len(historyHashForBlock) == 0 {
+			require.Zero(t, currentHistoryHash)
+		} else {
+			require.Equal(t, currentHistoryHash.Bytes(), historyHashForBlock)
+		}
+	}
```
HerbertJordan, line 327: nit: In every step, the full available history should be checked:
```suggestion
	var historicHashes []common.Hash
	for block := range bundle.MaxBlockRange * 2 {
		// randomly add bundles to some blocks, but not all
		if rand.Uint64()%2 == 0 {
			store.AddProcessedBundles(block, map[common.Hash]bundle.PositionInBlock{
				uint64ToHash(block): {},
			})
		} else {
			store.AddProcessedBundles(block, nil)
		}

		_, currentHistoryHash := store.GetProcessedBundleHistoryHash()
		historicHashes = append(historicHashes, currentHistoryHash)

		for past := range block + 1 {
			historyHashForBlock, err := store.table.ProcessedBundles.Get(getBlockHistoryHashKey(past))
			require.NoError(t, err)
			if len(historyHashForBlock) > 0 {
				require.Equal(t, historicHashes[past], common.BytesToHash(historyHashForBlock))
			}
		}
	}
```

gossip/store_processed_bundles.go
```diff
@@ -129,33 +133,51 @@ func (s *Store) addNewBundles(
...
-				s.Log.Crit("failed to delete old processed bundle hash", "error", err)
-			}
+	if blockNum < bundle.MaxBlockRange-1 {
+		return
+	}
+
+	highestOutdatedBlockNumber := blockNum - bundle.MaxBlockRange + 1
+
+	// Prune bundle-index entries and all associated keys ('i', 'e', 'h').
+	// key layout for 'i': 1 byte prefix + 8 bytes blockNum + 32 bytes execPlanHash
```
LorenzSchueler: Outdated, 'h' is pruned only in the next loop

gossip/store_processed_bundles.go
```diff
@@ -129,33 +133,51 @@ func (s *Store) addNewBundles(
...
+			batch.Delete(getEntryKey(hash)),
+		)
+		if err != nil {
+			s.Log.Crit("failed to delete old processed bundle hash", "error", err)
+		}
+	}
+
+	// Also prune per-block history-hash entries for blocks that had no bundles
+	// (those blocks have an 'h' key but no 'i' key, so the loop above misses them).
+	// key layout for 'h': 1 byte prefix + 8 bytes blockNum
```
LorenzSchueler: outdated

LuisPH3 (dismissed):

LorenzSchueler (dismissed):

gossip/store_processed_bundles_test.go
```diff
@@ -238,6 +253,72 @@ func TestStore_AddProcessedBundles_LogsOnBatchWriteError(t *testing.T) {
...
+	store, err := NewMemStore(t)
+	require.NoError(t, err)
+	// Run enough blocks that some 'h' entries from bundle-less blocks get pruned.
+	const totalBlocks = bundle.MaxBlockRange * 2
+	const firstBundle = bundle.MaxBlockRange / 2
+	for currentBlock := uint64(1); currentBlock <= totalBlocks; currentBlock++ {
+		for block := range currentBlock {
+			// add a single block with bundles
+			if block == firstBundle {
+				store.AddProcessedBundles(block, map[common.Hash]bundle.PositionInBlock{
```
HerbertJordan: This setup makes no sense to me ... you are adding the same blocks in the order 0, 0, 1, 0, 1, 2, 0, 1, 2, 3, ... why?

facuMH: outer loop removed.

gossip/store_processed_bundles.go
```diff
@@ -129,33 +133,50 @@ func (s *Store) addNewBundles(
...
+	it2 := s.table.ProcessedBundles.NewIterator([]byte{'h'}, nil)
+	defer it2.Release()
+	for it2.Next() {
+		key := it2.Key()
+		if len(key) != 1+8 {
+			continue
+		}
+		oldBlockNumber := binary.BigEndian.Uint64(key[1:])
+		if oldBlockNumber > highestOutdatedBlockNumber {
```
HerbertJordan: In order to validate the entire table content, you need to keep the hash of the predecessor of the last block for which the execution plan needs to be retained.

For the validation of a block N you need to compute
```
hash(N) = keccak(hash(N-1), hash(xor(added_plans), N)
```
So, to verify that the plans stored for block N are correct, you need to have the hash of block N-1. 
I think with the current setup you can not verify the oldest added execution plans.

gossip/store_processed_bundles_test.go
```diff
@@ -238,6 +253,73 @@ func TestStore_AddProcessedBundles_LogsOnBatchWriteError(t *testing.T) {
...
+				uint64ToHash(block): {},
+			})
+		} else {
+			// add blocks without bundles to advance the block number and trigger pruning
+			store.AddProcessedBundles(block, nil)
+		}
+	}
+
+	oldestBlock, _, ok := store.GetOldestRetainedBundleHistoryHash()
+	if currentBlock > firstBundle {
```
HerbertJordan: `currentBlock` and `firstBundle` are constants, and given the values, this branch is always true. Consider simplifying the test case by removing dead branches.

gossip/store_processed_bundles_test.go
```diff
@@ -296,6 +378,73 @@ func TestStore_GetProcessedBundleHistoryHash_LogsOnInvalidStateLength(t *testing
...
+	const totalBlocks = bundle.MaxBlockRange * 2
+	// The iteration over total blocks is to show that at different block heights
+	// the same block number can be retained, oldest or pruned.
+	for currentBlock := uint64(1); currentBlock <= totalBlocks; currentBlock++ {
+		var expectedOldestHash common.Hash
+		expectedOldestBlock := uint64(0)
+		foundOldest := false
+		// Run enough blocks that some 'h' entries from bundle-less blocks get pruned.
+		for block := range currentBlock {
+			store.AddProcessedBundles(block, map[common.Hash]bundle.PositionInBlock{
```
HerbertJordan: Also here, blocks are inserted in a double-nested loop like in a previous test case where you removed it. If I read this correctly, you are adding the blocks in the order 0,1,0,1,2,0,1,2,3,0,1,2,3,4,...

facuMH: ecah iteration of `currentBlock` must work with a different store. The importance of the outer loop is to enable testing that for different block heights, the last known (the history hash of the first block in range) is correct.

gossip/store_processed_bundles.go
facuMH: https://github.com/0xsoniclabs/sonic/pull/893/commits/ebe79ee3da82be88d09cc97403ace11f1f67b994

gossip/store_processed_bundles.go
```diff
@@ -250,6 +277,59 @@ func (s *Store) GetProcessedBundleHistoryHash() (uint64, common.Hash) {
 	return blockNum, hash
 }
 
+// GetOldestRetainedBundleHistoryHash returns the block number and bundle
+// history hash for the oldest retained block that is still within the
+// MaxBlockRange window. The store keeps one extra predecessor 'h' entry
+// before this block to allow independent verification of the hash chain,
+// but this function skips it and returns the first in-range entry.
```
HerbertJordan: Who is going to need the second oldest entry?

I thought this function is for the G file export feature. And this export needs the oldest, the one that is 1 block older than the oldest execution plans.

gossip/store_processed_bundles_test.go
```diff
@@ -238,6 +255,74 @@ func TestStore_AddProcessedBundles_LogsOnBatchWriteError(t *testing.T) {
...
+			// add blocks without bundles to advance the block number and trigger pruning
+			store.AddProcessedBundles(block, nil)
+		}
+	}
+
+	oldestBlock, _, ok := store.GetOldestRetainedBundleHistoryHash()
+	require.True(t, ok)
+
+	for block := range oldestBlock - 1 {
+		got, err := store.table.ProcessedBundles.Get(getBlockHistoryHashKey(block))
```
HerbertJordan: The logical problem with this test setup is that it is self-referential. The implementation of `GetOldestRetainedBundleHistoryHash` iterates through the table and looks for the first value. And then you check that it is indeed the first value by verifying that everything before is a `nil`.

So the test is not really checking a specific block number, but just that there is one.

HerbertJordan (commented):
A few first thoughts, will continue to review later.

gossip/store_processed_bundles.go
facuMH: second oldest entry removed.

gossip/store_processed_bundles_test.go
```diff
@@ -238,17 +255,90 @@ func TestStore_AddProcessedBundles_LogsOnBatchWriteError(t *testing.T) {
...
+		} else {
+			store.AddProcessedBundles(block, nil)
+		}
+
+		_, currentHistoryHash := store.GetLatestProcessedBundleHistoryHash()
+		historicHashes = append(historicHashes, currentHistoryHash)
+
+		for past := range block + 1 {
+			historyHashForBlock, found := store.GetProcessedBundleHistoryHash(past)
+			require.NoError(t, err)
```
HerbertJordan: nit: there is actually no error to be checked here

HerbertJordan (dismissed):

gossip/store_processed_bundles.go
```diff
@@ -129,34 +133,59 @@ func (s *Store) addNewBundles(
...
-				batch.Delete(getEntryKey(hash)),
-			)
-			if err != nil {
-				s.Log.Crit("failed to delete old processed bundle hash", "error", err)
-			}
+	if blockNum < bundle.MaxBlockRange-1 {
+		return
+	}
+
+	highestOutdatedBlockNumber := blockNum - bundle.MaxBlockRange + 1
```
LorenzSchueler:
```suggestion
	highestOutdatedBlockNumber := blockNum - bundle.MaxBlockRange
```
I think what you are computing right now is the lowest non-outdated block number.
Lets say the current block is 5 and max block range is 2. Then 4 and 5 are not outdated, but <=3 are. However this corrently computes `5 - 2 + 1 = 4` as the highest outdated block number.

gossip/store_processed_bundles.go
```diff
@@ -129,34 +133,59 @@ func (s *Store) addNewBundles(
...
+	defer it2.Release()
+	for it2.Next() {
+		key := it2.Key()
+		if len(key) != 1+8 {
+			continue
+		}
+		oldBlockNumber := binary.BigEndian.Uint64(key[1:])
+		// NOTE: we keep one extra history hash after than the other entries.
+		// this is useful for genesis export/import.
+		if oldBlockNumber >= highestOutdatedBlockNumber {
```
LorenzSchueler, line 182:
```suggestion
		// NOTE: we keep one extra history hash after than the other entries.
		// this is useful for genesis export/import.
		if oldBlockNumber > highestOutdatedBlockNumber {
```
I think this in an artifact of the `highestOutdatedBlockNumber` being off by one.

Please correct me, but as far as I understand, the idea is that we want to ensure the imported data is correct. In the future, the cumulative block hash is included in the epoch sealing, which we know will happen at least every 1024 blocks. Therefore, if we have the last 1024 hashes, one of them will be in the epoch sealing.

gossip/store_processed_bundles_test.go
```diff
@@ -238,17 +255,89 @@ func TestStore_AddProcessedBundles_LogsOnBatchWriteError(t *testing.T) {
...
+
+		// Check that the earliest bundle history block is updated accordingly.
+		earliestHashBlockNumber, _, found := store.GetEarliestBundleHistoryHash()
+		if block < firstBundle {
+			require.False(t, found)
+		} else {
+			require.True(t, found)
+
+			// Check that the block number of the earliest hash is moving one
+			// step at a time, even if the added bundles execution lists are.
```
LorenzSchueler: ... are what?

gossip/store_processed_bundles.go
LorenzSchueler: A discussion with @facuMH cleared things up a bit. This is because `blockNum` is of the old block, and we are basically deleting hashes that are obsolete for the new one. Maybe we should have 
```go
func (s *Store) deleteOutdatedBundles(oldBlock uint64, batch kvdb.Batch) {
	nextBlock := oldBlock + 1
	// ... use nextBlock from here on...
```

LorenzSchueler: maybe even currentBlock instead of nextBlock

gossip/store_processed_bundles.go
LorenzSchueler: Might be an artifact by blockNum being basically one lower than the one we are constructing the state for.

LorenzSchueler (approved):

PR #895, opened 2026-04-23
Add verifiability tests for blocks with bundles

LorenzSchueler:
This PR adds a test to make sure that blocks containing bundles can be replayed and verified. The utilities for the verification have been extracted from an existing test (first commit) and modified slightly to make the new tests work (part of second commit).


tests/block_verifiability_utils.go
```diff
@@ -212,19 +210,19 @@ func (s *State) ApplyGenesis(genesis *makefakegenesis.GenesisJson) error {
...
-	upgrades opera.Upgrades,
+	rules opera.Rules,
 	block *types.Block,
 ) (types.Receipts, error) {
 
 	chainConfig := opera.CreateTransientEvmChainConfig(
-		chainId,
-		nil,
+		rules.NetworkID,
+		[]opera.UpgradeHeight{{Height: 2, Upgrades: rules.Upgrades}},
```
LorenzSchueler: @HerbertJordan @LuisPH3 this was needed to make the new tests pass. Is there a better way than hard-coding the height?

tests/bundles/block_verifiabiity_test.go
```diff
@@ -0,0 +1,76 @@
...
+	defer client.Close()
+
+	signer := types.LatestSignerForChainID(net.GetChainId())
+
+	senders := tests.MakeAccountsWithBalance(t, net, 3, big.NewInt(1e18))
+
+	blockNumber, err := client.BlockNumber(t.Context())
+	require.NoError(t, err)
+
+	envelope, _, plan := bundle.NewBuilder().
```
HerbertJordan, line 43: In a follow-up we should test some more complex cases. In particular, cases including mixtures of bundles and sponsorship would be interesting.

tests/block_verifiability_utils.go
HerbertJordan: The fact that this and "Height:2" is needed might indicate that we are not enforcing Brio rules during the fake-genesis when starting up the fake network. This should be investigated.

HerbertJordan (approved):
Looks good to me, but the detected issue with the needed upgrade after block 2 may indicate a bug in the genesis processing of the client -- or in the processing of the genesis data in the validation utils. This should be investigated.

tests/block_verifiability_utils.go
LuisPH3: I find this strange as well.
Early exploration shows that genesis applies rules from block 1 (included? ), I cannot explain why here is 2. this needs investigation

LorenzSchueler: Please take a look at the latest commit. When supplying the hash in the `dummyHeaderReader` the upgrade height can be set to 1. Without this change, the hash of block 0 is assumed to be `0x0` in the history contract, and therefore we get a state root mismatch.

integration/makegenesis/genesis.go
```diff
@@ -169,6 +169,7 @@ type dummyHeaderReturner struct {
 func (d dummyHeaderReturner) Header(_ common.Hash, position uint64) *evmcore.EvmHeader {
 	if position < uint64(len(d.blocks)) {
 		return &evmcore.EvmHeader{
+			Hash:    common.Hash(d.blocks[position].BlockHash),
```
LuisPH3, line 172: This is a bugfix for an issue identified during development of this test:

dummyHeaderReturner is used in [ExecuteGenesisTxs](https://github.com/0xsoniclabs/sonic/blob/6d4979b4375e8758ee0e636ae5cc1f37cea72c81/integration/makegenesis/genesis.go#L263)

**ExecuteGenesisTxs** is used in: 
- [makeFakeGenesis](https://github.com/0xsoniclabs/sonic/blob/6d4979b4375e8758ee0e636ae5cc1f37cea72c81/integration/makefakegenesis/genesis.go#L169)
- [json genesis](https://github.com/0xsoniclabs/sonic/blob/6d4979b4375e8758ee0e636ae5cc1f37cea72c81/integration/makefakegenesis/json.go#L303)

The new semantics seem correct and the reach of this function seems to be only related to testing tooling. I think this fix can be merged.
@HerbertJordan wdyt?

tests/bundles/block_verifiabiity_test.go
LuisPH3: i agree

tests/block_verifiability_utils.go
```diff
@@ -0,0 +1,324 @@
...
+// and updating the state accordingly. It returns the receipts of the transactions
+// in the block, or an error if the block could not be processed.
+func (s *State) ApplyBlock(
+	rules opera.Rules,
+	block *types.Block,
+) (types.Receipts, error) {
+
+	chainConfig := opera.CreateTransientEvmChainConfig(
+		rules.NetworkID,
+		[]opera.UpgradeHeight{{Height: 1, Upgrades: rules.Upgrades}},
```
HerbertJordan: Your fix also seems to have fixed if you use block 0 here. To remove this odd special rule for block 0, please update the code here to
```suggestion
		[]opera.UpgradeHeight{{Height: 0, Upgrades: rules.Upgrades}}, 
```

HerbertJordan (commented):
Great find, and the fix makes sense to me.

HerbertJordan (dismissed):

LuisPH3 (approved):

HerbertJordan (approved):

PR #896, opened 2026-04-23
Update geth to 1.17.2

simonlechner:
This PR updates the geth dependency to version 1.17.2.
A `newGasPool` function has been introduced, the transfer function now requires a `params.Rules` parameters and the new `EmitLogsForBurnAccounts` has been added to the statedb interface.


api/ethapi/api.go
```diff
@@ -1182,7 +1182,7 @@ func DoCall(
 	}
 
 	// Add sufficient gas to the pool.
-	gp := new(core.GasPool).AddGas(math.MaxUint64)
+	gp := core.NewGasPool(math.MaxUint64)
```
HerbertJordan, line 1185: Please double-check that non of the functions we are using on the GasPool has changed its semantic.

HerbertJordan (commented):
Looks good, but I would like you to double-check that there is no change in the implementation of the GasPool that could affect us.

api/ethapi/api.go
simonlechner: Sonic is only calling `NewGasPool` and `Gas` functions from the gas pool. The latter returns the remaining gas from the pool, the [calculation](https://github.com/ethereum/go-ethereum/blob/822e7c648618a43e7f8326868b9334b9976185c8/core/state_transition.go#L569-L575) remains the same inside the geth dependency pre Amsterdam.

HerbertJordan (dismissed):
Thanks for the follow-up check with the GasPool. I think the best next step is to merge it and run release testing over it.

LuisPH3 (dismissed):

LuisPH3 (approved):

PR #900, opened 2026-04-24
Enable the emitor throttler by default

jmpike:
This PR enables the event emitter throttling in the default configuration.


LuisPH3:
Event throttler was enabled by default in the main branch here : https://github.com/0xsoniclabs/sonic/pull/634
We can update the boolean flag, if necessary, but the current setup overwrites it every time.

PR #901, opened 2026-04-24
Modify  `TransactionProcessor` to use current chainConfig instead of block-deduced one.

LuisPH3:
This PR simplifies retrieval of chainconfig for execution, not requiring to pass block details. 

This can be done because `TransactionProcessor` is inherently stateful in sonic, it uses `GetCurrentNetworkRules`. The change here just repeats the pattern for `ChainConfig` generation.


LorenzSchueler (dismissed):

gossip/emitter/txs.go
```diff
@@ -291,11 +291,16 @@ func (a *preCheckChainStateAdapter) Header(hash common.Hash, number uint64) *evm
...
+	blockNumber := uint64(0)
+	if block != nil {
+		blockNumber = block.Number
+	}
 	return opera.CreateTransientEvmChainConfig(
 		a.external.GetRules().NetworkID,
 		a.external.GetUpgradeHeights(),
-		blockHeight,
+		idx.Block(blockNumber),
 	)
```
HerbertJordan: Cheaper:

```suggestion
	return opera.CreateTransientEvmChainConfig(
		a.external.GetRules().NetworkID,
		a.external.GetUpgradeHeights(),
		a.external.GetLatestBlockIndex(),
	)
```

HerbertJordan (dismissed):

gossip/emitter/txs.go
LuisPH3: thanks!

LorenzSchueler (approved):

PR #905, opened 2026-04-27
Submit bundle RPC function with composable execution plan

jenikd:
This PR introduces composable bundle execution plan in the sonic_submitBundle RPC function

### DIFF
--- a/api/sonicapi/submit_bundle.go
+++ b/api/sonicapi/submit_bundle.go
@@ -129,17 +129,11 @@ func (a *PublicBundleAPI) SubmitBundle(
 		return common.Hash{}, fmt.Errorf("failed to sign bundle transaction: %w", err)
 	}
 
-	// Validate generated transaction
-	_, plan, err := bundle.ValidateEnvelope(signer, tx)
-	if err != nil {
-		return common.Hash{}, fmt.Errorf("failed to validate bundle transaction: %w", err)
-	}
-
 	// Submit the transaction to the network
 	_, err = ethapi.SubmitTransaction(ctx, a.b, tx)
 	if err != nil {
 		return common.Hash{}, fmt.Errorf("failed to submit bundle transaction: %w", err)
 	}
 
-	return plan.Hash(), nil
+	return execPlan.Hash(), nil
 }
--- a/api/sonicapi/submit_bundle.go
+++ b/api/sonicapi/submit_bundle.go
@@ -66,7 +66,7 @@ func (a *PublicBundleAPI) SubmitBundle(
 		return common.Hash{}, fmt.Errorf("invalid block range: %w", err)
 	}
 
-	execPlan, err := toBundleExecutionPlan(args.ExecutionPlan)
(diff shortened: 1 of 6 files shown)


LuisPH3 (dismissed):

api/sonicapi/submit_bundle.go
LuisPH3: We could argue that this is superfluous:
`SubmitTransaction` will validate the transaction in the transaction pool, this will validate the envelope and other deeper checks, error is returned if something is wrong.

api/sonicapi/submit_bundle.go
```diff
@@ -54,59 +54,44 @@ func (a *PublicBundleAPI) SubmitBundle(
...
 			return common.Hash{}, fmt.Errorf("failed to decode bundled transaction %d: %w", i, err)
 		}
-		txRef := bundle.TxReference{
-			From: args.ExecutionPlan.Steps[i].From,
-			Hash: args.ExecutionPlan.Steps[i].Hash,
-		}
-		transactions[txRef] = tx
-		steps[i] = bundle.NewTxStep(txRef).WithFlags(args.ExecutionPlan.Flags)
+		transactions[refs[i]] = tx
 		totalGas += tx.Gas()
```
LuisPH3: Please use https://github.com/0xsoniclabs/sonic/blob/b20c22a665fbc9147596577e0735666d2073e9c8/gossip/blockproc/bundle/gas.go#L30

We will have to export it, but it would be better not to duplicate code.

api/sonicapi/submit_bundle.go
```diff
@@ -54,59 +54,44 @@ func (a *PublicBundleAPI) SubmitBundle(
...
-	if latest < earliest {
-		return common.Hash{}, fmt.Errorf("latest block number cannot be smaller than earliest block number")
+	execPlan, err := ToBundleExecutionPlan(args.ExecutionPlan)
+	if err != nil {
+		return common.Hash{}, fmt.Errorf("invalid execution plan: %w", err)
 	}
 
-	if len(args.ExecutionPlan.Steps) != len(args.SignedTransactions) {
+	refs := execPlan.Root.GetTransactionReferencesInReferencedOrder()
+	if len(refs) != len(args.SignedTransactions) {
```
LuisPH3, line 74: Heads up: there is a possibility that one unique transaction is referenced multiple times in a plan. This case will break the indexing in the prepare method as well, I think we should tackle it in a later PR, not necessarily before the release.

api/sonicapi/submit_bundle.go
```diff
@@ -21,13 +21,13 @@ import (
 	"fmt"
 
 	"github.com/0xsoniclabs/sonic/api/ethapi"
+	evmcore "github.com/0xsoniclabs/sonic/evmcore"
```
LuisPH3: nit: remove alias

LuisPH3 (dismissed):
This method adds the tooling to submit bundles from transactions signed by third party wallets.

LuisPH3 (dismissed):

LuisPH3 (approved):

PR #907, opened 2026-04-27
Add GetPredecessor to store [gfile bundles]

facuMH:
This PR adds a new method `GetPredecessorBundleHistoryHash` to the Store, that retrieves the block number and hash of the very first entry in the 'h' key space of the processed bundles table.

**Motivation**
After pruning, the store retains one extra predecessor 'h' entry just before the oldest in-range block. This predecessor hash is needed to independently verify the hash chain starting at the oldest retained block (i.e., hash(K) = Keccak256(hash(K-1) || xor(...) || K)). Previously, there was no public API to retrieve this predecessor entry directly.

This PR depends on https://github.com/0xsoniclabs/sonic/pull/893


facuMH:
not needed.

PR #903, opened 2026-04-27
Add support for tracking transaction fees for bundles

HerbertJordan:
WIP: please ignore

PR #910, opened 2026-04-27
Add utilities to identify and parse fee-charging transactions

HerbertJordan:
This PR introduces two new utilities for subsidized transactions:
- `IsFeeChargeTransaction` to identify internal transactions introduced by the block processor to charge fees for sponsored transactions
- `ParseFeeChargeAmount` to extract the amount of fees charged for a sponsored transaction from a fee-charge transaction

Both utilities are preparations for improving the mechanism of tracking transaction fees in Sonic. For a full integration see #903.


LuisPH3 (approved):
The PR introduces utilities to identify sponsorship paying transactions and the amount to pay.

PR #911, opened 2026-04-27
Add parser for epoch sealing transactions

HerbertJordan:
This PR adds a utility for decoding the arguments provided to `sealEpoch` transactions at the end of each epoch.

This utility provides a first building block for building end-to-end integration tests verifying the content of those arguments.

For a full example making use of this function see #903.


LuisPH3 (approved):
This Pr introduces utilities to extract the arguments passed to the driver contract on an epoch sealing transaction. This method allows reading the values after the transaction as been executed and even in historic transactions.

PR #914, opened 2026-04-28
Update transaction fee tracking to cover bundles

HerbertJordan:
This PR adds support for the correct aggregation of transaction fees when processing bundles.

In order to facilitate the distribution of transaction fees among validators, a record of charged fees needs to be maintained by the sonic client. This book-keeping is conducted by the [DriverTxListener](https://github.com/0xsoniclabs/sonic/blob/8d1f49c73de18510c1950c0503c75ff46146758a/gossip/blockproc/drivermodule/driver_txs.go#L62).

The `DriverTxListener` is the only implementation (besides a mock) of the [TxListener](https://github.com/0xsoniclabs/sonic/blob/8d1f49c73de18510c1950c0503c75ff46146758a/gossip/blockproc/interface.go#L36) interface offered by the block formation process in the `c_block_callback.go` to inform about new transactions being accepted in a block ([here](https://github.com/0xsoniclabs/sonic/blob/8d1f49c73de18510c1950c0503c75ff46146758a/gossip/c_block_callbacks.go#L489)). 

### Old Behavior
In the current client, a record of the collected gas fees is kept as follows:
- after each block is completed, a call back to the `OnNewReceipt` function for each included transaction is conducted ([here](https://github.com/0xsoniclabs/sonic/blob/8d1f49c73de18510c1950c0503c75ff46146758a/gossip/c_block_callbacks.go#L489)). One of the parameters is the ID of the validator that proposed the accepted transaction in one of its events. 
- the fees  charged for the respective transaction are then added to the validator's originated transaction fees ([here](https://github.com/0xsoniclabs/sonic/blob/8d1f49c73de18510c1950c0503c75ff46146758a/gossip/blockproc/drivermodule/driver_txs.go#L170-L171))
- when sealing an epoch, the total sum of originated fees is reported to the SFC to handle the fee distribution ([here](https://github.com/0xsoniclabs/sonic/blob/8d1f49c73de18510c1950c0503c75ff46146758a/gossip/blockproc/drivermodule/driver_txs.go#L112-L132))

With the introduction of features like sponsored transactions and bundles, this mechanism is no longer sufficient. Sponsored transactions are payed indirectly, not derivable from the transactions receipt, and transactions ending up in blocks are not directly included in events but encoded in envelopes carried by events. Thus, the fee aggregation mechanism needs to updated.

### New Behavior
In order to preserve backward compatibility, this PR preserves the existing code for Pre-Brio configurations. However, with Brio, the following changes are added:
- callbacks notifying about new accpeted transactions now happen during the processing of each individual proposed transaction in the `processUserTransactions` function ([here](https://github.com/0xsoniclabs/sonic/blob/1c9f5392506f09d5e838d23c0dcb5734c28138e0/gossip/c_block_callbacks.go#L671))
- in the callback, the effective fees are [computed](https://github.com/0xsoniclabs/sonic/blob/1c9f5392506f09d5e838d23c0dcb5734c28138e0/gossip/blockproc/drivermodule/driver_txs.go#L216) for each accepted transaction using a new [ComputeEffectiveFee](https://github.com/0xsoniclabs/sonic/blob/1c9f5392506f09d5e838d23c0dcb5734c28138e0/gossip/blockproc/drivermodule/driver_txs.go#L239) function; this function identifies fee-payment transactions for sponsored transactions and extracts the amount of fees charged
- by integrating the callback into the `processUserTransactions` accepted transactions can be mapped locally to the validators who have proposed the transaction causing inclusion of the accepted transaction ([here](https://github.com/0xsoniclabs/sonic/blob/1c9f5392506f09d5e838d23c0dcb5734c28138e0/gossip/c_block_callbacks.go#L671))
- the old reporting method is disabled by enabling Brio ([here](https://github.com/0xsoniclabs/sonic/blob/1c9f5392506f09d5e838d23c0dcb5734c28138e0/gossip/c_block_callbacks.go#L495))

The result should be a accurate fee tracking for bundled and sponsored transactions and the preservation of the backward compatibility with pre-Brio handling.

### Gas Refunds for Validators
The new implementation of the gas-fee tracking in `driver_tx.go` does not refunds for excessive gas. The computation of that value for bundles is beyond the scope of this task, and the impact of dropping this feature is under investigation. See https://github.com/0xsoniclabs/sonic-admin/issues/748.

### DIFF
--- a/gossip/blockproc/drivermodule/driver_txs.go
+++ b/gossip/blockproc/drivermodule/driver_txs.go
@@ -215,8 +215,8 @@ func (p *DriverTxListener) onNewAcceptedTransactionInternal(
 ) {
 	fee, err := ComputeEffectiveFee(tx, r)
 	if err != nil {
-		// if there is an overflow in the fee computation, the safe default
-		// is not not attribute it to any validator to not pay out any fees.
+		// If there is an overflow in the fee computation, the safe default
+		// is to avoid attributing it to any validator, so no fees are paid out.
 		log.Warn("overflow in fee computation", "tx", tx.Hash(), "usedGas", r.GasUsed, "gasPrice", r.EffectiveGasPrice, "err", err)
 		return
 	}
@@ -235,7 +235,7 @@ func (p *DriverTxListener) onNewAcceptedTransactionInternal(
 // transaction and its receipt. For regular transactions, this is simply
 // gasUsed * effectiveGasPrice. For fee charge transactions, the fee is
 // extracted from the transaction input data. If the calculation leads to a
-// 256-bit overflow, a fee of 0 is returned.
+// 256-bit overflow, the function returns a non-nil error and a nil fee.
 func ComputeEffectiveFee(
 	tx *types.Transaction,
 	r *types.Receipt,
--- a/gossip/blockproc/drivermodule/driver_txs.go
+++ b/gossip/blockproc/drivermodule/driver_txs.go
@@ -17,6 +17,7 @@
 package drivermodule
 
 import (
+	"fmt"
 	"io"
 	"math"
 	"math/big"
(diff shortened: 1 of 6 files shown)


HerbertJordan:
replaced by #926

PR #915, opened 2026-04-28
Fix issues preventing prepare bundle from having reproducible plan hash

LuisPH3:
This PR fixes two issues:

- `ethapi.transactionArgs` -> `types.Transaction` creates legacy tx when possible. They need to be at least access list to be added to a bundle
- proposal to plan constructed plans (allOff(A)  while RPCPlan to plan creates (A) when converted into `bundle.ExecutionPlan` 

Although generated bundles could be executed, these issues prevented hashes from being equivalent, and therefore from validating the workflow.

### DIFF
--- a/api/sonicapi/execution_proposal_test.go
+++ b/api/sonicapi/execution_proposal_test.go
@@ -30,6 +30,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/params"
 	"github.com/holiman/uint256"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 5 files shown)


api/sonicapi/execution_proposal_test.go
```diff
@@ -798,3 +799,273 @@ func Test_transform(t *testing.T) {
...
+				RPCExecutionPlanGroup: RPCExecutionPlanGroup{
+					Steps: []any{
+						RPCExecutionStepProposal{
+							TransactionArgs: ethapi.TransactionArgs{
+								From:  &address1,
+								To:    &common.Address{123},
+								Nonce: rpctest.ToHexUint64(1),
+								Gas: rpctest.ToHexUint64(21000 +
+									// NOTE: builder adds marker costs
+									params.TxAccessListAddressGas + params.TxAccessListStorageKeyGas),
```
jenikd: Nit this is repeated for almost every subtest, maybe put it into the variable into tis test

api/sonicapi/prepare_bundle_test.go
```diff
@@ -912,12 +908,10 @@ func Test_PrepareBundle_SingleChildGroup_Plain_NotUnwrapped(t *testing.T) {
 	require.NoError(t, err)
 	require.Len(t, result.Transactions, 1)
 
-	// Plain group must not collapse its single child.
+	// Plain group must collapses its single child.
```
jenikd: Typo
```suggestion
	// Plain group must collapse its single child.
```

api/sonicapi/bundles_api_intergration_test.go
```diff
@@ -0,0 +1,343 @@
...
+}
+
+func newAccount(t testing.TB) account {
+	key, err := crypto.GenerateKey()
+	require.NoError(t, err)
+	return account{
+		address: crypto.PubkeyToAddress(key.PublicKey),
+		key:     key,
+	}
+}
```
jenikd: Usage of this could be replaced by using the `rpctest.Wallet`

jenikd (changes requested):
Looks good, just few small nits

api/sonicapi/bundles_api_intergration_test.go
LuisPH3: this test is a different PR

jenikd (approved):

PR #916, opened 2026-04-28
Add test covering the interaction of `sonic_prepareBundle` and `sonic_submitBundle`

LuisPH3:
This test covers the integration of `PrepareBundle` and `SubmitBundle`, ensuring that a bundle prepared by the API can be successfully submitted and recognized by the transaction pool as a valid bundle envelope with the correct execution plan hash.

### DIFF
--- a/api/sonicapi/bundles_api_integration_test.go
+++ b/api/sonicapi/bundles_api_integration_test.go
@@ -87,12 +87,14 @@ func Test_bundlesRPC_PrepareAndSubmit_PoolsValidBundles(t *testing.T) {
 					{
 						"from": "ACCOUNT1",
 						"to": "ACCOUNT2",
-						"nonce": "0x0"
+						"nonce": "0x0",
+						"gas": "0x5208"
 					},
 					{
 						"from": "ACCOUNT2",
 						"to": "ACCOUNT1",
-						"nonce": "0x0"
+						"nonce": "0x0",
+						"gas": "0x5208"
 					}
 				]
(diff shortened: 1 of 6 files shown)


api/sonicapi/bundles_api_integration_test.go
```diff
@@ -0,0 +1,285 @@
...
+func Test_bundlesRPC_PrepareAndSubmit_PoolsValidBundles(t *testing.T) {
+
+	// This test covers the integration of PrepareBundle and SubmitBundle,
+	// ensuring that a bundle prepared by the API can be successfully submitted
+	// and recognized by the transaction pool as a valid bundle envelope
+	// with the correct execution plan hash.
+
+	accounts := make(map[string]*rpctest.Wallet)
+	accounts["ACCOUNT1"] = rpctest.NewWallet(t)
+	accounts["ACCOUNT2"] = rpctest.NewWallet(t)
```
jenikd, line 44: Nit these strings are on many places, it would be harder to maintain this test. Can use variable for that

jenikd (dismissed):

api/sonicapi/bundles_api_integration_test.go
LuisPH3: the thing is that these strings are hardcoded in the json string. i would have to hardcode anther replacement... ?
 these strings declare the placeholders that can be used in the json templates.

jenikd (approved):

PR #919, opened 2026-04-29
Remove old RPC bundle types and refactor

jenikd:
This PR removes RPCExecutionPlan structure as it was replaced by the composable version. It also moves sanitizeBlockRange function and it's test into the the common file as it is used from PrepareBundle and SubmitBundle RPC functions


api/sonicapi/bundle_api.go
```diff
@@ -28,33 +29,48 @@ import (
 // therefore an upper bound is introduced.
 const MaxNumEstimableTransactions = 16
```
LuisPH3: nit: move to estimate.go?

LuisPH3 (dismissed):

api/sonicapi/bundle_api.go
jenikd: Moved

LuisPH3 (approved):

PR #920, opened 2026-04-29
Add integration tests for semantics of bundles with brio and bundles flags

LorenzSchueler:
This PR adds integration tests that ensure that bundle semantics are interpreted starting with brio and their execution is controlled by the bundle feature flag.

The tests also uncovered that the emitter was checking for the wrong flag, and the pool was not checking for the flag at all. This has been fixed and is also included in this PR.

### DIFF
--- a/evmcore/bundle_integration.go
+++ b/evmcore/bundle_integration.go
@@ -52,6 +52,11 @@ func newBundlesChecker(
 		stateDb: stateDB,
 	}
 	return func(tx *types.Transaction) bundlePoolStatus {
+		// Before Brio, bundle envelopes are regular transactions and must not
+		// be subject to bundle-specific filtering in the pool.
+		if !chain.CurrentRules().Upgrades.Brio {
+			return bundlePending
+		}
 		bundleState := GetBundleState(adapter, stateDB, tx)
 		if bundleState.Executable {
 			return bundlePending
--- a/evmcore/bundle_integration.go
+++ b/evmcore/bundle_integration.go
@@ -52,11 +52,6 @@ func newBundlesChecker(
 		stateDb: stateDB,
 	}
 	return func(tx *types.Transaction) bundlePoolStatus {
-		// Before Brio, bundle envelopes are regular transactions and must not
-		// be subject to bundle-specific filtering in the pool.
-		if !chain.CurrentRules().Upgrades.Brio {
-			return bundlePending
-		}
 		bundleState := GetBundleState(adapter, stateDB, tx)
 		if bundleState.Executable {
 			return bundlePending
--- a/evmcore/tx_list.go
+++ b/evmcore/tx_list.go
@@ -437,11 +437,14 @@ func (l *txList) Filter(
 		}
 
 		// bundle envelopes with valid nonces but non-pending status are demoted
-		demotedEnvelopes := l.txs.filter(func(tx *types.Transaction) bool {
-			return bundle.IsEnvelope(tx) &&
-				tx.Nonce() < firstInvalidNonce &&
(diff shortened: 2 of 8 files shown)


gossip/emitter/emitter.go
```diff
@@ -363,7 +363,7 @@ func removeBundleOnlyTxs(
 	upgrades opera.Upgrades,
 	pendingTxs map[common.Address]types.Transactions) {
 
-	if !upgrades.TransactionBundles {
+	if !upgrades.Brio {
```
LuisPH3, line 366: I think that we should do both.
This check exists for an scenario in which TransactionBundles are enabled but brio is not. 
This makes me think that we should validate in the rules for that.
if so, lets create an issue about that. 

@HerbertJordan wdyt?

gossip/emitter/emitter_test.go
```diff
@@ -611,7 +611,7 @@ func TestEmitter_fillExtraData_EncodesExpectedData(t *testing.T) {
 	}
 }
 
-func TestRemoveBundleOnlyTxs_OnlyFiltersTxIfFeatureIsEnabled(t *testing.T) {
+func TestRemoveBundleOnlyTxs_OnlyFiltersTxIfBrioIsEnabled(t *testing.T) {
```
LuisPH3, line 614: I am not sure about this test, i think we need to test both as mentioned before.

gossip/emitter/emitter.go
LorenzSchueler: Bundles require a hard fork. Since brio is the next hard fork, it is not possible to enable bundle before brio. And the filtering of bundles is introduced with brio, it does not matter whether they are enabled and get executed or not.

LuisPH3: nevermind, this code only filters bundle only. not bundles.

LuisPH3 (dismissed):

evmcore/bundle_integration.go
```diff
@@ -52,6 +52,11 @@ func newBundlesChecker(
 		stateDb: stateDB,
 	}
 	return func(tx *types.Transaction) bundlePoolStatus {
+		// Before Brio, bundle envelopes are regular transactions and must not
+		// be subject to bundle-specific filtering in the pool.
+		if !chain.CurrentRules().Upgrades.Brio {
+			return bundlePending
+		}
```
HerbertJordan: When reaching this code the given transaction was already classified as a bundle -- which it should not be. It should be treated like a regular transaction, checked against nonce and balance requirements.

Should the Brio flag be checked [here](https://github.com/0xsoniclabs/sonic/blob/dff1e59d53e7b5f04582ca41e0681e60bcef7413/evmcore/tx_list.go#L408) and [here](https://github.com/0xsoniclabs/sonic/blob/dff1e59d53e7b5f04582ca41e0681e60bcef7413/evmcore/tx_pool.go#L1532) instead, to enable default handling if disabled?

HerbertJordan (dismissed):
Looks good, thanks for adding this.

As commented, I am not certain about the rule check in the transaction pool. Please have a look.

evmcore/tx_list.go
```diff
@@ -387,6 +388,7 @@ func (l *txList) Filter(
 	gasLimit uint64,
 	isSponsored utils.TransactionCheckFunc,
 	evaluateBundleStatus func(*types.Transaction) bundlePoolStatus,
+	upgrades opera.Upgrades,
```
LuisPH3, line 391: 👍

LuisPH3 (dismissed):

HerbertJordan (dismissed):

LuisPH3 (approved):

PR #921, opened 2026-04-29
Add bundle enable and disable test

simonlechner:
This PR adds an integration test to ensure that bundles can be enabled and disabled.

### DIFF
--- a/tests/bundles/network_rules_test.go
+++ b/tests/bundles/network_rules_test.go
@@ -45,21 +45,13 @@ func TestBundles_BundlesCanBeEnabledAndDisabledStartingFromBrio(t *testing.T) {
 	buildEnvelope := func(t *testing.T) *types.Transaction {
 		t.Helper()
 		senderA := tests.MakeAccountWithBalance(t, net, big.NewInt(1e18))
-		senderB := tests.NewAccount()
-		addressB := senderB.Address()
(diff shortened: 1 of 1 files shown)


tests/bundles/network_rules_test.go
```diff
@@ -0,0 +1,105 @@
...
+			WithSigner(signer).
+			SetEarliest(block).
+			AllOf(
+				Step(t, net, senderA, &types.AccessListTx{
+					To:    &addressB,
+					Value: big.NewInt(1),
+				}),
+			).
+			BuildEnvelopeBundleAndPlan()
+		return envelope
```
LorenzSchueler: nit: This can be simplified a bit. This is then also more in line with the other existing tests.

```suggestion
		senderA := tests.MakeAccountWithBalance(t, net, big.NewInt(1e18))
		block, err := client.BlockNumber(t.Context())
		require.NoError(t, err)
		return bundle.NewBuilder().
			WithSigner(signer).
			SetEarliest(block).
			AllOf(Step(t, net, senderA, &types.AccessListTx{})).
			Build()
```

LorenzSchueler (dismissed):

LorenzSchueler (approved):

LuisPH3 (approved):

PR #923, opened 2026-04-29
Verify bundle history hash integrity on genesis import

facuMH:
This PR depends on https://github.com/0xsoniclabs/sonic/pull/893

**Problem**
Previously, genesis export stored only the latest bundle history hash, and import blindly overwrote it after replaying entries. This meant a corrupted or incomplete export could be imported without detection — the hash was simply force-set rather than independently reproduced and verified.

**Solution**
Export now includes two history hash checkpoints — earliest and latest — and import verifies the hash chain by replaying it from scratch:
- Export: Serializes a `BundleGenesisHistoryHashes` struct containing both checkpoints instead of just the last `HistoryHash`.
- Import: Seeds the hash chain with the _earliest hash_ (the entry just before the retention window).
Replays every block through latest, calling `AddProcessedBundles` with the exported entries (or nil for blocks without bundles).
Verifies the reproduced hash matches the latest checkpoint after all blocks.

This PR finishes https://github.com/0xsoniclabs/sonic-admin/issues/706

### DIFF
--- a/cmd/sonictool/app/app_test.go
+++ b/cmd/sonictool/app/app_test.go
@@ -242,44 +242,54 @@ func TestSonicTool_genesis_ExportImport_WithBundles(t *testing.T) {
 	net := tests.StartIntegrationTestNet(t, tests.IntegrationTestNetOptions{
 		Upgrades: &upgrades,
 	})
-
-	bundleHash, originalInfo := runBundle(t, net)
-
 	client, err := net.GetClient()
 	require.NoError(t, err)
 
-	// check that the bundle is still there after the export-import process
-	infoBeforeRestart, err := bundles.GetBundleInfo(t.Context(), client.Client(), bundleHash)
+	// run a bundle and check it can be queried
+	bundleHash1, originalInfo1 := runBundle(t, net)
+	info1, err := bundles.GetBundleInfo(t.Context(), client.Client(), bundleHash1)
 	require.NoError(t, err)
-	require.Equal(t, originalInfo, infoBeforeRestart,
+	require.Equal(t, originalInfo1, info1,
 		"bundle info mismatch after genesis export-import")
 
-	// close client before restarting the net
+	// restart the net
 	client.Close()
-
 	require.NoError(t, net.RestartWithExportImport())
-
 	client, err = net.GetClient()
 	require.NoError(t, err)
 
 	// check that the bundle is still there after the export-import process
-	infoAfterRestart, err := bundles.GetBundleInfo(t.Context(), client.Client(), bundleHash)
+	info1, err = bundles.GetBundleInfo(t.Context(), client.Client(), bundleHash1)
 	require.NoError(t, err)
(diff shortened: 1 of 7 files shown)


cmd/sonictool/app/app_test.go
```diff
@@ -261,13 +261,26 @@ func TestSonicTool_genesis_ExportImport_WithBundles(t *testing.T) {
...
+
+	// run enough blocks to make sure that if the history hash is pruned.
+	generateNBlocks(t, net, int(bundle.MaxBlockRange)+10)
+
+	require.NoError(t, net.RestartWithExportImport())
+
+	client, err = net.GetClient()
+	require.NoError(t, err)
+
+	// check that the bundle is still there after the export-import process
```
HerbertJordan:
```suggestion
	// check that the bundle is no longer there after the export-import process
```

cmd/sonictool/app/app_test.go
```diff
@@ -261,13 +261,26 @@ func TestSonicTool_genesis_ExportImport_WithBundles(t *testing.T) {
...
+	_, err = bundles.GetBundleInfo(t.Context(), client.Client(), bundleHash)
```
HerbertJordan: Well .. the bundle is already not there before the export import process. Maybe add another bundle just before re-exporting it and also  check for that one to be preserved?

gossip/apply_genesis.go
```diff
@@ -160,29 +158,64 @@ func (s *Store) ApplyGenesis(g genesis.Genesis) (err error) {
...
+			"oldestBlockNum", hh.Oldest.BlockNumber, "oldestHash", hh.Oldest.Hash,
+			"latestBlockNum", hh.Latest.BlockNumber, "latestHash", hh.Latest.Hash)
+
+		// When the oldest block has bundle entries, the hash chain was
+		// never pruned for that block, so we can replay from it with the
+		// hash starting at zero. When the oldest block has no entries,
+		// pruning occurred and the oldest hash is a predecessor: seed the
+		// chain with it and start replaying from the next block.
+		startBlock := hh.Oldest.BlockNumber
+		if _, oldestHasBundles := bundlesByBlock[hh.Oldest.BlockNumber]; !oldestHasBundles {
```
HerbertJordan: This assumes that there are bundles in every block, but this is unlikely.

Consider the case where be back up at block 9000. The oldest hash would be ~9000-1024. If this block happens to have not reported execution plans you would consider it a never-pruned chain and skip the initial hash setup.

I think you need to find another way to determine whether pruning has happend. Or you might get away by always setting the oldest value, even if it is 0.

HerbertJordan (changes requested):
Please have a look and the issue with determining pruned or non-pruned states as described in a comment.

gossip/apply_genesis.go
LorenzSchueler: This also confused me a bit. I think we can fix it by checking the size of the range.

HerbertJordan (approved):

PR #924, opened 2026-04-29
Include gas pool in in state processor snapshots

LorenzSchueler:
This PR adds the gas pool to the fields covered by the snapshots which are used to roll back bundles. This is needed to ensure that blocks are not underutilized when load is high.

### DIFF
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -529,6 +529,7 @@ func (b *bundleTransactionRunner) CreateSnapshot() int {
 		trueTxOffset:                   b.trueTxOffset,
 		processedTransactionListLength: len(b.processedTransactions),
 		usedGas:                        *b.ctxt.usedGas,
+		gasPool:                        *b.ctxt.gasPool,
 	}
 	b.snapshots = append(b.snapshots, snapshot)
 	return len(b.snapshots) - 1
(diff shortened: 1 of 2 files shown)


LuisPH3 (dismissed):

evmcore/state_processor.go
```diff
@@ -529,6 +529,7 @@ func (b *bundleTransactionRunner) CreateSnapshot() int {
 		trueTxOffset:                   b.trueTxOffset,
 		processedTransactionListLength: len(b.processedTransactions),
 		usedGas:                        *b.ctxt.usedGas,
+		gasPool:                        *b.ctxt.gasPool,
```
HerbertJordan: The `GasPool` has a [Snapshot](https://github.com/ethereum/go-ethereum/blob/01036bed831186d892a3e3e35b32fcf304d4cc29/core/gaspool.go#L90) function and a [Set](https://github.com/ethereum/go-ethereum/blob/01036bed831186d892a3e3e35b32fcf304d4cc29/core/gaspool.go#L99) function to restore it. Please use those to protect against future modifications in the go-ethereum code.

HerbertJordan (changes requested):
Looks good. Please update the code to use the go-ethereum infrastructure for deep-copying and restoring the gas-pool state.

LuisPH3 (approved):

HerbertJordan (approved):

PR #926, opened 2026-04-29
Update transaction fee tracking to cover bundles

HerbertJordan:
This PR adds support for the correct aggregation of transaction fees when processing bundles.

In order to facilitate the distribution of transaction fees among validators, a record of charged fees needs to be maintained by the sonic client. This book-keeping is conducted by the [DriverTxListener](https://github.com/0xsoniclabs/sonic/blob/8d1f49c73de18510c1950c0503c75ff46146758a/gossip/blockproc/drivermodule/driver_txs.go#L62).

### Old Behavior
A record of the collected gas fees is kept as follows:
- after each block is completed, a call back to the `OnNewReceipt` function for each included transaction is conducted ([here](https://github.com/0xsoniclabs/sonic/blob/8d1f49c73de18510c1950c0503c75ff46146758a/gossip/c_block_callbacks.go#L489)). One of the parameters is the ID of the validator that proposed the accepted transaction in one of its events. 
- the fees  charged for the respective transaction are then added to the validator's originated transaction fees ([here](https://github.com/0xsoniclabs/sonic/blob/8d1f49c73de18510c1950c0503c75ff46146758a/gossip/blockproc/drivermodule/driver_txs.go#L170-L171))
- when sealing an epoch, the total sum of originated fees is reported to the SFC to handle the fee distribution ([here](https://github.com/0xsoniclabs/sonic/blob/8d1f49c73de18510c1950c0503c75ff46146758a/gossip/blockproc/drivermodule/driver_txs.go#L112-L132))

With the introduction of features like sponsored transactions and bundles, this mechanism is no longer sufficient. Sponsored transactions are payed indirectly, not derivable from the transactions receipt, and transactions ending up in blocks are not directly included in events but encoded in envelopes carried by events. Thus, the fee aggregation mechanism needs to updated.

### New Changes
The following changes are enabled with the Brio flag:
- indirect causes are tracked in the `c_block_callback.go`, such that transactions introduced in blocks via subsidies or bundles are correctly associated to the respective validator
- in the [OnNewReceipt](https://github.com/0xsoniclabs/sonic/blob/80d3fa1a6bf4d995058ac825e36a81396958020a/gossip/blockproc/drivermodule/driver_txs.go#L152) callback, the computation rules are adapted post-Brio; the effective fees are [computed](https://github.com/0xsoniclabs/sonic/blob/80d3fa1a6bf4d995058ac825e36a81396958020a/gossip/blockproc/drivermodule/driver_txs.go#L228) for each accepted transaction using a new [ComputeEffectiveFee](https://github.com/0xsoniclabs/sonic/blob/80d3fa1a6bf4d995058ac825e36a81396958020a/gossip/blockproc/drivermodule/driver_txs.go#L257) function; this function identifies fee-payment transactions for sponsored transactions and extracts the amount of fees charged

The result should be a accurate fee tracking for bundled and sponsored transactions and the preservation of the backward compatibility with pre-Brio handling.

### Gas Refunds for Validators
The new implementation of the gas-fee tracking in `driver_tx.go` does not refunds for excessive gas. The computation of that value for bundles is beyond the scope of this task, and the impact of dropping this feature is considered low, but under investigation. See https://github.com/0xsoniclabs/sonic-admin/issues/748.

### DIFF
--- a/gossip/blockproc/drivermodule/driver_txs.go
+++ b/gossip/blockproc/drivermodule/driver_txs.go
@@ -229,11 +229,15 @@ func (p *DriverTxListener) onNewReceiptPostBrioInternal(
 	if err != nil {
 		// If there is an error in the fee computation, the safe default
 		// is to avoid attributing it to any validator, so no fees are paid out.
-		log.Warn("error in fee computation", "tx", tx.Hash(),
-			"usedGas", r.GasUsed, "gasPrice", r.EffectiveGasPrice,
-			"blobGasUsed", r.BlobGasUsed, "blobGasPrice", r.BlobGasPrice,
-			"err", err,
-		)
+		if r == nil {
+			log.Warn("error in fee computation", "tx", tx.Hash(), "err", err)
+		} else {
+			log.Warn("error in fee computation", "tx", tx.Hash(),
+				"usedGas", r.GasUsed, "gasPrice", r.EffectiveGasPrice,
+				"blobGasUsed", r.BlobGasUsed, "blobGasPrice", r.BlobGasPrice,
+				"err", err,
+			)
+		}
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


LuisPH3 (approved):
The PR introduces a new mechanism to collect the transactions to attribute fees to after the execution.  There are 2 main contributions:
- A mechanism to identify the causing transaction when transactions are not directly originated by an event.
- A new code path to attributing fees to the validator which proposed the causing Tx

This code includes the appropriate fork guard for the fees collection, and the collection of the causing Tx, which otherwise would produce unexpected forks.

LorenzSchueler (approved):
Looks Good

PR #927, opened 2026-04-29
Add function to compute effective transaction fees

HerbertJordan:
This PR introduces a helper function to calculate the effective fees that got charged for a transaction on the network.

In addition to the regular fees that can be derived from a receipt, internal fee-payment transactions introduced for sponsored transactions are covered as well.

### DIFF
--- a/gossip/blockproc/drivermodule/driver_txs.go
+++ b/gossip/blockproc/drivermodule/driver_txs.go
@@ -17,6 +17,7 @@
 package drivermodule
 
 import (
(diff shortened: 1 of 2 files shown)


LuisPH3 (approved):
The PR adds a utility to compute the fees charged in Sonic terms, where an internal transaction paying for a sponsorship does not have an effect on sonic fees equivalent to the gas costs of itself, but the value of the sponsorship instead.

PR #928, opened 2026-04-29
Add utility to use the same available hardforks in all tests iterating over hardforks

LuisPH3:
This Pr introduces `opera.GetAllHardForksInOrder()`, which returns a list of named hardforks. Some tests iterating over hardforks are stateful and require those to happen in order. For this reason the tool has been written to return pairs (name, `opera.Upgrades`) in order. Tests not requiring order are not compromised. 

Most of the tests were trivially migrated, but some tests had to be refactored to account for different behavior. Different commits have been done for these to facilitate review.


LorenzSchueler (dismissed):
Some tests are now covering more hard forks than before, but if they still pass it should be fine.

tests/single_proposer/single_proposer_protocol_test.go
```diff
@@ -33,13 +33,7 @@ import (
 
 func TestSingleProposerProtocol_CanProcessTransactions(t *testing.T) {
 
-	upgrades := map[string]opera.Upgrades{
-		// Single-proposer protocol has been rolled out with Allegro.
```
simonlechner, line 37: This test is not indent to work with sonic, but if it passes I think this is not an issue.

simonlechner (approved):
Nice refactor, with more hardforks in the future this makes a lot of sense.

tests/single_proposer/single_proposer_protocol_test.go
LuisPH3: yes, this is intended. With sonic, SBP on/off is equivalent, the flag is ignored. But the expectations are still valid. Since the test is fast, I choose to reduce boilerplate.

LorenzSchueler (approved):

HerbertJordan (commented):
I trust other reviewers to have checked this properly. No need for me to look it through.

PR #930, opened 2026-04-29
Fix Flaky TestRandao_randaoIntegrationTest 

LuisPH3:
Add wait around time sensitive test expectation. Not all nodes process blocks at the same time.

### DIFF
--- a/tests/randao_test.go
+++ b/tests/randao_test.go
@@ -72,7 +72,10 @@ func TestRandao_randaoIntegrationTest(t *testing.T) {
 				defer cancel()
 				err = WaitFor(timeout, func(ctx context.Context) (bool, error) {
 					block, err = client.BlockByNumber(ctx, receipt.BlockNumber)
-					if err != nil && err != ethereum.NotFound {
+					if err == ethereum.NotFound {
+						return false, nil
+					}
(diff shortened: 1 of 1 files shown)


LorenzSchueler (dismissed):

tests/randao_test.go
```diff
@@ -64,7 +67,16 @@ func TestRandao_randaoIntegrationTest(t *testing.T) {
...
 
-				block, err := client.BlockByNumber(t.Context(), receipt.BlockNumber)
+				var block *types.Block
+				timeout, cancel := context.WithTimeout(t.Context(), 3*time.Second)
+				defer cancel()
+				err = WaitFor(timeout, func(ctx context.Context) (bool, error) {
+					block, err = client.BlockByNumber(ctx, receipt.BlockNumber)
+					if err != nil && err != ethereum.NotFound {
+						return false, err
+					}
```
HerbertJordan: Should this be ..
```suggestion
					if err == ethereum.NotFound {
						return false, nil
					}
					if err != nil {
						return false, err
					}
```

LuisPH3: thanks

LorenzSchueler (approved):

PR #934, opened 2026-04-29
Add execution plan state hash into epoch hash

HerbertJordan:
This PR expands the EpochState to cover additional information:
- the hash of the block before the block containing the sealing transactions
- the hash of the execution plan history of the block before the block containing the sealing transactions
- the hashes of the pre-internal transactions used to seal the epoch

The update introduces a new EpochState version (version V2) expanding the previous V1 version by three extra fields. The decoder of the new version is customized to be backward compatible. In particular, epoch states serialized as V1 should be readable into the new V2 instance. This should preserve backward compatibility with existing state DB instances and G files.

Manual tests to be conducted:
- [x] run in Norma to check compatibility with older Sonic client versions (e.g. 2.1.6)
- [x] check that G-files can be loaded
- [x] check that the gossip store created by an older client can be opened and used for continuing syncing


LuisPH3 (approved):
The PR introduces a new epoch state format, including 3 new fields.
- hash of the last block
- hashes of the pre-internal txs if any.
- bundles history hash accumulated during the epoch.

Adding this information to the epoch modifies the epoch hash, and therefore adds early detection of forks caused by any of these 3 sources. 

The serialization has been extended to be forwards compatible, but not backwards, this forces nodes running 2.1.x versions to update before being able to load a genesis generated with a 2.2 version of the client.

LuisPH3:
-  check that the gossip store created by an older client can be opened and used for continuing syncing

The image here shows a client being restated with version 2.1.6 (~9:25), and restarted with a build of this branch (~10:10) using the same datadir. 2.1.6 period included at least two epoch changes. 
The metrics show the node being able to process new blocks, and connect to peers.
The client built with this branch sealed an epoch some time later. New epochs are produced. 
 

image

LuisPH3:
The branch can load `sonic-75130-pruned.g`, sync, and continues producing blocks, and sealing epochs.

PR #937, opened 2026-04-30
Add time limit to bundle

HerbertJordan:
This PR adds an additional option for a real-time time limit to bundles.

### DIFF
--- a/api/sonicapi/execution_plan.go
+++ b/api/sonicapi/execution_plan.go
@@ -112,8 +112,9 @@ func ToBundleExecutionPlan(rpcPlan RPCExecutionPlanComposable) (bundle.Execution
 	}
 
 	return bundle.ExecutionPlan{
-		Range: rpcPlan.BlockRange.toBundleBlockRange(),
-		Root:  root,
+		Range:  rpcPlan.BlockRange.toBundleBlockRange(),
+		Root:   root,
+		Period: bundle.MakeUnrestrictedTimePeriod(),
 	}, nil
 }
 
--- a/evmcore/bundle_precheck_test.go
+++ b/evmcore/bundle_precheck_test.go
@@ -186,7 +186,7 @@ func Test_GetBundleState_ToEarlyForTimePeriod_ReturnsTemporaryBlocked(t *testing
 	require.Equal(t, state, makeTemporaryBlockedState("bundle targets future time period"))
 }
 
-func Test_GetBundleState_HasBundleHasAlreadyBeenProcessed_ReturnsPermanentlyBlocked(t *testing.T) {
+func Test_GetBundleState_BundleHasAlreadyBeenProcessed_ReturnsPermanentlyBlocked(t *testing.T) {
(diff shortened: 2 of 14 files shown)


evmcore/bundle_precheck_test.go
```diff
@@ -125,11 +126,99 @@ func Test_GetBundleState_FutureBundle_ReturnsTemporaryBlocked(t *testing.T) {
...
+		Build()
+
+	_, _, err := bundle.ValidateEnvelope(signer, envelope)
+	require.NoError(t, err)
+
+	state := GetBundleState(chainState, nil, envelope)
+	require.Equal(t, state, makeTemporaryBlockedState("bundle targets future time period"))
+}
+
+func Test_GetBundleState_HasBundleHasAlreadyBeenProcessed_ReturnsPermanentlyBlocked(t *testing.T) {
```
LorenzSchueler:
```suggestion
func Test_GetBundleState_BundleHasAlreadyBeenProcessed_ReturnsPermanentlyBlocked(t *testing.T) {
```

LorenzSchueler (dismissed):
Looks good.

LuisPH3 (approved):

LuisPH3 (changes requested):
one sec

api/sonicapi/execution_plan.go
```diff
@@ -112,8 +112,9 @@ func ToBundleExecutionPlan(rpcPlan RPCExecutionPlanComposable) (bundle.Execution
 	}
 
 	return bundle.ExecutionPlan{
-		Range: rpcPlan.BlockRange.toBundleBlockRange(),
-		Root:  root,
+		Range:  rpcPlan.BlockRange.toBundleBlockRange(),
+		Root:   root,
+		Period: bundle.MakeUnrestrictedTimePeriod(),
 	}, nil
```
LuisPH3, line 118: Unfortunately this is too late to modify the execution plan. 
The reason is that the execution plan has been returned from prepare bundle to the user to confirm the execution plan, users can then hash the plan and check that the transactions they are submitting match the hash.

This function is used during submit bundle, where txs are already signed with a plan without duration.
Although the timestamp may be serialized as nil, aggregating a new field to the rlp will change the hash. 

This error could have been identified if https://github.com/0xsoniclabs/sonic/pull/916 had been merged.  (the plan hash of prepare must match plan hash of submit)

LuisPH3 (changes requested):
The new field breaks the RPC hash consistency.

LuisPH3 (approved):
I rebased and reviewed the previous comment, the current hashing is alright.

api/sonicapi/execution_plan.go
LuisPH3: This is not a problem.
The reason is that this function injecting the duration into the pan is both used in submit and prepare, yielding the same plan hash.
The one problem introduced with this Pr is that users cannot reproduce the same hash without adding the period themselves. This can be solved in a follow-up PR by adding support into the json types.

PR #939, opened 2026-05-04
Process sponsored transactions as regular transactions during replay

LorenzSchueler:
Now that the state processor is replay-aware, this information can be used to process sponsored transactions as normal transactions during replay, so that the processing of the sponsored transactions does not introduce a second payment transaction. This then allows to not modify the rules during replay.


LuisPH3 (approved):

LuisPH3 (approved):

HerbertJordan (approved):
Nice improvement, thanks!

PR #940, opened 2026-05-04
Report execution cost from state processor

LorenzSchueler:
This PR implements reporting of the execution cost of transactions. This cost includes the cost for bundle steps which are rolled back. In an upcoming PR, this cost is used to implement an efficiency filter for bundles.

### DIFF
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -497,8 +497,7 @@ func (r *transactionRunner) runTransactionBundleInternal(
 		legacyTxOffset: legacyTxOffset,
 		trueTxOffset:   trueTxOffset,
 	}
-	success := bundle.RunBundle(txBundle, &runner)
-	if !success {
+	if !bundle.RunBundle(txBundle, &runner) {
 		// Mark the execution plan as processed in the StateDB to prevent processing
 		// another bundle with the same execution plan in the same block. Also keep
 		// track of the position of the bundle in the block.
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -79,9 +79,13 @@ func NewStateProcessorForReplay(
 }
 
 // ProcessSummary contains the result of processing a list of transactions,
-// including the list of processed transactions with their receipts.
+// including the list of processed transactions with their receipts and the
+// total execution cost. The execution cost tracks the gas consumed during
+// processing, including gas from rolled-back bundles, which distinguishes it
+// from the usedGas counter that gets reverted on snapshot rollback.
 type ProcessSummary struct {
 	ProcessedTransactions []ProcessedTransaction
+	ExecutionCost         core_types.ExecutionCost
 }
 
 // ProcessedTransaction represents a transaction that was considered for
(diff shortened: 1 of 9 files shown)


gossip/blockproc/bundle/bundle_processor.go
LuisPH3, line 28:
```
//  and the
// total execution cost. The execution cost tracks the gas consumed during
// processing, including gas from rolled-back bundles, which distinguishes it
// from the usedGas counter that gets reverted on snapshot rollback
```

LuisPH3 (dismissed):

LuisPH3 (dismissed):

LuisPH3 (dismissed):

gossip/blockproc/bundle/bundle_processor.go
HerbertJordan, line 1: I think that the total execution cost is not something the bundle processor should have to deal with. In essence, it is just forwarding information from one place to the other, not using it for anything.

I would suggest to revert this change and simplify the code by summing up the total execution cost in the implementation of the `TransactionRunner` interface -- in particular the `bundleTransactionRunner` defined in the state processor.

This should reduce the impact of this change significantly. Wdyt?

HerbertJordan (changes requested):

evmcore/state_processor.go
```diff
@@ -267,22 +276,32 @@ func runTransaction(
...
 	if !context.forReplay && context.upgrades.GasSubsidies && subsidies.IsSponsorshipRequest(tx) {
 		res, result := context.runner.runSponsoredTransaction(context, tx, legacyTxIndexOffset, trueTxIndexOffset)
-		return res, result
+		execCost := core_types.ExecutionCost(0)
+		for _, r := range res {
+			if r.Receipt != nil {
+				execCost += core_types.ExecutionCost(r.Receipt.GasUsed)
+			}
+		}
+		return res, result, execCost
```
HerbertJordan, line 297: To be aligned with the other "complex" runner implementation, the bundle, the computation of the `execCost` should be moved inside the `runSponsoredTransaction`. Maybe even for the `runRegularTransaction` case. Wdyt?

evmcore/state_processor_test.go
```diff
@@ -1139,7 +1141,7 @@ func TestRunTransaction_GasSubsidiesEnabled_RunsRegularTransactionWithoutSponsor
 		upgrades: opera.Upgrades{GasSubsidies: true},
 	}
 	runner.EXPECT().runRegularTransaction(context, tx, 0, 123).Return(processed, core_types.TransactionResultSuccessful)
-	got, status := runTransaction(context, tx, 0, 123)
+	got, status, _ := runTransaction(context, tx, 0, 123)
```
HerbertJordan: Please check in those error cases the reported `execGas`. It is just one line extra in each of those cases, and the "return 0" is covered.

HerbertJordan (dismissed):

evmcore/state_processor.go
HerbertJordan: Feel free to ignore this if this makes the changes more difficult to test.

LorenzSchueler: I agree that it would be more consistent, but on the other hand it also means we have significantly more return locations dealing with the exec cost, and we need to adjust all uses of the mocks. I'd prefer to keep it as is.

HerbertJordan (approved):

PR #941, opened 2026-05-04
Add bundle efficiency filter

LorenzSchueler:
This PR implements a bundle efficiency filter, to reject bundles which consume a lot of gas in the state processor, but only little gas in the transactions that make it into the block. This is needed as a protection against DoS attacks, because the gas pool is captured in the snapshots of the state processor, and therefore rolled back parts of a bundle are not captured by it.

Depends on https://github.com/0xsoniclabs/sonic/pull/940

### DIFF
--- a/evmcore/bundle_precheck.go
+++ b/evmcore/bundle_precheck.go
@@ -31,6 +31,11 @@ import (
 
 //go:generate mockgen -source=bundle_precheck.go -destination=bundle_precheck_mock.go -package=evmcore
 
+// MinBundleEfficiency is the minimum efficiency threshold for accepting
+// bundles. The efficiency of a bundle is defined as:
+// sum of usedGas of receipts / total execution cost.
+const MinBundleEfficiency = 0.2
+
 // BundleState represents the current evaluation state of a transaction bundle.
 // It indicates whether the bundle is executable as is, may become executable
 // at a later point or will never be executable. If the bundle is not executable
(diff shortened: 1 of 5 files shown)


evmcore/bundle_precheck.go
```diff
@@ -31,6 +31,11 @@ import (
 
 //go:generate mockgen -source=bundle_precheck.go -destination=bundle_precheck_mock.go -package=evmcore
 
+// MinBundleEfficiency is the minimum efficiency threshold for accepting
+// bundles. The efficiency of a bundle is defined as:
+// sum of usedGas of receipts / total execution cost.
+const MinBundleEfficiency = 0.2
```
HerbertJordan, line 37: Heads-up: this is good for this PR, but eventually this should become a client configuration parameter.

evmcore/bundle_precheck_test.go
```diff
@@ -762,6 +762,79 @@ func Test_trialRunBundle_DoesRunTransactionsThroughEVMAndReturnsIfTransactionsGo
...
+		},
+		"above threshold with multiple receipts": {
+			processedTxs: []ProcessedTransaction{
+				{Receipt: &types.Receipt{GasUsed: 10}},
+				{Receipt: &types.Receipt{GasUsed: 11}},
+			},
+			execCost:     core_types.ExecutionCost(100),
+			expectAccept: true,
+		},
+	}
```
HerbertJordan, line 807:
```suggestion
       "rejects empty bundles": {
			processedTxs: []ProcessedTransaction{},
			execCost:     core_types.ExecutionCost(0),
			expectAccept: false,
		},
	}
```

tests/bundles/bundle_efficiency_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+			if tc.ExpectBelowThreshold {
+				require.Less(t, float64(usedGas)/float64(execGas), evmcore.MinBundleEfficiency)
+				_, err = net.Send(envelope)
+				require.ErrorContains(t, err, "bundle trial-run failed")
+			} else {
+				require.GreaterOrEqual(t, float64(usedGas)/float64(execGas), evmcore.MinBundleEfficiency)
+				_, err = net.Send(envelope)
+				require.NoError(t, err)
+			}
```
HerbertJordan, line 116: Please wait for the bundle to be processed here and check that the valid Tx has a receipt to check that it is also correctly handled in the emitter and state processor (where it should not matter),

HerbertJordan (dismissed):
Looks good, thanks for adding it!

Another place where this efficiency limit should be enforced by a different PR is [here](https://github.com/0xsoniclabs/sonic/blob/ea7310eeaa579b7aa4c84116a78fa1c8d2d6291b/gossip/emitter/scheduler/processor.go#L95-L112).

HerbertJordan (approved):
Nice, thanks!

LuisPH3 (approved):

PR #945, opened 2026-05-05
Simplify dry runner and fix missclassification

HerbertJordan:
This PR simplifies the implementation of the dry-runner by performing two dry-run iterations instead of one:
- the first checks whether the current nonces would be fine for the bundle
- the second checks whether any future nonces would be fine for the bundle

This simplifies the overall code structure, removes complexity, and fixes https://github.com/0xsoniclabs/sonic-admin/issues/742

### DIFF
--- a/evmcore/bundle_precheck.go
+++ b/evmcore/bundle_precheck.go
@@ -183,79 +183,32 @@ func checkForNonceConflicts(
 	signer types.Signer,
 	nonceSource NonceSource,
 ) BundleState {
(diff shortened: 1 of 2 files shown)


LuisPH3 (approved):
This PR solves the issue of dry-running bundles not detecting gapped nonces as temporarily nonexecutable.

The solution proposed adds a flag which tolerates first found nonce for each sender not being the next expected nonce. Running in this mode allows gapped nonces, running without the option does not, visit twice the bundle with each configuration differentiates executable from temporarily nonexecutable.

PR #946, opened 2026-05-05
DoEstimateGas reports cancelled error reason

facuMH:
When `DoCall` is called with `timeout` parameter set to `0`, it uses a context without timeout. In the case of a cancelled context it assumed it was a timeout and printed the timeout parameter, instead of the error. This changes make it so that if the timeout parameters was not zero, then it is printed, otherwise the reason for the context cancellation is printed. 
This change allows for better error reporting in a flaky test.

### DIFF
--- a/api/ethapi/api.go
+++ b/api/ethapi/api.go
@@ -1204,8 +1204,10 @@ func DoCall(
 	}
 
 	// If the timer caused an abort, return an appropriate error message
-	if evm.Cancelled() {
+	if evm.Cancelled() && timeout > 0 {
 		return nil, fmt.Errorf("execution aborted (timeout = %v)", timeout)
+	} else if evm.Cancelled() {
+		return nil, fmt.Errorf("execution aborted, (reason: %v)", ctx.Err())
 	}
 	if err != nil {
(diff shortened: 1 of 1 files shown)


api/ethapi/api.go
```diff
@@ -1205,8 +1204,10 @@ func DoCall(
 	}
 
 	// If the timer caused an abort, return an appropriate error message
-	if evm.Cancelled() {
+	if evm.Cancelled() && timeout > 0 {
 		return nil, fmt.Errorf("execution aborted (timeout = %v)", timeout)
+	} else if evm.Cancelled() {
+		return nil, fmt.Errorf("execution aborted, (reason: %v)", ctx.Err())
 	}
```
jenikd: Do wee need to check timeout? It would probably be in the ctx.Err()
```suggestion
	if evm.Cancelled() {
		return nil, fmt.Errorf("execution aborted, (reason: %v)", ctx.Err())
     }
```

facuMH: this way only tells "there was a context timeout" not how long the wait was.

facuMH: changed to print both.

LuisPH3:
the underlaying issue is solved now. this pr would not have provided any extra information about it.

PR #949, opened 2026-05-06
Add bundle efficiency filter for single proposer mode

LorenzSchueler:
This PR adds the bundle efficiency filter to the emitter also for single proposer mode.

### DIFF
--- a/gossip/emitter/scheduler/processor.go
+++ b/gossip/emitter/scheduler/processor.go
@@ -84,9 +84,7 @@ type evmProcessor struct {
 	stateDb   state.StateDB
 }
 
-func (p *evmProcessor) run(tx *types.Transaction) (
-	result bool, gasUsed uint64,
-) {
+func (p *evmProcessor) run(tx *types.Transaction) (bool, uint64) {
 	snapshot := p.stateDb.InterTxSnapshot()
 
 	// Note: the index can be set to 0 since code running inside the EVM can not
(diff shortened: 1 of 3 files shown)


gossip/emitter/scheduler/processor.go
```diff
@@ -87,30 +87,27 @@ type evmProcessor struct {
...
 	}
-	return hasAcceptedTransaction, gasUsed
+
+	if summary.ExecutionCost == 0 ||
+		float64(gasUsedByTx)/float64(summary.ExecutionCost) < evmcore.MinBundleEfficiency {
+		p.stateDb.RevertToInterTxSnapshot(snapshot)
+		return false, gasUsed
+	}
+
+	return true, gasUsed + gasUsedByTx
```
HerbertJordan: `gasUsed` is always 0 in this function. It is also a bit confusing that there are two `gasUsed` variables in this code. Please merge them into one by either removing the named return value or using the return value name consistently.

gossip/emitter/scheduler/processor_test.go
```diff
@@ -81,60 +88,130 @@ func TestEvmProcessor_Run_IfExecutionProducesMultipleProcessedTransactions_SumsU
...
-		runner := NewMockevmProcessorRunner(ctrl)
-		txA := &types.Transaction{}
-		txB := &types.Transaction{}
-		runner.EXPECT().Run(0, gomock.Any()).Return(evmcore.ProcessSummary{
-			ProcessedTransactions: []evmcore.ProcessedTransaction{
-				{Transaction: txA, Receipt: nil},
-				{Transaction: txB, Receipt: nil},
+			expectAccept: false,
+		},
+		"below threshold with empty bundle": {
```
HerbertJordan:
```suggestion
		"below threshold with no processed transaction": {
```

gossip/emitter/scheduler/processor_test.go
```diff
@@ -81,60 +88,130 @@ func TestEvmProcessor_Run_IfExecutionProducesMultipleProcessedTransactions_SumsU
...
+				},
+				ExecutionCost: core_types.ExecutionCost(100),
+			},
+			expectAccept: false,
+		},
+		"below threshold with single receipt": {
+			tx: tx,
+			summary: evmcore.ProcessSummary{
+				ProcessedTransactions: []evmcore.ProcessedTransaction{
+					{Transaction: tx, Receipt: &types.Receipt{GasUsed: 19}},
```
HerbertJordan: Could you derive those values from the `evmcore.MinBundleEfficiency` such that if the constant is changed or turned into a parameter this tests do not need to be updated?

HerbertJordan (changes requested):
Looks good, a few minor cleanup requests though. Thanks.

gossip/emitter/scheduler/processor.go
LorenzSchueler: Sorry, I got confused by the named return values. I thought that we got the previous gasUsed passed in and had to return the new one.

HerbertJordan (approved):
Looks good, thanks!

facuMH (approved):

PR #951, opened 2026-05-06
Restrict data returned by OnNewLog call

HerbertJordan:
This PR eliminates the in-flight adjustment of the transaction index of log messages reported by the state processor [here](https://github.com/0xsoniclabs/sonic/blob/c847f4bb9ba7c3d1c9ae38cf89e8b5bdbe5cd841/gossip/blockproc/evmmodule/evm.go#L158-L162).

It turns out that no consumer of the log message requires the transaction index to be set. To make sure this remains like this, a new internal log type (`core_types.Log`) is introduced not covering this field.

The rest of the changes are mere adjustments to the type switch.

### DIFF
--- a/evmcore/core_types/log.go
+++ b/evmcore/core_types/log.go
@@ -1,0 +1,47 @@
+// Copyright 2026 Sonic Operations Ltd
+// This file is part of the Sonic Client
+//
+// Sonic is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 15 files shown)


LuisPH3 (approved):
This PR introduces a sonic type for logs, which does not include transaction index. The existing transaction index used at the time of emitting a log may be off and should not be used.

LorenzSchueler (approved):
I already wondered in bertha when playing around with the onNewLog callback, why the other optional fields were not filled in. This PR cleans this up 👍🏽

PR #952, opened 2026-05-06
Add sponsored transaction and bundle metrics to c_block_callbacks

simonlechner:
This PR adds metrics for bundles and sponsored transactions as well as metrics for the effective gas used in a block and in a bundle.


gossip/c_block_callbacks.go
```diff
@@ -614,12 +621,33 @@ func processUserTransactionsNoLimits(
 	blockBuilder *inter.BlockBuilder,
 	orderedTxs []*types.Transaction,
 	userTransactionGasLimit uint64) {
-	for _, processed := range evmProcessor.Execute(orderedTxs, userTransactionGasLimit).ProcessedTransactions {
+	sumGasUsed := uint64(0)
+	for _, tx := range orderedTxs {
+		if subsidies.IsSponsorshipRequest(tx) {
+			sponsoredCounter.Inc(1)
```
HerbertJordan: I think this should not be counted on the input list but on the list of accepted transactions.

gossip/c_block_callbacks.go
```diff
@@ -668,12 +707,31 @@ func processUserTransactions(
...
 					remainingSize -= processed.Transaction.Size()
 					causedBy[processed.Transaction.Hash()] = tx.Hash()
+					receiptGasSum += processed.Receipt.GasUsed
+				}
+
+				// The envelope is only part of the processed transactions if the bundle failed
+				if bundle.IsEnvelope(processed.Transaction) {
+					skippedBundleCounter.Inc(1)
+				}
+				if subsidies.IsSponsorshipRequest(processed.Transaction) && processed.Receipt == nil {
```
HerbertJordan: Unfortunately, an envelope could also be identified as a Sponsorship request, since it may have a gas price of zero. So if it is an envelope, it should not be counted as a sponsorship request again.

gossip/c_block_callbacks.go
```diff
@@ -668,12 +707,31 @@ func processUserTransactions(
 					remainingGas -= processed.Receipt.GasUsed
 					remainingSize -= processed.Transaction.Size()
 					causedBy[processed.Transaction.Hash()] = tx.Hash()
+					receiptGasSum += processed.Receipt.GasUsed
+				}
+
+				// The envelope is only part of the processed transactions if the bundle failed
+				if bundle.IsEnvelope(processed.Transaction) {
+					skippedBundleCounter.Inc(1)
```
HerbertJordan: This would count every bundle as skipped, isn't it?

gossip/c_block_callbacks.go
```diff
@@ -668,12 +707,31 @@ func processUserTransactions(
...
 	}
+
+	if summaryGasSum == 0 && receiptGasSum == 0 {
+		effectiveGasGauge.Update(100)
+	} else if summaryGasSum == 0 {
+		effectiveGasGauge.Update(0)
+	} else {
+		effective := float64(receiptGasSum) / float64(summaryGasSum)
+		effectiveGasGauge.Update(int64(effective * 100))
+	}
```
HerbertJordan, line 732: Alright, this gives us the full block-level efficiency of the effective gas, which is also an interesting metric.  However, what I think we should also monitor is the effectiveness of individual bundles. Thus, one sample point for each bundle instead of one sample per block. This can be an additional metric.

gossip/c_block_callbacks.go
```diff
@@ -76,6 +77,12 @@ var (
 
 	confirmedEventsMeter = metrics.GetOrRegisterMeter("chain/events/confirmed", nil) // events received from lachesis
 	spilledEventsMeter   = metrics.GetOrRegisterMeter("chain/events/spilled", nil)   // tx excluded because of MaxBlockGas
+
+	sponsoredCounter        = utils.MetricsCounterWrapper(metrics.GetOrRegisterCounter("chain/sponsored", nil))
+	skippedSponsoredCounter = utils.MetricsCounterWrapper(metrics.GetOrRegisterCounter("chain/sponsored/skipped", nil))
+	bundleCounter           = utils.MetricsCounterWrapper(metrics.GetOrRegisterCounter("chain/bundles", nil))
+	skippedBundleCounter    = utils.MetricsCounterWrapper(metrics.GetOrRegisterCounter("chain/bundles/skipped", nil))
+	effectiveGasGauge       = utils.MetricsGaugeWrapper(metrics.GetOrRegisterGauge("chain/gas/effective", nil))
```
HerbertJordan: Why a Gauge and not a histogram? Gauges are note very useful here as it only recalls the last recorded value, right?

gossip/c_block_callbacks.go
HerbertJordan, line 621: The code for the metrics is duplicated between this function and `processUserTransactions`. Maybe you can factor this out somehow?

HerbertJordan (changes requested):

gossip/c_block_callbacks.go
simonlechner: I went through the code of the `bundle_processor` and talked to @LorenzSchueler. As far as I understood, the envelope is only added to the processed transaction in case of a failed/skipped bundle.

gossip/c_block_callbacks.go
simonlechner: I thought Gauges keep historical data in a time series format. Changed to histogram.

evmcore/state_processor.go
```diff
@@ -524,6 +532,13 @@ func (r *transactionRunner) runTransactionBundleInternal(
 	// copies of themselves without finding a hash-function collision.
 	ctxt.statedb.AddProcessedBundle(planHash, positionInBlock)
 
+	bundleMaxGasUsed := uint64(0)
+	for _, bundleTx := range txBundle.Transactions {
+		bundleMaxGasUsed += bundleTx.Gas()
+	}
+	effective := float64(bundleMaxGasUsed) / float64(runner.executionCost) * 100
+	effectiveGasHistogram.Update(int64(effective))
```
HerbertJordan, line 540: That is not a good place for this metric. The reason is that this code is reused a lot:
- in the actual head state processing
- in the bundle prechecker
- in RPC calls test-running bundles
The resulting metric would mix up all those calls, and we would get hardly comparable data from the nodes.

To make results interpretable for diagnoses, we need those metrics separated.

gossip/c_block_callbacks.go
```diff
@@ -668,15 +689,49 @@ func processUserTransactions(
...
 		}
 	}
+
+	updateMetricsEffectiveGas(uint64(summaryGasSum), receiptGasSum)
+
 	return skippedCounter, causedBy
 }
 
+func updateMetricsInputTx(tx *types.Transaction) {
+	if bundle.IsEnvelope(tx) {
```
HerbertJordan, line 708: here, something like this is missing:

```suggestion
	if rules.Brio && bundle.IsEnvelope(tx) {
```

gossip/c_block_callbacks.go
```diff
@@ -668,15 +689,49 @@ func processUserTransactions(
...
+		bundleCounter.Inc(1)
+	} else if subsidies.IsSponsorshipRequest(tx) {
+		sponsoredCounter.Inc(1)
+	}
+}
+
+func updateMetricsSkippedTx(tx *types.Transaction, receipt *types.Receipt) {
+	// The envelope is only part of the processed transactions if the bundle failed
+	if bundle.IsEnvelope(tx) {
```
HerbertJordan, line 717:
```suggestion
	if rules.Brio && bundle.IsEnvelope(tx) {
```

gossip/c_block_callbacks.go
```diff
@@ -668,15 +689,49 @@ func processUserTransactions(
...
+		skippedSponsoredCounter.Inc(1)
+	}
+}
+
+func updateMetricsEffectiveGas(summaryGasSum, receiptGasSum uint64) {
+	if summaryGasSum == 0 && receiptGasSum == 0 {
+		effectiveGasHistogram.Update(100)
+	} else if summaryGasSum == 0 {
+		effectiveGasHistogram.Update(0)
+	} else {
```
HerbertJordan, line 729: simpler?

```suggestion
	if summaryGasSum == 0 {
		effectiveGasHistogram.Update(100)
	} else {
```

gossip/c_block_callbacks.go
```diff
@@ -668,15 +689,49 @@ func processUserTransactions(
...
+		effective := float64(receiptGasSum) / float64(summaryGasSum)
+		effectiveGasHistogram.Update(int64(effective * 100))
```
HerbertJordan, line 731: I would keep the effectiveness in the 0-1 range and not multiply with 100 here. This is a presentation detail that can and should be handled in grafana.

HerbertJordan (changes requested):

evmcore/state_processor.go
HerbertJordan: There a two ways to separate those metrics:
 - metrics can have [labels](https://prometheus.io/docs/concepts/data_model/); you can tag each recording whether it is made while processing a head state, a precheck, or a RPC call; however, (a) this would need you to get this information here, and (b) I don't think the go-ethereum infra is good in supporting labels
 - have different metrics for different contexts

simonlechner:
Closed in favor of #1003.

PR #953, opened 2026-05-06
Remove tracking of legacy tx-offset

### LINKED ISSUE #856: trace_block returns wrong transactionPosition due to receipt LRU cache contamination
## Description

`trace_block` returns `transactionPosition` values that are higher than `eth_getBlockByNumber`'s `transactionIndex` for the same transactions in certain blocks. The bug is intermittent and transient — the same block can show the mismatch on one node but not another, and previously affected blocks can later return correct data.

## Reproduction

Block `0x408b08f` (67678351) on mainnet. As of April 16, 2026, this block shows the mismatch on `rpc.soniclabs.com` but not on all nodes — confirming the transient nature.

```bash
# eth_getBlockByNumber — returns transactionIndex 0x0 and 0x1
curl -s -X POST https://rpc.soniclabs.com -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0x408b08f",true],"id":1}' \
  | jq '.result.transactions[] | {hash: .hash, transactionIndex}'

# trace_block — returns transactionPosition 1 and 2 (should be 0 and 1)
curl -s -X POST https://rpc.soniclabs.com -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","method":"trace_block","params":["0x408b08f"],"id":1}' \
  | jq '[.result[] | select(.traceAddress == []) | {hash: .transactionHash, transactionPosition}]'
```

Both endpoints confirmed running `Sonic/v2.1.6-26dd110c-1773221015`.

## Root cause

There are two compounding inflation sources, plus a cache path that makes the bug transient.

### Source 1: `statedb.TxIndex()` inflation within a single Execute batch

In `evmcore/state_processor.go`, `_runTransaction` calls `statedb.SetTxContext(tx.Hash(), txIndex)` where `txIndex` is the raw slot position including skipped (nil-receipt) transactions. The receipt then gets `receipt.TransactionIndex = uint(statedb.TxIndex())`. When a sponsored/bundle transaction fails (not covered, insufficient gas, invalid), it produces no receipt but still consumes a slot. The next valid transaction gets a `TransactionIndex` that counts the skipped slot.

There is an existing TODO comment in `state_processor_test.go` (`// TODO: this should be 1; this needs to be investigated`) acknowledging this.

### Source 2: `txsOffset` inflation across Execute batches

In `gossip/blockproc/evmmodule/evm.go:146`:
```go
txsOffset := uint(len(p.processedTxs))
```
This counts ALL previously processed transactions including nil-receipt ones. At lines 158-163, subsequent receipts get `p.Receipt.TransactionIndex += txsOffset`, compounding the inflation.

`Finalize()` (lines 171-181) then strips nil-receipt entries from `evmBlock.Transactions`, but the receipts already carry the inflated index.

### The cache contamination path (explains transient behavior)

This is the key mechanism that makes the bug appear and disappear:

1. **`SetReceipts`** (`gossip/evmstore/store_receipts.go:34-43`) persists receipts to disk AND caches the in-memory `types.Receipts` objects (with inflated `TransactionIndex`) in the LRU cache.

2. **`GetReceipts`** (lines 95-100) checks the LRU cache first. On **cache hit**, it returns the inflated values directly — no `DeriveFields` call.

3. On **cache miss** (lines 103-111), it reads from disk and calls `UnwrapStorageReceipts` → `DeriveFields`, which unconditionally overwrites `TransactionIndex` with `uint(i)` (the correct slice position). `TransactionIndex` is not part of `ReceiptForStorage` RLP encoding, so disk reads always produce correct indices.

4. **`replayBlock`** (`api/ethapi/tx_trace.go:399`) uses `receipts[i].TransactionIndex` as the `index` parameter that flows through to `transactionPosition`:
   ```go
   txTraces, err := s.traceTx(ctx, msg, state, block, tx, uint64(receipts[i].TransactionIndex), receipts[i].Status)
   ```

5. Meanwhile, `eth_getBlockByNumber` uses the literal loop index from iterating `evmBlock.Transactions` and never touches receipts — so it's always correct.

**Result:** cache hot from live block processing → wrong `transactionPosition`. Cache evicted (node restart, LRU pressure) → re-derived from disk → correct. Nodes that sync via LLR block-record ingestion (`WriteFullBlockRecord` → `SetRawReceipts`, which evicts cache without populating it with in-memory objects) are never affected.

## Relationship to PR #842

[PR #842](https://github.com/0xsoniclabs/sonic/pull/842) introduces a `trueTxIndexOffset` counter to address the cross-batch offset inflation (Source 2). However:

- It does not address Source 1 (`statedb.TxIndex()` within a single batch)
- It does not address the cache contamination path — even with correct indices at processing time, the fundamental issue of `SetReceipts` caching in-memory objects that `GetReceipts` returns without re-deriving would persist for any future index bugs

A more complete fix could also have `GetReceipts` call `DeriveFields` on cache-hit results, or have `SetReceipts` derive fields before caching.

## Environment

- Sonic v2.1.6 (`26dd110c`)
- Mainnet (chain ID 146)
- Affects `trace_block`, `trace_transaction`, and likely `trace_replayBlockTransactions`
- `debug_traceBlockByNumber` is NOT affected (uses a different code path)
- `eth_getBlockByNumber` is NOT affected

HerbertJordan:
This PR removes the faulty legacy transaction index tracking from the state processor and switches to use the true-tx-indexing in all contexts.

From studying the code, the transaction index is not reaching any persistent state or participating any operation relevant for consensus. However, as issue #856 shows, the faulty offset is accessible through RPC calls, and should thus be fixed.

The [TransactionIndex](https://github.com/0xsoniclabs/go-ethereum/blob/e9dfccd41dbeaafc8489794ca8e01906929f24ab/core/types/receipt.go#L74) is not stored in the gossip store: Receipts are recorded in the gossip store by a call to `SetReceipts` [here](https://github.com/0xsoniclabs/sonic/blob/837431721af5c85242827b8930527a8e29e13c8f/gossip/c_block_callbacks.go#L518), where receipts are [wrapped](https://github.com/0xsoniclabs/sonic/blob/837431721af5c85242827b8930527a8e29e13c8f/gossip/evmstore/store_receipts.go#L35) into the `types.ReceiptForStorage` type for which RLP encoding is overwritten [here](https://github.com/0xsoniclabs/go-ethereum/blob/e9dfccd41dbeaafc8489794ca8e01906929f24ab/core/types/receipt.go#L321-L353) to only include the  status, the cumulative gas used, and the logs. Thus, the transaction index does not end up on the disk.

The [TransactionIndex](https://github.com/0xsoniclabs/go-ethereum/blob/e9dfccd41dbeaafc8489794ca8e01906929f24ab/core/types/receipt.go#L74) is also not consensus relevant: As documented in the Receipt type [definition](https://github.com/0xsoniclabs/go-ethereum/blob/e9dfccd41dbeaafc8489794ca8e01906929f24ab/core/types/receipt.go#L53-L74), the TransactionIndex is not part of the consensus fields. This is confirmed by the Receipt's [EncodeRLP](https://github.com/0xsoniclabs/go-ethereum/blob/e9dfccd41dbeaafc8489794ca8e01906929f24ab/core/types/receipt.go#L123-L135) function, which does not cover the Transaction index.

A code search reveals only three explicit references to the `TransactionIndex` field in the Sonic code:
- in the state processor [here](https://github.com/0xsoniclabs/sonic/blob/837431721af5c85242827b8930527a8e29e13c8f/evmcore/state_processor.go#L784) and [here](https://github.com/0xsoniclabs/sonic/blob/837431721af5c85242827b8930527a8e29e13c8f/evmcore/state_processor.go#L895): these are the locations where the transaction index is set, and where this PR is switching over the legacy indexing to the true indexing
- in the transaction trace API [here](https://github.com/0xsoniclabs/sonic/blob/837431721af5c85242827b8930527a8e29e13c8f/api/ethapi/tx_trace.go#L399): the field is used for setting up the tracing of transactions; receipts are fetched from the store using [GetReceipts](https://github.com/0xsoniclabs/sonic/blob/837431721af5c85242827b8930527a8e29e13c8f/gossip/evmstore/store_receipts.go#L95) which either serves those receipts from the cache (which used to contain the faulty transaction indices before this PR) or [restores](https://github.com/0xsoniclabs/sonic/blob/837431721af5c85242827b8930527a8e29e13c8f/gossip/evmstore/store_receipts.go#L105) them from the disk and derives the transaction index from meta information [here](https://github.com/0xsoniclabs/sonic/blob/837431721af5c85242827b8930527a8e29e13c8f/gossip/evmstore/store_receipts.go#L90).
- in the marshaling and unmarshaling of Receipts from and to JSON ([here](https://github.com/0xsoniclabs/go-ethereum/blob/e9dfccd41dbeaafc8489794ca8e01906929f24ab/core/types/gen_receipt_json.go#L50) and [here](https://github.com/0xsoniclabs/go-ethereum/blob/e9dfccd41dbeaafc8489794ca8e01906929f24ab/core/types/gen_receipt_json.go#L125))

None of those cases are critical for backward compatibility or consensus purposes.

Furthermore, if, contrary to the analysis presented above, the index is consensus relevant it is expected to be detected during release testing when comparing different client versions processing skipped transactions.

Should fix #856

***For reviewers:***
To keep the modifications of this PR limited, this PR only removes the legacy transaction offsets. It deliberately does not rename the "trueTxOffest" to "txOffset" to avoid confusion. This cleanup should be done in a follow-up PR.

### DIFF
--- a/gossip/blockproc/evmmodule/evm_test.go
+++ b/gossip/blockproc/evmmodule/evm_test.go
@@ -274,8 +274,7 @@ func TestOperaEVMProcessor_Execute_StateProcessorProducesTransactionsAndBundles_
 	require.Equal(summary, result)
 }
 
-/*
-func TestOperaEVMProcessor_Execute_UsesLengthOfProcessedTransactionsAsTransactionIndexOffsetInReceipts(t *testing.T) {
+func TestOperaEVMProcessor_Execute_UsesNumberOfAcceptedTransactionsAsTransactionIndexOffsetInProcessorCall(t *testing.T) {
 	tests := map[string][]evmcore.ProcessedTransaction{
 		"nil":   nil,
 		"empty": {},
@@ -302,12 +301,19 @@ func TestOperaEVMProcessor_Execute_UsesLengthOfProcessedTransactionsAsTransactio
 			any := gomock.Any()
 			factory.EXPECT().NewStateProcessorForHeadState(any, any, any).Return(stateProcessor)
 
+			numPreAccepted := 0
+			for _, cur := range processedTransactions {
+				if cur.Receipt != nil {
+					numPreAccepted++
+				}
+			}
+
 			stateProcessor.EXPECT().
-				Process(any, any, any, any, any, any, any).
+				Process(any, any, any, any, any, numPreAccepted, any).
 				Return(evmcore.ProcessSummary{
 					ProcessedTransactions: []evmcore.ProcessedTransaction{
-						{Receipt: &types.Receipt{TransactionIndex: 0}},
-						{Receipt: &types.Receipt{TransactionIndex: 1}},
+						{Receipt: &types.Receipt{TransactionIndex: uint(numPreAccepted)}},
+						{Receipt: &types.Receipt{TransactionIndex: uint(numPreAccepted) + 1}},
 					},
 				})
 
@@ -321,13 +327,12 @@ func TestOperaEVMProcessor_Execute_UsesLengthOfProcessedTransactionsAsTransactio
 			require.Len(t, summary.ProcessedTransactions, 2)
 			for i, cur := range summary.ProcessedTransactions {
 				got := cur.Receipt.TransactionIndex
-				want := len(processedTransactions) + i
+				want := numPreAccepted + i
 				require.EqualValues(t, want, got)
 			}
 		})
 	}
 }
-*/
 
 func TestOperaEVMProcessor_Execute_UsesNumberOfTransactionsWithReceiptsAsTransactionOffsetInEvmProcessor(t *testing.T) {
 	tests := map[string][]evmcore.ProcessedTransaction{
--- a/gossip/blockproc/evmmodule/evm_test.go
+++ b/gossip/blockproc/evmmodule/evm_test.go
@@ -197,7 +197,7 @@ func TestOperaEVMProcessor_Execute_StateProcessorIntroducesTransactions_Produces
 	any := gomock.Any()
 	factory.EXPECT().NewStateProcessorForHeadState(any, any, any).Return(stateProcessor).AnyTimes()
 
-	summary := evmcore.ProcessSummary{
+	summary1 := evmcore.ProcessSummary{
 		ProcessedTransactions: []evmcore.ProcessedTransaction{
 			{Receipt: &types.Receipt{TransactionIndex: 0}},
 			{Receipt: &types.Receipt{TransactionIndex: 1}},
@@ -207,9 +207,21 @@ func TestOperaEVMProcessor_Execute_StateProcessorIntroducesTransactions_Produces
 		},
 	}
 
+	summary2 := evmcore.ProcessSummary{
+		ProcessedTransactions: []evmcore.ProcessedTransaction{
+			{Receipt: &types.Receipt{TransactionIndex: 5}},
+			{Receipt: &types.Receipt{TransactionIndex: 6}},
+			{Receipt: &types.Receipt{TransactionIndex: 7}},
+		},
+	}
+
 	stateProcessor.EXPECT().Process(
-		any, any, any, any, any, any, any,
-	).Return(summary).Times(2)
+		any, any, any, any, any, 0, any,
+	).Return(summary1)
+
+	stateProcessor.EXPECT().Process(
+		any, any, any, any, any, 5, any,
+	).Return(summary2)
 
 	processor := &OperaEVMProcessor{
 		processorFactory: factory,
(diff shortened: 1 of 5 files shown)


gossip/blockproc/evmmodule/evm_test.go
```diff
@@ -264,6 +274,7 @@ func TestOperaEVMProcessor_Execute_StateProcessorProducesTransactionsAndBundles_
 	require.Equal(summary, result)
 }
 
+/*
```
LuisPH3: is this intended?

LuisPH3 (dismissed):
The changes remove the concurrent counting of transaction indices in favor of the new infrastructure provided by `trueTxIndexOffset`.

- The new infrastructure is aware of bundle rollbacks [here](https://github.com/0xsoniclabs/sonic/blob/19e9245787bf44d24b91a44b26a5884dbef0d7bb/evmcore/state_processor.go#L546).
- the new counter is aware of sponsorships and bundle multiple txs executions [here](https://github.com/0xsoniclabs/sonic/blob/19e9245787bf44d24b91a44b26a5884dbef0d7bb/evmcore/state_processor.go#L236), 
- the new counter accounts for successful internal txs [here](https://github.com/0xsoniclabs/sonic/blob/19e9245787bf44d24b91a44b26a5884dbef0d7bb/gossip/blockproc/evmmodule/evm.go#L150), the older counter had no notion of failed internal txs. 

 The older counter would patch the receipt index, the new counter creates receipts with the correct count [here](https://github.com/0xsoniclabs/sonic/blob/19e9245787bf44d24b91a44b26a5884dbef0d7bb/evmcore/state_processor.go#L603) & [here](https://github.com/0xsoniclabs/sonic/blob/19e9245787bf44d24b91a44b26a5884dbef0d7bb/evmcore/state_processor.go#L895)

gossip/blockproc/evmmodule/evm_test.go
HerbertJordan: No, thanks for spotting. Forgot about this. Fixed in 6d233f0ec39f436136643d1ea96d99041f26c803

LuisPH3 (approved):

PR #954, opened 2026-05-06
Cache bundle.OpenEnvelope

LuisPH3:
This PR caches OpenBundle results into the transaction instance storage added [here](https://github.com/0xsoniclabs/go-ethereum/pull/15)

This is an impactful performance optimization as opening the bundle implies creating new transaction instances for the bundled transactions. These new instances do not have the sender cached, and the singer needs to be used to recompute every time a bundle is open, which happens a few times along the pipeline. 
 
 
An example of the impact of this operation:
**Before**: 
image

**After**: 
image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -150,6 +150,6 @@ require (
 	pgregory.net/rand v1.0.2 // indirect
 )
 
-replace github.com/ethereum/go-ethereum => github.com/0xsoniclabs/go-ethereum v0.0.0-20260506145648-6b3ddf5db680
+replace github.com/ethereum/go-ethereum => github.com/0xsoniclabs/go-ethereum v0.0.0-20260507120815-e9dfccd41dbe
 
 replace github.com/Fantom-foundation/lachesis-base => github.com/Fantom-foundation/lachesis-base-sonic v0.0.0-20260429065829-930ae462cd09
--- a/go.mod
+++ b/go.mod
@@ -150,6 +150,6 @@ require (
 	pgregory.net/rand v1.0.2 // indirect
 )
 
-replace github.com/ethereum/go-ethereum => github.com/0xsoniclabs/go-ethereum v0.0.0-20260424113012-971561538c58
+replace github.com/ethereum/go-ethereum => github.com/0xsoniclabs/go-ethereum v0.0.0-20260506145648-6b3ddf5db680
 
 replace github.com/Fantom-foundation/lachesis-base => github.com/Fantom-foundation/lachesis-base-sonic v0.0.0-20260429065829-930ae462cd09
--- a/gossip/blockproc/bundle/bundle.go
+++ b/gossip/blockproc/bundle/bundle.go
@@ -62,7 +62,7 @@ func OpenEnvelope(
 
 	const openedBundleKey = "openedBundle"
 	if txBundle, cached := types.GetSonicPayload[TransactionBundle](tx, openedBundleKey); cached {
-		return txBundle, nil
+		return txBundle.Clone(), nil
 	}
 
 	txBundle, err := decode(signer, tx.Data())
@@ -70,7 +70,7 @@ func OpenEnvelope(
 		return TransactionBundle{}, err
 	}
 	types.SetSonicPayload(tx, openedBundleKey, txBundle)
-	return txBundle, nil
+	return txBundle.Clone(), nil
 }
 
 var (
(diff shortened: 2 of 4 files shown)


gossip/blockproc/bundle/bundle.go
```diff
@@ -49,14 +49,28 @@ func IsEnvelope(tx *types.Transaction) bool {
...
 	tx *types.Transaction,
 ) (TransactionBundle, error) {
 	if !IsEnvelope(tx) {
 		return TransactionBundle{}, fmt.Errorf("not an envelope")
 	}
-	return decode(signer, tx.Data())
+
+	const openedBundleKey = "openedBundle"
+	if txBundle, cached := types.GetSonicPayload[TransactionBundle](tx, openedBundleKey); cached {
+		return txBundle, nil
```
HerbertJordan: Something that concerns me here is that the resulting objects are not immutable. Since all fields are public, users receiving those from this function can modify them.

The [ExecutionPlan](https://github.com/0xsoniclabs/sonic/blob/ed0040cd3de735836a44fe61b1a162cb0ce867da/gossip/blockproc/bundle/bundle.go#L93) should be fine, since it is a value, and it gets deep-copied implicitly. The deep-copy stops with the root [ExecutionStep](https://github.com/0xsoniclabs/sonic/blob/ed0040cd3de735836a44fe61b1a162cb0ce867da/gossip/blockproc/bundle/execution_plan.go#L103), but this one seems to be immutable.

The [Transactions](https://github.com/0xsoniclabs/sonic/blob/ed0040cd3de735836a44fe61b1a162cb0ce867da/gossip/blockproc/bundle/bundle.go#L92) however, are mutable.

How about adding a `Clone()` function to the `TransactionBundle` and return a clone from this function instead of the reference that is stored in the cache? 

`Clone` does not need to copy objects that are designed immutable (like, maybe, the `ExecutionStep`)

Wdyt?

gossip/blockproc/bundle/bundle_test.go
```diff
@@ -126,6 +127,45 @@ func TestOpenEnvelope_SuccessfullyDecodesEnvelopes(t *testing.T) {
...
+		Step(key, &types.AccessListTx{Nonce: 1}),
+	).BuildBundle()
+
+	payload, err := bundle.Encode()
+	require.NoError(t, err)
+
+	envelope := types.NewTx(&types.LegacyTx{
+		To:   &BundleProcessor,
+		Data: payload,
+	})
```
HerbertJordan: nit:

```suggestion
	envelope := NewBuilder().AllOf(
		Step(key, &types.AccessListTx{Nonce: 1}),
	).Build()
```

go.mod
```diff
@@ -150,6 +150,6 @@ require (
 	pgregory.net/rand v1.0.2 // indirect
 )
 
-replace github.com/ethereum/go-ethereum => github.com/0xsoniclabs/go-ethereum v0.0.0-20260424113012-971561538c58
+replace github.com/ethereum/go-ethereum => github.com/0xsoniclabs/go-ethereum v0.0.0-20260506145648-6b3ddf5db680
```
HerbertJordan: Please only merge once this version number points to the version officially included in the respective branch of Sonic's go-ethereum fork.

gossip/blockproc/bundle/bundle.go
LuisPH3: I saw this as well, but I am not as concerned about it:
`type.Transaction` are immutable objects (except for the cached values)
- getters construct new instances, therefore code like these panics:
```golang
	tx := types.NewTx(&types.SetCodeTx{})
	tx.AccessList()[0] = types.AccessTuple{ Address: common.Address{123} }
	fmt.Println(tx.AccessList()[0])
``` 
-  caches reflect the lazy behavior of evaluating transactions during their lifetime. In this case (for bundled txs) `hash`, `size` and `from` are already filled in, therefore it would be ok to clone it.

Nevertheless, we introduce a setter in this PR for the named map. If we wanted to use this in the future to cache new values. It makes sense that these are retrospectively applied to any instance of the transaction in the system, just like existing caches do. Transactions are unique entities and in some cases instances are exchanged to benefit from pre-cached values (like [here](https://github.com/0xsoniclabs/sonic/blob/ed0040cd3de735836a44fe61b1a162cb0ce867da/gossip/handler.go#L847)) 

The required change is not needed to support this feature, and the current code changes are in line with the current semantics of the `types.Transaction` type. I believe we should not clone them.

go.mod
LuisPH3: updated

gossip/blockproc/bundle/bundle.go
HerbertJordan: My concern was actually about the `Transactions` map, which is a `map[TxReference]*types.Transaction` that remains mutable and could be cloned easily.

With the transactions referenced in the map, I agree, no cloning needed.

LuisPH3: you are right

LuisPH3: https://github.com/0xsoniclabs/sonic/pull/954/commits/b2aef343583df6ea2fdfb64aba009d761668851a

HerbertJordan (approved):

PR #956, opened 2026-05-07
Add bundle evaluation cache

LuisPH3:
This PR adds a cache to prevent re-evaluating bundles within the same block number. This resource is to be used by the pool and emitter, as recently promoted transactions may be emitted within the same state context. 

```
	// The cache size of 100k entries is chosen arbitrarily and can be tuned
	// Evaluation of block current-1 are never used again, so we can consider
	// that the cache is only useful for the current block. This means that the
	//  cache contains evaluations for up to 100k envelopes pending of execution.
	//
	// Maximum memory consumption is a constant factor of the following formula:
	// number of bundles in the cache * (size of bundle plan hash + size of block number + size of BundleState)
	// 32  + 8 + ( 1+1 + 24, aligned to 32) = 72 bytes per entry, so ~7.2MiB for 100k entries.
	// Notice that the reasons in the bundle state point to static strings,
	// so they do not contribute to the memory consumption of each entry.
```

### DIFF
--- a/evmcore/bundle_precheck.go
+++ b/evmcore/bundle_precheck.go
@@ -20,13 +20,15 @@ import (
 	"crypto/rand"
 	"fmt"
 	"maps"
-	big "math/big"
+	"math/big"
 
 	"github.com/0xsoniclabs/sonic/evmcore/core_types"
 	"github.com/0xsoniclabs/sonic/gossip/blockproc/bundle"
-	state "github.com/0xsoniclabs/sonic/inter/state"
+	"github.com/0xsoniclabs/sonic/inter/state"
+	"github.com/Fantom-foundation/lachesis-base/common/bigendian"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
+	lru "github.com/hashicorp/golang-lru"
 )
 
 //go:generate mockgen -source=bundle_precheck.go -destination=bundle_precheck_mock.go -package=evmcore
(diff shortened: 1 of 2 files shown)


evmcore/bundle_precheck.go
```diff
@@ -441,3 +443,84 @@ func makeTemporaryBlockedState(reason string) BundleState {
...
+) BundleState {
+
+	if !bundle.IsEnvelope(envelope) {
+		return makeRunnableState()
+	}
+
+	chainId := big.NewInt(int64(chain.GetCurrentNetworkRules().NetworkID))
+	signer := types.LatestSignerForChainID(chainId)
+
+	// Does this even pay off? cache envelope vs cache plan
```
HerbertJordan: nit: left over of development or serious question?

If it is a question still open, maybe add a ticket. If it is resolved, please remove.

HerbertJordan (dismissed):

evmcore/bundle_precheck_test.go
```diff
@@ -1127,3 +1128,175 @@ func createKeys(t *testing.T) ([]*ecdsa.PrivateKey, []common.Address) {
...
+
+	stateDb := state.NewMockStateDB(ctrl)
+	// Expect two evaluations: one for block 100 and one for block 101.
+	stateDb.EXPECT().HasBundleRecentlyBeenProcessed(gomock.Any()).Return(true).Times(2)
+
+	envelope := bundle.NewBuilder().Build()
+	cache := NewBundleEvaluationCache()
+
+	result1 := cache.GetBundleState(chainState, stateDb, envelope)
+	blockNum = 101
```
simonlechner: This new value is not used anywhere.

simonlechner (dismissed):

evmcore/bundle_precheck.go
LuisPH3: left over.

evmcore/bundle_precheck_test.go
LuisPH3: Not very intuitive, but this variable changes the block number as it is captured in the return closure of the `GetLatestHeader` expectation. This variable changes the block number.

simonlechner (approved):

PR #957, opened 2026-05-08
Introduce shared cache between tx pool and emitter to reuse bundle evaluations within the same block number

LuisPH3:
This PR adds a cache shared between the emitter and the transaction pool to cache evaluations of bundles, whenever they can be executed, are blocked or can be discarded. 

The cache is initialized in [config/make_node.go](https://github.com/0xsoniclabs/sonic/compare/luis/tx_pool_bundles_performance?expand=1#diff-e4d4a10709e3b24b15ee28d75652f62379009629d0824f4b170b5be75c22ae60) and then shared with the instances of the pool and the emitter. 

The pool uses this facility in the following locations:
- to validate envelopes as they arrive in the pool for the first time.
- to evaluate if envelopes are dropped from the queued/pending lists. 
- to promote envelopes from queued into pending and vice-versa. 

The emitter uses the facility when evaluating transactions for inclusion in the next event.

Test using `TestBundle_StressWithExpensiveInternalRollback/ComputeHeavy`, this test can emit one envelope per event, therefore creating congestion in the pending list, which forces evaluations for both demoting envelopes to queued and to attempt to emit them again. This test benefits from caching evaluations.

**before** 
image

**after** 
image

### DIFF
--- a/evmcore/tx_pool.go
+++ b/evmcore/tx_pool.go
@@ -197,12 +197,6 @@ type subsidiesCheckFuncFactory func(
 	signer types.Signer,
 ) utils.TransactionCheckFunc
 
-// bundleCheckerFuncFactory is a factory method to create a bundle checker instance.
-type bundleCheckerFuncFactory func(
-	chain StateReader,
-	state state.StateDB,
-) func(*types.Transaction) bundlePoolStatus
-
 // TxPoolConfig are the configuration parameters of the transaction pool.
 type TxPoolConfig struct {
 	Locals    []common.Address // Addresses that should be treated by default as local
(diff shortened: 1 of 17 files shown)


evmcore/tx_pool.go
```diff
@@ -359,13 +353,15 @@ type txpoolResetRequest struct {
...
+	chain StateReader,
+	bundlesCache BundleEvaluator,
+) *TxPool {
 	return newTxPool(
 		config,
 		chainconfig,
 		chain,
 		newSubsidiesChecker,
-		newBundlesChecker,
+		nil,
```
HerbertJordan: shouldn't this be

```suggestion
		bundlesCache,
```

HerbertJordan (dismissed):

HerbertJordan (approved):

PR #959, opened 2026-05-08
Add native prometheus handle

simonlechner:
This PR adds a native Prometheus handle, which is required to support data formats that are not offered by the geth dependency e.g. Histograms.

The default handle from geth is writing data to `/debug/metrics/prometheus`, this PR adds a second handle writing to `/debug/metrics/prometheus/native`. A new metrics path has to be added in order to read the data.


LuisPH3 (dismissed):
The PR introduces a new endpoint at the metrics port (6060) of the sonic client.
This endpoint serves metrics collected via the Prometheus API and not the geth abstraction. 

Prometheus instances scrapping metrics need to be configured to scrap this new endpoint, or metrics will not be collected.

HerbertJordan (dismissed):

LuisPH3 (dismissed):

LuisPH3 (approved):

PR #961, opened 2026-05-11
Remove uses of BigIntToUint256 in Tests and RPC tools

LuisPH3:
Use non-panicking version `BigIntToUint256Clamped`.

Changes in conversion `ethapi.TransactionArgs` into `types.Transaction` sanitize input, this function is used to convert arbitrary user input into core types. Sanitation is justified.


utils/to_ftm_test.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestToFtm_doesNotOverflow256Ints(t *testing.T) {
+	ftm := ToFtm(math.MaxUint64)
+	require.Less(t, ftm.BitLen(), 256, "ToFtm should overflow for large input")
```
HerbertJordan, line 28: nit: the error message makes no sense ..

api/ethapi/transaction_args.go
```diff
@@ -305,8 +305,8 @@ func (args *TransactionArgs) toTransaction() *types.Transaction {
 			ChainID:    uint256.MustFromBig((*big.Int)(args.ChainID)),
 			Nonce:      uint64(*args.Nonce),
 			Gas:        uint64(*args.Gas),
-			GasFeeCap:  utils.BigIntToUint256((*big.Int)(args.MaxFeePerGas)),
-			GasTipCap:  utils.BigIntToUint256((*big.Int)(args.MaxPriorityFeePerGas)),
+			GasFeeCap:  utils.BigIntToUint256Clamped((*big.Int)(args.MaxFeePerGas)),
```
HerbertJordan, line 308: I do not think this should use the Clamped version of the conversion, since it would silently modify the data defined by the user. I think errors should be handled explicitly.

Alternative, you may consider updating the `TransactionArgs` to use `uint256.Int` instead of `big.Int` and fail already in the JSON parsing. Wdyt?

PR #965, opened 2026-05-12
Remove panicking cast in emitter

LuisPH3:
This PR implements defensive programming for panicking cast.
If a transaction with overflowing or negative gas fee or tip reaches the emitter, It will be skipped from emission instead of panicking. Log will be prompted. 

Note, previous code used `utils.BigIntToUint256` which would panic on negative values. The new code allows binary conversion of negative `big.Int` to `uint256.Int`,

### DIFF
--- a/gossip/emitter/emitter.go
+++ b/gossip/emitter/emitter.go
@@ -19,6 +19,7 @@ package emitter
 import (
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


LuisPH3:
Replaced by: 
https://github.com/0xsoniclabs/sonic/pull/971

PR #966, opened 2026-05-12
Metrics: Add emitter gas efficiency

simonlechner:
This PR forwards the gas efficiency from the pre runner to the emitter to log it as a metric.

### DIFF
--- a/evmcore/bundle_precheck.go
+++ b/evmcore/bundle_precheck.go
@@ -399,7 +399,8 @@ func trialRunBundleInternal(
 		}
 	}
 
-	gasEfficiency := 1.0
+	// Calculate the gas efficiency of the bundle and check if it meets the minimum threshold.
+	gasEfficiency := 0.0
 	if summary.ExecutionCost > 0 {
 		gasEfficiency = float64(usedGas) / float64(summary.ExecutionCost)
 	}
--- a/evmcore/bundle_precheck.go
+++ b/evmcore/bundle_precheck.go
@@ -92,7 +92,7 @@ func getBundleState(
 	chain ChainStateForBundleEval,
 	stateDb state.StateDB,
 	envelope *types.Transaction,
-	trialRunner func(*types.Transaction, ChainStateForBundleEval, state.StateDB) (bool, float64),
+	trialRunner func(*types.Transaction, ChainStateForBundleEval, state.StateDB) (float64, bool),
 ) BundleState {
 	chainId := big.NewInt(int64(chain.GetCurrentNetworkRules().NetworkID))
 	signer := types.LatestSignerForChainID(chainId)
@@ -151,8 +151,8 @@ func getBundleState(
 	snapshot := stateDb.InterTxSnapshot()
 	defer stateDb.RevertToInterTxSnapshot(snapshot)
 
-	success, gasEfficiency := trialRunner(envelope, chain, stateDb)
-	if !success {
+	gasEfficiency, valid := trialRunner(envelope, chain, stateDb)
+	if !valid {
 		return makePermanentlyBlockedState("bundle trial-run failed")
 	}
 	return makeRunnableState(gasEfficiency)
@@ -336,14 +336,14 @@ func (t *nonceTracker) restore(backup *nonceTracker) {
 
 // --- Trial Run Logic ---
 
-// trialRunBundle performs a trial run of the bundle on the EVM to check whether
-// it can succeed or not. It returns true if the trial run results in at least
-// one accepted transaction, and false otherwise.
+// trialRunBundle performs a trial run of the bundle on the EVM to check whether it can
+// succeed or not. It returns the gas efficiency (ratio of billed gas to execution cost)
+// and a boolean indicating if the trial run results in at least one accepted transaction.
 func trialRunBundle(
 	envelope *types.Transaction,
 	chain ChainStateForBundleEval,
 	stateDb state.StateDB,
-) (bool, float64) {
+) (float64, bool) {
 	return trialRunBundleInternal(
 		envelope,
 		chain,
@@ -361,14 +361,14 @@ func trialRunBundleInternal(
 	stateDb state.StateDB,
 	factory transactionProcessorFactory,
 	readRandom func([]byte) (int, error),
-) (bool, float64) {
+) (float64, bool) {
 	latestHeader := chain.GetLatestHeader()
 	blobBaseFee := GetBlobBaseFee()
 
 	// Create a random fake-PrevRandao for the trial run.
 	var fakePrevRandao common.Hash
 	if n, err := readRandom(fakePrevRandao[:]); n != len(fakePrevRandao) || err != nil {
-		return false, 0
+		return 0, false
 	}
 
 	// Make sure that everything this function does is reverted at the end.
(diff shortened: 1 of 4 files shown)


evmcore/bundle_precheck.go
LuisPH3: Missing doc
when is efficiency expected and what does it mean

evmcore/bundle_precheck.go
```diff
@@ -150,10 +151,11 @@ func getBundleState(
 	snapshot := stateDb.InterTxSnapshot()
 	defer stateDb.RevertToInterTxSnapshot(snapshot)
 
-	if success := trialRunner(envelope, chain, stateDb); !success {
+	success, gasEfficiency := trialRunner(envelope, chain, stateDb)
```
LuisPH3: nit: I find it a little weird `(bool, float)` instead of `(float, bool)`, but i guess that both values are orthogonal and `efficiency` is still valid for `false` runs.

gossip/emitter/txs.go
```diff
@@ -37,6 +38,14 @@ import (
 	"github.com/0xsoniclabs/sonic/utils/txtime"
 )
 
+var (
+	effectiveBundleGasHistogram = utils.MetricsHistogramWrapper(utils.NewPrometheusHistogram(prometheus.HistogramOpts{
+		Name:    "emitter_bundle_gas_effective",
+		Help:    "Effective gas usage ratio for bundle transactions",
+		Buckets: prometheus.LinearBuckets(0.0, 0.1, 11), // 0, 0.1, 0.2, ..., 1.0
```
LuisPH3: maybe this comment could be less cryptic. 
```suggestion
		Buckets: prometheus.LinearBuckets(0.0, 0.1, 11), // bucket intervals [0, 0.1, 0.2, ..., 1.0]
```

LuisPH3 (changes requested):

LuisPH3 (dismissed):

evmcore/bundle_precheck_test.go
```diff
@@ -1300,3 +1315,153 @@ func Test_BundleEvaluationCache_ReturnsErrorForInvalidEnvelope(t *testing.T) {
...
+		expectedEfficiency float64
+		expectedAccept     bool
+	}{
+		"zero execution cost returns zero efficiency": {
+			processedTxs:       []ProcessedTransaction{{Receipt: &types.Receipt{GasUsed: 50}}},
+			execCost:           core_types.ExecutionCost(0),
+			expectedEfficiency: 0.0,
+			expectedAccept:     false,
+		},
+		"50% efficiency below threshold": {
```
HerbertJordan: nit: the name states 50% efficiency, but the test sets up a 10% efficiency; if the 50% refers to the 50% of to 20% needed, it is a bit contrived. Then just

```suggestion
		"below threshold efficiency": {
```

gossip/emitter/txs.go
```diff
@@ -218,9 +229,14 @@ func (em *Emitter) addTxs(e *inter.MutableEventPayload, sorted *transactionsByPr
...
-		if em.world.GetRules().Upgrades.Brio && bundle.IsEnvelope(resolvedTx) && !em.isValidBundleTx(resolvedTx) {
-			sorted.Pop()
-			continue
+		if em.world.GetRules().Upgrades.Brio && bundle.IsEnvelope(resolvedTx) {
+			gasEfficiency, valid := em.evaluateBundleTx(resolvedTx)
+			if !valid {
+				sorted.Pop()
+				continue
+			} else {
+				effectiveBundleGasHistogram.Update(gasEfficiency)
```
HerbertJordan: This would hide everything below 20% efficiency, right? Those buckets would remain empty. But we would like to see those as well, to know if they are frequently reaching this point.

The function `evaluateBundleTxInternal` should return 3 values:
- was the evaluation attempted (bool)
- the efficiency of the evaluation (float64)
- the valid decision (bool)

If the evaluation was attempted, the efficiency should be reported. The `valid` flag is then used to decide whether to ignore the transaction or continue.

gossip/emitter/txs.go
```diff
@@ -232,50 +248,51 @@ func (em *Emitter) addTxs(e *inter.MutableEventPayload, sorted *transactionsByPr
...
 	if stateDb.HasBundleRecentlyBeenProcessed(plan.Hash()) {
-		return false
+		return 0.0, false
 	}
 
 	// Skip bundles that are not runnable in the current state.
 	adapter := &preCheckChainStateAdapter{external: em.world}
-	return evalBundle.GetBundleState(adapter, stateDb, tx).Executable
+	bundleState := evalBundle.GetBundleState(adapter, stateDb, tx)
+	return bundleState.GasEfficiency, bundleState.Executable
```
HerbertJordan: Alternatively to the suggestion above, to return 3 values, you could also move the metric reporting here. and revert this function to only return a simple bool.

If you move the metric reporting here, you can also extend this function to accept the metric to report to as an argument, enabling you to pass a mock for testing.

gossip/emitter/txs.go
HerbertJordan: Better idea: see comment below.

HerbertJordan (changes requested):

simonlechner:
Screenshot from 2026-05-13 12-22-59

This screenshot shows the heatmap output, running an integration testnet submitting successful bundles with ~25%, ~35% and 100% efficiency in different proportions. 

Visualization: Heatmap
Metric operation: Rate (range 1m for the short running test in the screenshot)
Options, Format: Heatmap (last line of the screenshot, ensures that the buckets of the histogram are not aggregated)

gossip/emitter/txs.go
```diff
@@ -37,6 +38,16 @@ import (
 	"github.com/0xsoniclabs/sonic/utils/txtime"
 )
 
+var (
+	effectiveBundleGasHistogram = utils.MetricsHistogramWrapper(utils.NewPrometheusHistogram(prometheus.HistogramOpts{
+		Name: "emitter_bundle_gas_effective",
+		Help: "Effective gas usage ratio for bundle transactions",
+		// Due to the accuracy of floating point numbers, the width is increase a little
+		// to avoid having 100% efficient bundles in the +inf bucket.
+		Buckets: prometheus.LinearBuckets(0.1, 0.10000000001, 10), // buckets: [0.0, 0.1, ..., 1.0, +inf]
```
HerbertJordan: How about increasing the resolution to 100 buckets, 1% each, for more details? Any objections to that?

gossip/emitter/txs.go
```diff
@@ -232,13 +243,13 @@ func (em *Emitter) addTxs(e *inter.MutableEventPayload, sorted *transactionsByPr
 	}
 }
 
-// isValidBundleTx checks whether the given transaction is a valid bundle that
-// could be emitted by this emitter.
-func (em *Emitter) isValidBundleTx(tx *types.Transaction) bool {
-	return em.isRunnableBundleTxInternal(tx, em.bundleCache)
+// evaluateBundleTx calculates the effective gas of the bundle and whether the
+// given transaction is a valid bundle that could be emitted by this emitter.
+func (em *Emitter) evaluateBundleTx(tx *types.Transaction) bool {
```
HerbertJordan: If this function name is updated, the names of tests targeting it in `txs_test.go` should also be updated.

evmcore/bundle_precheck.go
```diff
@@ -69,6 +69,7 @@ type BundleState struct {
 	Executable         bool     // True if the bundle can be executed with the current state.
 	TemporarilyBlocked bool     // True if the bundle is currently blocked but may become executable later.
 	Reasons            []string // A list of human-readable strings describing why the bundle is not executable or is blocked.
+	GasEfficiency      float64  // The gas efficiency of the bundle, defined as sum of usedGas of receipts / total execution cost.
```
HerbertJordan: It is difficult to derive from this value and the rest whether the `GasEfficiency` was actually evaluated.

I would suggest to make this a `*float64` and have it nil in all cases where the efficiency was not actually evaluated (instead of 0.0). This way, you can monitor every non-nil value in the emitter. Wdyt?

gossip/emitter/txs.go
simonlechner: Since the return type changed again, I reverted the rename of the function.

HerbertJordan (approved):

LuisPH3 (approved):

PR #968, opened 2026-05-12
Fix flaky tests: receipt retrieval timeout.

LuisPH3:
Some flaky tests suffered time-outs while retrieving the receipt from the first transaction executed. This could be caused in multi-node scenarios by having a lack of consensus. 

This PR adds a very large tolerance to consensus delays to avoid stalling during integration test bootstrap. 
Scenarios where bootstrapping the test network is slow and larger than the `StallThresshold` are penalized by stalling, and consensus formation is delayed even more. Default  `StallThresshold` is 30 s and `StallInterval` 60 s. together they approximate the 100 s timeout.


HerbertJordan (approved):
Nice find!

PR #969, opened 2026-05-12
Add check for out-of-bound gas prices in EVM context creation

HerbertJordan:
This PR introduces an integration test reproducing an vulnerability that was detected by an external code reviewer that causes nodes to crash if transactions with gas prices exceeding the 256-bit limit are processed. The PR also includes a corresponding fix for the vulnerability.

Follow-up work will add further safeguards against similar issues.

### DIFF
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -729,7 +729,11 @@ func ApplyTransactionWithEVM(msg *core.Message, config *params.ChainConfig, gp *
 		}
 	}
 	// Create a new context to be used in the EVM environment.
-	txContext := NewEVMTxContext(msg)
+	txContext, err := NewEVMTxContext(msg)
+	if err != nil {
+		statedb.EndTransaction()
+		return nil, fmt.Errorf("failed to create EVM transaction context: %w", err)
+	}
 	evm.SetTxContext(txContext)
 
 	// For now, Sonic only supports Blob transactions without blob data.
@@ -745,6 +749,7 @@ func ApplyTransactionWithEVM(msg *core.Message, config *params.ChainConfig, gp *
 	// Apply the transaction to the current state (included in the env).
 	result, err := core.ApplyMessage(evm, msg, gp)
 	if err != nil {
+		statedb.EndTransaction()
 		return nil, err
 	}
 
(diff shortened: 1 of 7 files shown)


tests/regressions/gas_fee_overflow_test.go
```diff
@@ -0,0 +1,193 @@
...
+		"rejected by RPC": {
+			options:                 tests.IntegrationTestNetOptions{},
+			expectedRejectionReason: "exceeds the configured cap (100.00 FTM)",
+		},
+		"rejected by TxPool": {
+			options: tests.IntegrationTestNetOptions{
+				// Disable the fee-cap check in the RPC
+				ModifyConfig: func(config *config.Config) {
+					config.Opera.RPCTxFeeCap = math.Pow(2, 256) // (FTM)
+				},
```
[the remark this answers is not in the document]
HerbertJordan: math.Pow(2,256) does not overflow float64. `float64` can cover the range all the way up to almost 2^1024

evmcore/state_processor.go
```diff
@@ -745,6 +749,7 @@ func ApplyTransactionWithEVM(msg *core.Message, config *params.ChainConfig, gp *
 	// Apply the transaction to the current state (included in the env).
 	result, err := core.ApplyMessage(evm, msg, gp)
 	if err != nil {
+		statedb.EndTransaction()
```
LuisPH3, line 752: This is surprising, clearly insufficient testing.
Note, all code paths after `	txContext, err := NewEVMTxContext(msg)` end transactions, this could be a `defer`.

LuisPH3 (dismissed):
The changes proposed create a wrapper function around `core.NewEVMTxContext(msg)` to protect against panicking overflows.

evmcore/state_processor.go
HerbertJordan: This function is only used for tracing in RPC calls, and it is practically a copy of the `applyTransaction` also defined in this file. The function is not covered by unit tests at all.

I think the cleanup of this should be its own task.

I would suggest to keep this addition here, to have all branches covering it. Alternatively, I can remove it again, and we keep this branch as it was. Up to you.

LuisPH3 (approved):

PR #971, opened 2026-05-13
Removing panicking cast from big to uint256 in favour of error-returning version.

LuisPH3:
This PR is rather large, but I think it is beneficial to have it atomically to avoid having multiple versions of the function `BigIntToUint256` in use. This PR entirely removes uses of the panicking version. 

To facilitate review, different commits have been separated. 
- The change to the library itself and testing
- changes to the emitter
- changes to the RPC tools and testing. 
- changes to the RPC
- use clamp version in other tests


LuisPH3:
Replaced in favour of multiple PRs

PR #976, opened 2026-05-13
Fix flaky tests where multiple nodes wait for same block.

LuisPH3:
Tests using multiple nodes shall not assume that a block is available in every node immediately. 
This PR adds a tool to wait on block creation.


LorenzSchueler (approved):
I trust you that all tests using multiple nodes have been adjusted.

PR #980, opened 2026-05-14
Fix epoch bounds check truncation in dag_getHeads

thaarok:
The `dag_getHeads` RPC method accepts an epoch number as `rpc.BlockNumber` (int64), which is validated against the current epoch before use. The previous check cast the input to `idx.Epoch` (uint32) before comparing, silently truncating the upper 32 bits.

A call `dag_getHeads(4294968296)` before the fix would return epoch 1000, instead of returning not-existing-epoch error.


LuisPH3 (approved):

jenikd (approved):
Looks good.

PR #981, opened 2026-05-14
Reject lamport underflow in event parent deserialization

thaarok:
A peer-controlled lamportDiff exceeding the event's own lamport causes a silent uint32 underflow, producing a garbage parent ID that stalls in the DAG buffer indefinitely.
Reject such events with ErrMalformedEncoding instead.

This should not break compatibility with older versions, since events with incorrect lamport are already being rejected by consensus in parents_check. This fix only ensure such events does not consume space in the DAG buffer - avoids spamming.

Documented in https://github.com/0xsoniclabs/sonic-admin/issues/781


LuisPH3 (approved):

HerbertJordan (approved):
Great find and fix, thanks!

LuisPH3 (approved):

PR #982, opened 2026-05-15
Avoid sync stalling on empty EventsStreamResponse chunks

thaarok:
Avoid sync stalling on empty EventsStreamResponse chunk receiving.

A Done=false chunk with no Events and no IDs passes checkLenLimits. The resulting zero hash never resolves, leaving a stale entry in processingChunks and blocking further chunk requests for up to ~150 seconds until the session timeouts.


HerbertJordan (approved):
The PR identifies a weakness in the handling of event streaming messages, where an empty message could lead to a stalled synchronization process. The change declares empty stream responses without an done flag as invalid, which will result in a termination of the peer connection.

Thanks for spotting and fixing this issue.

PR #983, opened 2026-05-18
Combine user tx process loops

simonlechner:
To ensure the original behavior is preserved, the user transaction processing loop has been split into pre brio and after functions. To ensure the new code is tested by historical tests, the processing functions are combined in this PR.

### DIFF
--- a/gossip/c_block_callbacks.go
+++ b/gossip/c_block_callbacks.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"cmp"
 	"fmt"
+	"math"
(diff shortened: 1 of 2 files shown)


gossip/c_block_callbacks.go
HerbertJordan, line 1: As discussed in our call, to gain confidence, we should see whether the old `processUserTransactionsNoLimits` function can be gradually upgraded to also support the new features.

simonlechner:
Closed in favor of #988.

PR #984, opened 2026-05-18
Enable sonic RPC namespace by default

jenikd:
This PR enables `sonic` RPC namespace by default. Users can use transaction bundles on the RPC as they are enabled by this flag. `Sonic` tag is no longer experimental, as it was used for the `SCC`, which was moved into separate RPC namespace and warning for an experimental namespace is now used for `scc`.

https://github.com/0xsoniclabs/sonic-admin/issues/778

### DIFF
--- a/cmd/sonicd/app/launcher.go
+++ b/cmd/sonicd/app/launcher.go
@@ -275,8 +275,8 @@ func lachesisMainInternal(
 			if strings.ToLower(module) == "ftm" {
 				log.Warn(fmt.Sprintf("The 'ftm' API is deprecated, use 'eth' instead (--%s).", flag))
 			}
-			if strings.ToLower(module) == "sonic" {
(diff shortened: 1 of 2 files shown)


cmd/sonicd/app/launcher.go
```diff
@@ -275,8 +275,8 @@ func lachesisMainInternal(
 			if strings.ToLower(module) == "ftm" {
 				log.Warn(fmt.Sprintf("The 'ftm' API is deprecated, use 'eth' instead (--%s).", flag))
 			}
-			if strings.ToLower(module) == "sonic" {
-				log.Warn(fmt.Sprintf("The 'sonic' API is experimental and should not be used in production environments (--%s).", flag))
+			if strings.ToLower(module) == "scc" {
```
LuisPH3, line 278: good catch, thanks

LuisPH3 (approved):

LuisPH3 (approved):

PR #985, opened 2026-05-18
Add tooling to test rolled back bundles

LuisPH3:
Bundles are tested in 3 locations, intake, emitter, and block processor. Bundles may be dropped in any of these locations and therefore its worths testing both bundles which roll back both during emission or later in the block processor. 

This PR introduces tooling to use the same tests with both modes, factoring the submission method and the verification. 
Although diff seems quite large, the only change is to nest tests and invoke the new `require`method to verify bundle rollback, or emission skipped.

### DIFF
--- a/tests/bundles/empty_bundle_test.go
+++ b/tests/bundles/empty_bundle_test.go
@@ -17,10 +17,8 @@
 package bundles
 
 import (
-	"context"
 	"math/big"
 	"testing"
-	"time"
 
 	"github.com/0xsoniclabs/sonic/gossip/blockproc/bundle"
 	"github.com/0xsoniclabs/sonic/opera"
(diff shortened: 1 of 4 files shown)


tests/bundles/skip_envelope_tools.go
```diff
@@ -0,0 +1,124 @@
...
+func GetUnfilteredNetVariants(t testing.TB, upgrades opera.Upgrades) iter.Seq2[string, UnfilteredNet] {
+	return func(yield func(string, UnfilteredNet) bool) {
+		if !yield("non-validating net", NewNonValidatedIntegrationTestNet(t, upgrades)) {
+			return
+		}
+		if !yield("bypass-pool net", NewBypassPoolIntegrationTestNet(t, upgrades)) {
+			return
+		}
+	}
+}
```
LorenzSchueler, line 44: Not a suggestion, just a question: Why not use `map[string]UnfilteredNet`? Because you want to create the nets lazily, so they are not running both at the same time?

tests/bundles/skip_envelope_tools.go
```diff
@@ -0,0 +1,124 @@
...
+type BypassPoolNet struct {
+	tests.IntegrationTestNetSession
+}
+
+// NewBypassPoolIntegrationTestNet creates a test network where the TxPool is
+// completely bypassed by sending envelopes directly to the consensus engine.
+// This allows testing bundle execution in a scenario where transactions reach
+// the block processor, with the intention of testing its resilience against
+// invalid transactions.
+func NewBypassPoolIntegrationTestNet(t testing.TB, upgrades opera.Upgrades) *BypassPoolNet {
```
LorenzSchueler: Naming can be improved. `NonValidating` sound to me like it disables more validation than `BypassPool`, but as far as I understand, the opposite is the case: `NonValidating` only disables validation on intake, and `BypassPool` disables validation on intake and emission (by bypassing these steps).
What about `NoIntakeValidationTestNet` and `NoIntakeAndEmissionValidationTestNet`?

tests/bundles/empty_bundle_test.go
```diff
@@ -33,104 +31,97 @@ func TestBundle_BundleContainingAnyEmptyGroupIsRejected(t *testing.T) {
...
-			ExpectReject: false,
-		},
-		"Layered/OnlyEmptySubGroups": {
-			root: bundle.AllOf(
-				bundle.AllOf(),
-			),
-			ExpectReject: true,
-		},
-	}
+	for mode, net := range GetUnfilteredNetVariants(t, upgrades) {
```
LorenzSchueler, line 34: This diff is horrible. Sometimes I wonder whether `git diff` could use some AI to figure out which lines correspond to each other 😆

LorenzSchueler (dismissed):

tests/bundles/skip_envelope_tools.go
LuisPH3: correct

tests/bundles/empty_bundle_test.go
LuisPH3: this is bad UX in github, when you open this diff in vs code, you see different shades of green between lines which have been indented and code changes.
This is a core feature of git: `git diff -w`

LorenzSchueler: Or even better you use `git diff` with `difftastic` 😉

tests/bundles/skip_envelope_tools.go
```diff
@@ -0,0 +1,124 @@
...
+
+// GetUnfilteredNetVariants returns a sequence of different unfiltered network variants to test against.
+// Each variant represents a different way of bypassing transaction validation, allowing testing of bundle execution
+// in various scenarios where transactions may not be filtered by the TxPool or the emitter.
+func GetUnfilteredNetVariants(t testing.TB, upgrades opera.Upgrades) iter.Seq2[string, UnfilteredNet] {
+	return func(yield func(string, UnfilteredNet) bool) {
+		if !yield("non-validating net", NewNonValidatedIntegrationTestNet(t, upgrades)) {
+			return
+		}
+		if !yield("bypass-pool net", NoNewIntakeAndEmissionValidationTestNet(t, upgrades)) {
```
LorenzSchueler:
```suggestion
		if !yield("no-intake-validation net", NewNoIntakeValidationIntegrationTestNet(t, upgrades)) {
			return
		}
		if !yield("no-intake-and-emission-validation net", NewNoIntakeAndEmissionValidationTestNet(t, upgrades)) {
```

tests/bundles/skip_envelope_tools.go
```diff
@@ -0,0 +1,124 @@
...
+type NoIntakeValidationTestNet struct {
+	tests.IntegrationTestNetSession
+}
+
+// NewNonValidatedIntegrationTestNet creates a test network where the TxPool
+// validation is disabled, allowing transactions that would normally be synchronously
+// rejected to be pooled. These transactions may not be promoted or emitted in events.
+// Therefore execution of these transactions is not guaranteed and may still be
+// filtered by the frontend of the system.
+func NewNonValidatedIntegrationTestNet(t testing.TB, upgrades opera.Upgrades) *NoIntakeValidationTestNet {
```
LorenzSchueler:
```suggestion
func NewNoIntakeValidationIntegrationTestNet(t testing.TB, upgrades opera.Upgrades) *NoIntakeValidationTestNet {
```

tests/bundles/skip_envelope_tools.go
```diff
@@ -0,0 +1,124 @@
...
+type NoIntakeAndEmissionValidationTestNet struct {
+	tests.IntegrationTestNetSession
+}
+
+// NoNewIntakeAndEmissionValidationTestNet creates a test network where the TxPool is
+// completely bypassed by sending envelopes directly to the consensus engine.
+// This allows testing bundle execution in a scenario where transactions reach
+// the block processor, with the intention of testing its resilience against
+// invalid transactions.
+func NoNewIntakeAndEmissionValidationTestNet(t testing.TB, upgrades opera.Upgrades) *NoIntakeAndEmissionValidationTestNet {
```
LorenzSchueler:
```suggestion
func NewNoIntakeAndEmissionValidationTestNet(t testing.TB, upgrades opera.Upgrades) *NoIntakeAndEmissionValidationTestNet {
```

LorenzSchueler (dismissed):

LorenzSchueler (approved):

PR #987, opened 2026-05-19
Rewrite throttler integration test to reduce flackiness

LuisPH3:
This PR attempts to reduce flakiness in test by:

- Replace `time.Sleep(1 * time.Second)` with a polling loop that waits until at least ~60 events exist (with a generous timeout).

### DIFF
--- a/tests/validator_stakes/low_stake_validator_does_not_emit_test.go
+++ b/tests/validator_stakes/low_stake_validator_does_not_emit_test.go
@@ -54,18 +54,18 @@ func TestEventThrottler_NonDominantValidatorsProduceLessEvents_WhenEventThrottle
 				ClientExtraArguments: extraArguments,
 			})
 
-			client, err := net.GetClient()
-			require.NoError(t, err)
-			defer client.Close()
-
 			net.AdvanceEpoch(t, 1)
 
-			// wait until some events are generated
-			time.Sleep(1 * time.Second)
+			// Poll until enough events are collected for statistical stability
+			const minEvents = 60
+			var eventsInEpoch eventMap
+			require.Eventually(t, func() bool {
+				eventsInEpoch = getEventsInEpoch(t, net)
+				return len(eventsInEpoch) >= minEvents
+			}, 100*time.Second, 200*time.Millisecond,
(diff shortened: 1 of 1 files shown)


tests/validator_stakes/low_stake_validator_does_not_emit_test.go
```diff
@@ -54,28 +54,35 @@ func TestEventThrottler_NonDominantValidatorsProduceLessEvents_WhenEventThrottle
...
-				// Without emitter throttling, both validators should create the same amount of events
-				require.InDelta(t, percentages[1], percentages[2], 0.1,
-					"Both validators should create equal amount of events")
+				// Without throttling, neither validator should dominate by more than 3x.
+				require.Less(t, highStake, 3*lowStake,
+					"Without throttling, high stake validator should not dominate (got %d vs %d)",
+					highStake, lowStake)
+				require.Less(t, lowStake, 3*highStake,
+					"Without throttling, low stake validator should not be starved (got %d vs %d)",
+					lowStake, highStake)
```
LorenzSchueler: If we consider >3x **strong** evidence for throttling, then <3x is not **strong** evidence for non-throttling. We need a safety margin here. Does >5x and <2x work?

tests/validator_stakes/low_stake_validator_does_not_emit_test.go
```diff
@@ -158,10 +158,8 @@ func fetchEvent(t *testing.T, client *tests.PooledEhtClient, eventID hash.Event)
 }
 
 func calculateValidatorEmissionPercentages(
-	t *testing.T,
 	allEvents eventMap,
 ) map[idx.ValidatorID]float64 {
-	t.Helper()
```
LorenzSchueler, line 163: Not sure if this was intentional or just not reverted correctly, I don't care either way.

LorenzSchueler (approved):

tests/validator_stakes/low_stake_validator_does_not_emit_test.go
LuisPH3: there was no use to pass t. go cleaned up.

PR #988, opened 2026-05-19
Limit block size in state processor

simonlechner:
The enforcing of the block size limit ([eip-7934](https://eips.ethereum.org/EIPS/eip-7934)) is moved to the state processor to increase test coverage with historical tests.

### DIFF
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -258,8 +258,8 @@ func runTransactions(
 	}
 }
 
-// realTxSize returns the size of the transaction,
-// including any overhead introduced by sponsorship requests.
+// realTxSize returns the size of the transaction, including any
+// overhead introduced by sponsorship requests and bundles.
 func realTxSize(context *runContext, tx *types.Transaction) uint64 {
 	if context.upgrades.TransactionBundles && bundle.IsEnvelope(tx) {
 		size := uint64(0)
@@ -267,6 +267,9 @@ func realTxSize(context *runContext, tx *types.Transaction) uint64 {
 		if err != nil {
 			return size // = 0, in case the bundle can not be unpacked, it will not be added to the block.
 		}
+
+		// The full size of the bundle have to be considered even if not
+		// all transactions of the bundle are included in the block.
 		for _, innerTx := range txBundle.Transactions {
 			size += realTxSize(context, innerTx)
 		}
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -86,6 +86,7 @@ func NewStateProcessorForReplay(
 type ProcessSummary struct {
 	ProcessedTransactions []ProcessedTransaction
 	ExecutionCost         core_types.ExecutionCost
+	CausedBy              map[common.Hash]common.Hash
 }
 
 // ProcessedTransaction represents a transaction that was considered for
@@ -121,10 +122,10 @@ type ProcessedTransaction struct {
 // consistent.
 func (p *StateProcessor) Process(
 	block *EvmBlock, statedb state.StateDB, cfg vm.Config, gasLimit uint64,
-	usedGas *uint64, trueTxOffset int, onNewLog func(*core_types.Log),
+	usedGas *uint64, trueTxOffset int, onNewLog func(*core_types.Log), remainingSize uint64,
 ) ProcessSummary {
 	sonicDifficulty := big.NewInt(1)
-	return p.ProcessWithDifficulty(block, statedb, cfg, gasLimit, usedGas, trueTxOffset, onNewLog, sonicDifficulty)
+	return p.ProcessWithDifficulty(block, statedb, cfg, gasLimit, usedGas, trueTxOffset, onNewLog, sonicDifficulty, remainingSize)
 }
 
 // ProcessWithDifficulty is the same as Process, but allows specifying a custom
@@ -134,7 +135,7 @@ func (p *StateProcessor) Process(
 func (p *StateProcessor) ProcessWithDifficulty(
 	block *EvmBlock, statedb state.StateDB, cfg vm.Config, gasLimit uint64,
 	usedGas *uint64, trueTxOffset int, onNewLog func(*core_types.Log),
-	difficulty *big.Int,
+	difficulty *big.Int, remainingSize uint64,
 ) ProcessSummary {
 	var (
 		gp           = core.NewGasPool(gasLimit)
@@ -155,7 +156,7 @@ func (p *StateProcessor) ProcessWithDifficulty(
 	return runTransactions(newRunContext(
 		signer, header.BaseFee, statedb, gp, blockNumber, block.Time, usedGas,
 		onNewLog, p.upgrades, &transactionRunner{evm{vmenv}}, p.forReplay,
-	), block.Transactions, trueTxOffset)
+	), block.Transactions, trueTxOffset, remainingSize)
 }
 
 // runContext bundles the parameters required for processing transactions in a
@@ -217,9 +218,11 @@ func runTransactions(
 	context *runContext,
 	transactions types.Transactions,
 	trueTxIndexOffset int,
+	remainingSize uint64,
 ) ProcessSummary {
 	processedTxs := make([]ProcessedTransaction, 0, len(transactions))
 	totalExecCost := core_types.ExecutionCost(0)
+	causedBy := make(map[common.Hash]common.Hash)
 	for _, tx := range transactions {
 		// Bundle-only transactions are only valid within a bundle and must
 		// be rejected at the top level when processing the head state.
@@ -228,12 +231,21 @@ func runTransactions(
 			continue
 		}
 
+		neededSpace := realTxSize(context, tx)
+		if context.upgrades.Brio && neededSpace > remainingSize {
+			// Add the transaction to the processed ones with a nil receipt to increase skipped counter.
+			processedTxs = append(processedTxs, ProcessedTransaction{Transaction: tx})
+			continue
+		}
+
 		txs, _, execCost := runTransaction(context, tx, trueTxIndexOffset)
 		totalExecCost += execCost
 
-		for _, tx := range txs {
-			if tx.Receipt != nil { // < only transactions included in the block
+		for _, processedTx := range txs {
+			if processedTx.Receipt != nil { // < only transactions included in the block
 				trueTxIndexOffset++
+				remainingSize -= processedTx.Transaction.Size()
+				causedBy[processedTx.Transaction.Hash()] = tx.Hash()
 			}
 		}
 		processedTxs = append(processedTxs, txs...)
@@ -242,7 +254,32 @@ func runTransactions(
 	return ProcessSummary{
 		ProcessedTransactions: processedTxs,
 		ExecutionCost:         totalExecCost,
+		CausedBy:              causedBy,
 	}
+}
+
+// realTxSize returns the size of the transaction,
+// including any overhead introduced by sponsorship requests.
+func realTxSize(context *runContext, tx *types.Transaction) uint64 {
+	if context.upgrades.TransactionBundles && bundle.IsEnvelope(tx) {
+		size := uint64(0)
+		txBundle, _, err := bundle.ValidateEnvelope(context.signer, tx)
+		if err != nil {
+			return size // = 0, in case the bundle can not be unpacked, it will not be added to the block.
+		}
+		for _, innerTx := range txBundle.Transactions {
+			size += realTxSize(context, innerTx)
+		}
+		return size
+	}
+
+	if context.upgrades.GasSubsidies && subsidies.IsSponsorshipRequest(tx) {
+		size := tx.Size()
+		size += subsidies.RlpEncodedFeeChargingTxSizeInBytes
+		return size
+	}
+
+	return tx.Size()
 }
 
 // runTransaction processes the given transaction and returns a list of all
@@ -713,7 +750,7 @@ func (tp *TransactionProcessor) Run(i int, tx *types.Transaction) ProcessSummary
 		tp.signer, tp.header.BaseFee, tp.stateDb, tp.gp, tp.blockNumber, tp.blockTime,
 		&tp.usedGas, tp.onNewLog, tp.upgrades, &transactionRunner{evm{tp.vmEnvironment}},
 		false,
-	), []*types.Transaction{tx}, i)
+	), []*types.Transaction{tx}, i, math.MaxUint64)
 }
 
 // ApplyTransactionWithEVM attempts to apply a transaction to the given state database
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -237,8 +237,14 @@ func runTransactions(
 		for _, processedTx := range txs {
 			if processedTx.Receipt != nil { // < only transactions included in the block
 				trueTxIndexOffset++
-				remainingSize -= processedTx.Transaction.Size()
 				causedBy[processedTx.Transaction.Hash()] = tx.Hash()
+
+				if remainingSize < processedTx.Transaction.Size() {
+					log.Debug("Block size limit exceeded,", "tx", processedTx.Transaction.Hash().Hex(),
+						"txSize", processedTx.Transaction.Size(), "remainingSize", remainingSize)
+					break
+				}
+				remainingSize -= processedTx.Transaction.Size()
 			}
 		}
 		processedTxs = append(processedTxs, txs...)
@@ -362,7 +368,8 @@ func (r *transactionRunner) runSponsoredTransaction(
 ) ([]ProcessedTransaction, core_types.TransactionResult) {
 	// check whether the size limit is large enough for the transaction and its payment
 	if tx.Size()+subsidies.RlpEncodedFeeChargingTxSizeInBytes > sizeLimit {
-		log.Debug("Transaction skipped due to block size limit", "tx", tx.Hash().Hex(), "txSize", tx.Size(), "estimatedPaymentTxSize", subsidies.RlpEncodedFeeChargingTxSizeInBytes, "sizeLimit", sizeLimit)
+		log.Debug("Transaction skipped due to block size limit",
+			"tx", tx.Hash().Hex(), "txSize", tx.Size(), "estimatedPaymentTxSize", subsidies.RlpEncodedFeeChargingTxSizeInBytes, "sizeLimit", sizeLimit)
 		return []ProcessedTransaction{{Transaction: tx}}, core_types.TransactionResultInvalid
 	}
 
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -242,9 +242,10 @@ func runTransactions(
 				if remainingSize < processedTx.Transaction.Size() {
 					log.Debug("Block size limit exceeded,", "tx", processedTx.Transaction.Hash().Hex(),
 						"txSize", processedTx.Transaction.Size(), "remainingSize", remainingSize)
-					break
+					remainingSize = 0
+				} else {
+					remainingSize -= processedTx.Transaction.Size()
 				}
-				remainingSize -= processedTx.Transaction.Size()
 			}
 		}
 		processedTxs = append(processedTxs, txs...)
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -261,7 +261,7 @@ func runTransactions(
 // computeRealTxSize returns the size of the transaction, including any
 // overhead introduced by sponsorship requests and bundles.
 func computeRealTxSize(context *runContext, tx *types.Transaction) uint64 {
-	if context.upgrades.TransactionBundles && bundle.IsEnvelope(tx) {
+	if context.upgrades.Brio && context.upgrades.TransactionBundles && bundle.IsEnvelope(tx) {
 		size := uint64(0)
 		txBundle, err := bundle.OpenEnvelope(context.signer, tx)
 		if err != nil {
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -84,9 +84,9 @@ func NewStateProcessorForReplay(
 // processing, including gas from rolled-back bundles, which distinguishes it
 // from the usedGas counter that gets reverted on snapshot rollback.
 type ProcessSummary struct {
-	ProcessedTransactions []ProcessedTransaction
-	ExecutionCost         core_types.ExecutionCost
-	CausedBy              map[common.Hash]common.Hash
+	ProcessedTransactions []ProcessedTransaction      // List of processed transactions with their receipts (nil receipt for skipped transactions).
+	ExecutionCost         core_types.ExecutionCost    // Total execution cost (gas used) for all processed transactions, including rolled-back bundles.
+	CausedBy              map[common.Hash]common.Hash // Mapping from transaction hash to the hash of the transaction that caused it.
 }
 
 // ProcessedTransaction represents a transaction that was considered for
@@ -231,7 +231,7 @@ func runTransactions(
 			continue
 		}
 
-		neededSpace := realTxSize(context, tx)
+		neededSpace := computeRealTxSize(context, tx)
 		if context.upgrades.Brio && neededSpace > remainingSize {
 			// Add the transaction to the processed ones with a nil receipt to increase skipped counter.
 			processedTxs = append(processedTxs, ProcessedTransaction{Transaction: tx})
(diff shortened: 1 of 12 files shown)


evmcore/state_processor.go
```diff
@@ -242,9 +254,34 @@ func runTransactions(
...
+		CausedBy:              causedBy,
 	}
 }
 
+// realTxSize returns the size of the transaction,
+// including any overhead introduced by sponsorship requests.
+func realTxSize(context *runContext, tx *types.Transaction) uint64 {
+	if context.upgrades.TransactionBundles && bundle.IsEnvelope(tx) {
+		size := uint64(0)
+		txBundle, _, err := bundle.ValidateEnvelope(context.signer, tx)
```
LuisPH3: Avoid re-validating.


```suggestion
     txBundle, err := bundle.OpenEnvelope(context.signer, tx)
```

evmcore/state_processor.go
```diff
@@ -242,9 +254,34 @@ func runTransactions(
...
 		ProcessedTransactions: processedTxs,
 		ExecutionCost:         totalExecCost,
+		CausedBy:              causedBy,
 	}
 }
 
+// realTxSize returns the size of the transaction,
+// including any overhead introduced by sponsorship requests.
+func realTxSize(context *runContext, tx *types.Transaction) uint64 {
+	if context.upgrades.TransactionBundles && bundle.IsEnvelope(tx) {
```
LuisPH3:
```suggestion
	if context.upgrades.Brio && context.upgrades.TransactionBundles && bundle.IsEnvelope(tx) {
```
before brio envelopes are normal txs

evmcore/state_processor.go
```diff
@@ -228,12 +231,21 @@ func runTransactions(
...
+			processedTxs = append(processedTxs, ProcessedTransaction{Transaction: tx})
+			continue
+		}
+
 		txs, _, execCost := runTransaction(context, tx, trueTxIndexOffset)
 		totalExecCost += execCost
 
-		for _, tx := range txs {
-			if tx.Receipt != nil { // < only transactions included in the block
+		for _, processedTx := range txs {
```
LuisPH3, line 237: nit: if the collection is txs, the iterator should be tx. 
there is a variable in context `processedTxs` which makes this a little weird.

evmcore/state_processor.go
```diff
@@ -242,9 +254,34 @@ func runTransactions(
 	return ProcessSummary{
 		ProcessedTransactions: processedTxs,
 		ExecutionCost:         totalExecCost,
+		CausedBy:              causedBy,
 	}
 }
 
+// realTxSize returns the size of the transaction,
+// including any overhead introduced by sponsorship requests.
```
LorenzSchueler:
```suggestion
// including any overhead introduced by sponsorship requests or bundles.
```

evmcore/state_processor.go
```diff
@@ -228,12 +231,21 @@ func runTransactions(
...
 
-		for _, tx := range txs {
-			if tx.Receipt != nil { // < only transactions included in the block
+		for _, processedTx := range txs {
+			if processedTx.Receipt != nil { // < only transactions included in the block
 				trueTxIndexOffset++
+				remainingSize -= processedTx.Transaction.Size()
+				causedBy[processedTx.Transaction.Hash()] = tx.Hash()
 			}
 		}
```
LorenzSchueler, line 250: This uses the size which includes the full overhead of bundles in the check before executing but then subtracts the size of the transactions that made it into the block. This means the block might not be completely filled, but I think that should be fine, since it is essentially the same as execution cost vs. used gas.

evmcore/state_processor.go
```diff
@@ -242,9 +254,34 @@ func runTransactions(
 	return ProcessSummary{
 		ProcessedTransactions: processedTxs,
 		ExecutionCost:         totalExecCost,
+		CausedBy:              causedBy,
 	}
 }
 
+// realTxSize returns the size of the transaction,
+// including any overhead introduced by sponsorship requests.
+func realTxSize(context *runContext, tx *types.Transaction) uint64 {
```
LuisPH3: nit: function needs a verb.
`computeTxSize` or `GetTxSize` .

evmcore/state_processor_test.go
```diff
@@ -3817,3 +3818,83 @@ func TestRunTransaction_Bundle_ReturnsExecutionCostFromBundleExecution(t *testin
...
+			context := &runContext{signer: signer, upgrades: opera.Upgrades{
+				Brio:               true,
+				TransactionBundles: true,
+				GasSubsidies:       true,
+			}}
+			size := realTxSize(context, test.tx)
+			require.Equal(t, test.expectedSize, size)
+		})
+	}
+}
```
LorenzSchueler, line 4014: A test for running into the size limit is missing.

LorenzSchueler (changes requested):

evmcore/state_processor.go
simonlechner: Yes, I added a comment in the `realTxSize` function during the size calculation.

gossip/c_block_callbacks_test.go
LuisPH3, line 1: I think that we need a new test, but it may be that this is already covered by an old test:
in  `processUserTransactions`
> block size can be consumed with internal txs only, all user transactions are skipped.

evmcore/state_processor.go
```diff
@@ -86,6 +86,7 @@ func NewStateProcessorForReplay(
 type ProcessSummary struct {
 	ProcessedTransactions []ProcessedTransaction
 	ExecutionCost         core_types.ExecutionCost
+	CausedBy              map[common.Hash]common.Hash
```
LuisPH3: this could use some docu, the reason is that hashes could be anything, therefore it will be nice to say this maps tx -> tx.

gossip/c_block_callbacks.go
```diff
@@ -639,54 +618,38 @@ func processUserTransactions(
...
-				if processed.Receipt != nil { // < nil if skipped
-					blockBuilder.AddTransaction(
-						processed.Transaction,
-						processed.Receipt,
-					)
-					remainingGas -= processed.Receipt.GasUsed
-					remainingSize -= processed.Transaction.Size()
-					causedBy[processed.Transaction.Hash()] = tx.Hash()
-				}
+	var remainingSize uint64
```
LuisPH3: nit:
```
		remainingSize := math.MaxUint64
        if !upgrades.Brio {
		      remainingSize = uint64(params.MaxBlockSize - rlpEncodedMaxHeaderSizeInBytes)
        }
```

reduce branches, remove zero init variable... does not change anything but it is easier on the eyes

evmcore/state_processor_test.go
LuisPH3, line 1: Are we missing a test: 
`StateProcessor_Process` collects CausedBy for bundles and sponsored txs.

evmcore/state_processor_test.go
```diff
@@ -3817,3 +3818,204 @@ func TestRunTransaction_Bundle_ReturnsExecutionCostFromBundleExecution(t *testin
...
+					runner.EXPECT().runRegularTransaction(ctx, regular2, 1).Return(
+						ProcessedTransaction{Transaction: regular2, Receipt: &types.Receipt{}},
+						core_types.TransactionResultSuccessful,
+					),
+				)
+			},
+			wantProcessed: 3,
+			wantIncluded:  2,
+		},
+		"transaction followed by sponsored transactions": {
```
LuisPH3: There is a behavior that is not clear to me: a sponsorship needs to "reserve" space for the payment internal tx. there should be a test which tests this.  there is `"transaction followed by sponsored transactions"`, which may test this, but it would be nice to have it more explicitly, maybe just rename/add cases: 

- "not enough space for sponsorship payment, skips tx"
- "not enough space for sponsorship payment after normal tx, skips tx"

evmcore/state_processor_test.go
```diff
@@ -3817,3 +3818,204 @@ func TestRunTransaction_Bundle_ReturnsExecutionCostFromBundleExecution(t *testin
...
+			setupMock: func(ctx *runContext, runner *Mock_transactionRunner) {
+				runner.EXPECT().runRegularTransaction(ctx, regular1, 0).Return(
+					ProcessedTransaction{Transaction: regular1, Receipt: &types.Receipt{}},
+					core_types.TransactionResultSuccessful,
+				)
+			},
+			wantProcessed: 2,
+			wantIncluded:  1,
+		},
+		"bundle transaction exceeding limit": {
```
LuisPH3: add case where bundle fits in.

evmcore/state_processor_test.go
```diff
@@ -3817,3 +3818,204 @@ func TestRunTransaction_Bundle_ReturnsExecutionCostFromBundleExecution(t *testin
...
+	sponsored := getSponsorshipRequest(t)
+
+	tests := map[string]struct {
+		txs           []*types.Transaction
+		remainingSize uint64
+		setupMock     func(*runContext, *Mock_transactionRunner)
+		wantProcessed int
+		wantIncluded  int // transactions with non-nil receipt
+	}{
+		"single large transaction": {
```
LuisPH3: make two cases:

- single large transaction fits in
- single large transaction does not fit in

evmcore/state_processor_test.go
```diff
@@ -3817,3 +3818,204 @@ func TestRunTransaction_Bundle_ReturnsExecutionCostFromBundleExecution(t *testin
...
+			txs:           []*types.Transaction{regular1},
+			remainingSize: regular1.Size() - 1, // too small
+			setupMock:     func(_ *runContext, _ *Mock_transactionRunner) {},
+			wantProcessed: 1,
+			wantIncluded:  0,
+		},
+		"multiple transactions exceeding limit": {
```
LuisPH3: add case, multiple transactions  fit in

gossip/c_block_callbacks_test.go
simonlechner: [TestProcessUserTransactions_SkipUserTransactionIfInternalTransactionsExceedBlockSizeLimit](https://github.com/0xsoniclabs/sonic/blob/30201ccea1332a1daaf5285286431d85e9836bb4/gossip/c_block_callbacks_test.go#L1375)

evmcore/state_processor.go
simonlechner: Bundles can not be turned on pre Brio, the original code already treats envelop transactions as normal transactions if TransactionBundles are not enabled.

evmcore/state_processor.go
simonlechner: There is already a variable called `tx`, which has been shadowed in the original implementation. It is a processed transaction that is appended to the `processedTxs`. Please suggest a new name if this is not fitting.

evmcore/state_processor.go
LuisPH3: Bundles should not be turned on pre Brio, but: is this even checked?

evmcore/state_processor_test.go
simonlechner: I added new test with the passing test cases for all transaction types.

LorenzSchueler:
Bertha run up to 5M blocks was successful.

evmcore/state_processor.go
```diff
@@ -242,7 +254,35 @@ func runTransactions(
 	return ProcessSummary{
 		ProcessedTransactions: processedTxs,
 		ExecutionCost:         totalExecCost,
+		CausedBy:              causedBy,
+	}
+}
+
+// computeRealTxSize returns the size of the transaction, including any
+// overhead introduced by sponsorship requests and bundles.
+func computeRealTxSize(context *runContext, tx *types.Transaction) uint64 {
```
HerbertJordan: For bundles, this function actually computes an upper bound of the size, since it assumes all transactions listed in the bundle are included. Maybe rename it to `computeMaxTxSize` ?

evmcore/state_processor.go
```diff
@@ -258,17 +258,20 @@ func runTransactions(
...
 }
 
-// realTxSize returns the size of the transaction,
-// including any overhead introduced by sponsorship requests.
-func realTxSize(context *runContext, tx *types.Transaction) uint64 {
-	if context.upgrades.TransactionBundles && bundle.IsEnvelope(tx) {
+// computeRealTxSize returns the size of the transaction, including any
+// overhead introduced by sponsorship requests and bundles.
+func computeRealTxSize(context *runContext, tx *types.Transaction) uint64 {
+	if context.upgrades.Brio && context.upgrades.TransactionBundles && bundle.IsEnvelope(tx) {
```
LorenzSchueler: I'm not sure whether this should maybe only check for brio, not bundles. Because brio controls if we apply bundle semantics, and the bundle flag controls only if we execute bundles.

evmcore/state_processor_test.go
```diff
@@ -3817,3 +3821,390 @@ func TestRunTransaction_Bundle_ReturnsExecutionCostFromBundleExecution(t *testin
 	_, _, execCost := runTransaction(context, tx, 0)
 	require.Equal(t, core_types.ExecutionCost(150_000), execCost)
 }
+
+func TestRealTransactionSize_CanHandleAllTxTypes(t *testing.T) {
```
HerbertJordan: nit: 
```suggestion
func TestComputeRealTxSize_CanHandleAllTxTypes(t *testing.T) {
```

evmcore/state_processor_test.go
```diff
@@ -3817,3 +3821,390 @@ func TestRunTransaction_Bundle_ReturnsExecutionCostFromBundleExecution(t *testin
...
+	bundleTx := bundle.AllOf(bundle.Step(key, basicTx)).Build()
+
+	bundleWithSponsorTx := bundle.AllOf(
+		bundle.Step(key, basicTx),
+		bundle.Step(key, &types.AccessListTx{
+			To:       &common.Address{0x42},
+			GasPrice: big.NewInt(0), // < makes it a sponsorship request
+		}),
+	).Build()
+
```
HerbertJordan, line 3967: A case with a nested bundle is missing; something like
```
	bundleTxWithNestedBundle := bundle.AllOf(
		bundle.Step(key, bundleTx),
		bundle.Step(key, bundleWithSponsorTx),
	).Build()
```
with
```
expectedBundleTxWithNestedBundleSize := expectedBundleSize + expectedBundleWithSponsorSize
```

evmcore/state_processor_test.go
```diff
@@ -3817,3 +3821,390 @@ func TestRunTransaction_Bundle_ReturnsExecutionCostFromBundleExecution(t *testin
...
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			runner := NewMock_transactionRunner(ctrl)
+
+			context := &runContext{
+				upgrades: opera.Upgrades{Brio: true, GasSubsidies: true, TransactionBundles: true},
+				signer:   types.LatestSignerForChainID(big.NewInt(1)),
+				runner:   runner,
+			}
```
HerbertJordan: A test checking that this limit is not enforced before Brio should be included -- either as a dedicated test or as an extension of this test.

evmcore/state_processor_test.go
```diff
@@ -3817,3 +3821,390 @@ func TestRunTransaction_Bundle_ReturnsExecutionCostFromBundleExecution(t *testin
...
+					},
+					core_types.TransactionResultSuccessful,
+				)
+			},
+			wantProcessed: 2,
+			wantIncluded:  2,
+		},
+		"bundle transaction": {
+			txs:           []*types.Transaction{bundle.AllOf(bundle.Step(key, regular1)).Build()},
+			remainingSize: regular1.Size() + 1024, // Add some extra space for the bundle overhead
```
HerbertJordan:
```suggestion
			remainingSize: regular1.Size() + 1024, // Add some extra space for the bundle marker added by the builder
```

evmcore/state_processor.go
```diff
@@ -228,12 +231,21 @@ func runTransactions(
 			continue
 		}
 
+		neededSpace := computeRealTxSize(context, tx)
```
HerbertJordan: Instead of computing an upper boundary of the required size here, it would be cleaner to pass the limit to the `runTransaction` function below, and check the limit while processing the transactions.

This would:
 - avoid the need for the `computeRealTxSize` function
 - avoid the implicit need to keep the transaction processing in sync with the `comuteRealTxSize` (single point of definition)
 - capture the used size accurately

Wdyt?

evmcore/state_processor.go
```diff
@@ -228,12 +231,21 @@ func runTransactions(
...
+
 		txs, _, execCost := runTransaction(context, tx, trueTxIndexOffset)
 		totalExecCost += execCost
 
-		for _, tx := range txs {
-			if tx.Receipt != nil { // < only transactions included in the block
+		for _, processedTx := range txs {
+			if processedTx.Receipt != nil { // < only transactions included in the block
 				trueTxIndexOffset++
+				remainingSize -= processedTx.Transaction.Size()
```
HerbertJordan: Defensive coding would have to assume here that the size computed above, by `computeRealTxSize` may be wrong, in which case this could lead to an underflow.

evmcore/state_processor.go
LuisPH3: Before brio, tx is a normal tx. after brio it should not reach this code path, until bundles are enabled.
I think as well that we could remove `context.upgrades.TransactionBundles`

simonlechner: No longer applicable.

evmcore/state_processor.go
simonlechner: No longer applicable.

evmcore/state_processor_test.go
simonlechner: No longer applicable.

evmcore/state_processor_test.go
simonlechner: No longer applicable.

evmcore/state_processor_test.go
simonlechner: No longer applicable.

evmcore/state_processor.go
simonlechner: No longer applicable.

evmcore/state_processor.go
```diff
@@ -228,12 +231,20 @@ func runTransactions(
...
+			if processedTx.Receipt != nil { // < only transactions included in the block
 				trueTxIndexOffset++
+				causedBy[processedTx.Transaction.Hash()] = tx.Hash()
+
+				if remainingSize < processedTx.Transaction.Size() {
+					log.Debug("Block size limit exceeded,", "tx", processedTx.Transaction.Hash().Hex(),
+						"txSize", processedTx.Transaction.Size(), "remainingSize", remainingSize)
+					break
+				}
+				remainingSize -= processedTx.Transaction.Size()
```
HerbertJordan: I don't think we should `break` here. At this point, the damage is already done, the transaction was processed, and the break does not undo the damage. What the code should do here is

```suggestion
				if remainingSize < processedTx.Transaction.Size() {
					log.Debug("Block size limit exceeded,", "tx", processedTx.Transaction.Hash().Hex(),
						"txSize", processedTx.Transaction.Size(), "remainingSize", remainingSize)
					remainingSize = 0
				} else {
 					remainingSize -= processedTx.Transaction.Size()
 				}
```

Wdyt?

Alternatively, the loop should create an inter-transaction-snapshot before running the transaction and roll back to that if the limit is exceeded.

evmcore/state_processor_test.go
```diff
@@ -3808,12 +3923,392 @@ func TestRunTransaction_Bundle_ReturnsExecutionCostFromBundleExecution(t *testin
...
+			got, status := runner.runSponsoredTransaction(context, tx, 0, tc.sizeLimit)
+			require.Equal(t, core_types.TransactionResultInvalid, status)
+			require.Len(t, got, 1)
+			require.Equal(t, tx, got[0].Transaction)
+			require.Nil(t, got[0].Receipt)
+		})
+	}
+}
+
+func TestBundleTransactionRunner_Run_DeductsSizeLimit(t *testing.T) {
```
HerbertJordan: A test checking that the size limit is not reduced if the receipt of a transaction is nil should be added.

evmcore/state_processor_test.go
```diff
@@ -3808,12 +3923,392 @@ func TestRunTransaction_Bundle_ReturnsExecutionCostFromBundleExecution(t *testin
...
+		remainingSize uint64
+		setupMock     func(*runContext, *Mock_transactionRunner)
+		wantProcessed int
+		wantIncluded  int
+	}{
+		"single transaction": {
+			txs:           []*types.Transaction{regular1},
+			remainingSize: regular1.Size(),
+			setupMock: func(ctx *runContext, runner *Mock_transactionRunner) {
+				runner.EXPECT().runRegularTransaction(ctx, regular1, 0, gomock.Any()).Return(
```
HerbertJordan: A test that the runXXXTransaction functions are called with the correct size limits, decreasing for every accepted transaction, staying the same for every skipped transaction, should be added.

HerbertJordan (commented):
Some initial feedback, but I want to pass over it another time tomorrow morning, with fresh eyes.

evmcore/state_processor.go
simonlechner: Yes right, all processed transactions need to be added to the summary.

evmcore/state_processor_test.go
```diff
@@ -3808,12 +3923,427 @@ func TestRunTransaction_Bundle_ReturnsExecutionCostFromBundleExecution(t *testin
...
+	require.Equal(t, initialSizeLimit-tx1.Size()-tx2.Size(), bundleRunner.sizeLimit)
+}
+
+func TestBundleTransactionRunner_Run_RevertsWhenSizeLimitExceeded(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	runner := NewMock_transactionRunner(ctrl)
+
+	stateDb := state.NewMockStateDB(ctrl)
+	stateDb.EXPECT().InterTxSnapshot().Return(1).AnyTimes()
+	stateDb.EXPECT().RevertToInterTxSnapshot(1).AnyTimes()
```
HerbertJordan: Since the revert-calls are the target of this test, can you strengthen the requirements here; in particular, check the number of invocations, the return values and arguments, and their ordering.

evmcore/state_processor.go
LorenzSchueler: What about `txs` -> `newProcessedTxs`, `processedTx` -> `newProcessedTx`?

gossip/c_block_callbacks.go
```diff
@@ -609,84 +604,49 @@ func consensusCallbackBeginBlockFn(
...
-						processed.Receipt,
-					)
-					remainingGas -= processed.Receipt.GasUsed
-					remainingSize -= processed.Transaction.Size()
-					causedBy[processed.Transaction.Hash()] = tx.Hash()
-				}
+	remainingSize := uint64(math.MaxUint64)
+	if upgrades.Brio {
+		remainingSize = uint64(params.MaxBlockSize - rlpEncodedMaxHeaderSizeInBytes)
+		for _, tx := range blockBuilder.GetTransactions() {
```
HerbertJordan, line 627: No test is failing if I remove this for loop.

I guess `TestProcessUserTransactions_DeductsInternalTxsSize` was covering this, but it is not any more.

evmcore/state_processor.go
```diff
@@ -228,12 +231,21 @@ func runTransactions(
...
-		for _, tx := range txs {
-			if tx.Receipt != nil { // < only transactions included in the block
+		for _, processedTx := range txs {
+			if processedTx.Receipt != nil { // < only transactions included in the block
 				trueTxIndexOffset++
+				causedBy[processedTx.Transaction.Hash()] = tx.Hash()
+
+				if remainingSize < processedTx.Transaction.Size() {
+					log.Debug("Block size limit exceeded,", "tx", processedTx.Transaction.Hash().Hex(),
+						"txSize", processedTx.Transaction.Size(), "remainingSize", remainingSize)
```
LorenzSchueler, line 244: I hate passing around mutable state all over the place, but having unreachable code is also not very nice. Should we maybe pass a pointer to `remainingSize` to `runTransaction` so that is can be decreased there, and we don't have to do anything about the remaining size here?

LuisPH3 (dismissed):

gossip/c_block_callbacks_test.go
```diff
@@ -1191,12 +1188,21 @@ func TestProcessUserTransactions_DeductsInternalTxsSize(t *testing.T) {
...
+	upgrades := opera.Upgrades{Brio: true}
+	processUserTransactions(evmProcessor, blockBuilder, orderedTxs, 10000, upgrades)
+
+	// Only the internal tx should be present
 	gotTxs := blockBuilder.GetTransactions()
 	require.Equal(t, types.Transactions{internalTx}, gotTxs)
-	require.Equal(t, 1, skippedCount)
 }
 
 func TestProcessUserTransactions_SkipsTxsExceedingSizeLimit(t *testing.T) {
```
HerbertJordan, line 1208: I think this test could be copied for the runTransactions in the state processor. It makes sure that a large one is skipped, but smaller succeeding transactions are still processed.

HerbertJordan (changes requested):

LorenzSchueler (dismissed):

evmcore/state_processor.go
simonlechner: As discussed offline, this will just move the same issue a level deeper into the runner.

gossip/c_block_callbacks.go
```diff
@@ -609,84 +604,49 @@ func consensusCallbackBeginBlockFn(
...
+			txSize := tx.Size()
+			if txSize > remainingSize {
```
HerbertJordan, line 629: No test fails if I remove this condition.

HerbertJordan (dismissed):

LuisPH3 (approved):

PR #992, opened 2026-05-20
Test bundle store after network restart

LuisPH3:
This integration test checks the persistence of the bundles execution store. 

Bundles can be queried after a network restart.


LorenzSchueler (approved):
Looks Good.

PR #994, opened 2026-05-20
Push size checks into transaction runner

HerbertJordan:


evmcore/state_processor.go
```diff
@@ -545,16 +531,25 @@ type bundleTransactionRunner struct {
...
 	b.executionCost += execCost
 	b.processedTransactions = append(b.processedTransactions, processed...)
 
 	for _, p := range processed {
 		if p.Receipt != nil {
 			b.trueTxOffset++
+			if b.sizeLimit < p.Transaction.Size() {
+				b.RevertToSnapshot(snapshot)
+				return core_types.TransactionResultInvalid
+			}
```
HerbertJordan, line 551: This might be over-cautious ... we could assume that the `runTransaction` does its job and does not accept transactions beyond the limit.

HerbertJordan:
Obsolete, migrated to #988

PR #995, opened 2026-05-20
Add integration test to verify chainId check of bundled txs

LuisPH3:
In this test: both bundle envelopes and bundled transactions are signed with a different chain ID than the integration test network. These transactions must return errors when sent to the pool, and if forced into the consensus they shall be skipped and yield no results.

### DIFF
--- a/tests/bundles/chainid_test.go
+++ b/tests/bundles/chainid_test.go
@@ -1,3 +1,19 @@
+// Copyright 2026 Sonic Operations Ltd
+// This file is part of the Sonic Client
+//
+// Sonic is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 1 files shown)


LorenzSchueler (approved):
The test is fine, but there are two possible improvements:
- Use the new tooling to separate the intake and emission rejection into two separate cases each.
- Also have a case with the correct signature where the tx is processed so make sure the rest of the test setup is correct.

LuisPH3:
>     * Use the new tooling to separate the intake and emission rejection into two separate cases each.
It would be nice to have the code reusing tooling, but setup is slightly different and the test is more straight forward as it is. 

>     * Also have a case with the correct signature where the tx is processed so make sure the rest of the test setup is correct.
the setup is not unusual, this is tested by any other test checking bundles.

PR #996, opened 2026-05-20
Fix flacky behaviour in TestLogBloom_query

LuisPH3:
If the network produced a block without transactions, the bloom is not correct,

Reported output: 
```
08:30:51  --- FAIL: TestLogBloom_query (0.94s)
08:30:51      log_subscription_test.go:108: expected non-empty bloom in block 280, got empty
08:30:51  FAIL
08:30:51  FAIL	github.com/0xsoniclabs/sonic/tests	1399.975s
``` 

This pr checks blooms during processing, only if the block contains txs. 
Additionally replaces the busy loop with a cancelable wait call.

### DIFF
--- a/tests/log_subscription_test.go
+++ b/tests/log_subscription_test.go
@@ -104,8 +104,14 @@ func TestLogBloom_query(t *testing.T) {
 			}
 			require.NoError(err)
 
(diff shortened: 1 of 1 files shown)


jenikd (approved):
This is better approach with waiting for a block. Thanks

PR #998, opened 2026-05-21
Remove a must-to-uint256 from transaction utils

HerbertJordan:
This PR eliminates a potential panic from the transaction utils by replacing it with explicit error handling.

The utility is used in the bundle validation, and is thus part of critical code.  This update removes the panic from the code.

### DIFF
--- a/utils/txs.go
+++ b/utils/txs.go
@@ -17,19 +17,27 @@
 package utils
 
 import (
+	"errors"
 	"fmt"
 	"math/big"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/holiman/uint256"
 )
 
 // GetTxData extracts the inner TxData from a given transaction, which is
 // handy for mutating transactions in various contexts.
-func GetTxData(tx *types.Transaction) types.TxData {
-
+func GetTxData(tx *types.Transaction) (types.TxData, error) {
 	// TODO: consider adding a modification to Sonic's go-ethereum fork to
 	// enable a direct call to tx.inner.copy(), having the same effect.
+	if tx == nil {
+		return nil, fmt.Errorf("transaction is nil")
+	}
+	return getTxDataInternal(tx)
+}
+
+func getTxDataInternal(tx txDataSource) (types.TxData, error) {
 
 	// Manually create a copy of the transactions's inner data type.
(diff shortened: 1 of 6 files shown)


utils/txs.go
```diff
@@ -77,53 +85,111 @@ func GetTxData(tx *types.Transaction) types.TxData {
...
 	case types.BlobTxType:
+		chainId, err1 := toUint256(tx.ChainId())
+		gasTipCap, err2 := toUint256(tx.GasTipCap())
+		gasFeeCap, err3 := toUint256(tx.GasFeeCap())
+		value, err4 := toUint256(tx.Value())
+		blobFeeCap, err5 := toUint256(tx.BlobGasFeeCap())
+		v, err6 := toUint256(v)
+		r, err7 := toUint256(r)
+		s, err8 := toUint256(s)
+
```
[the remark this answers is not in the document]
HerbertJordan: It obviously compiles. The variables are new local definitions.

LuisPH3 (approved):

utils/txs.go
```diff
@@ -17,19 +17,27 @@
...
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/holiman/uint256"
 )
 
 // GetTxData extracts the inner TxData from a given transaction, which is
 // handy for mutating transactions in various contexts.
-func GetTxData(tx *types.Transaction) types.TxData {
-
+func GetTxData(tx *types.Transaction) (types.TxData, error) {
```
LuisPH3, line 31: I recently had a situation with this function,  when used to create bundles containing txs signed by a different chain id. 

The problem was that an already signed tx cannot be signed again with different tx. it just panics. The source of this problem was this function copying the signature. 
This function would be more intuitive if the signature is not copied, after all we are unpacking a closed tx to make it mutable, therefore invalidating the signature. 

I propose to rename this function to `GetUnsignedTxData` and avoid copying the signature fields and chainId.

HerbertJordan: That may be an independent follow-up task. 

The so far, the idea of this function was to extract the inner data truthfully, which could be a one-liner if geth would offer support for it (see [here](https://github.com/0xsoniclabs/sonic/blob/f86ef94249461d5776e62cb8198892c8fd77eb00/utils/txs.go#L31-L32)).

PR #999, opened 2026-05-21
Add monitoring metrics to bundle pre-checks

LuisPH3:
This metric can be used to localize load issues in the pool-emitter as they share this utility.
These metrics do not provide overview of the bundles used submitted to the network, but instead it can be used to monitor the overhead of processing these bundles in the frontend. 

Bundles pre-check is shared among TX validation on intake, pool promotion and demotion, and the emitter. Values reported by these metrics can be correlated to CPU usage to diagnose the client. 

These metrics are normal counters and can be integrated into norma without further development.

### DIFF
--- a/utils/metrics.go
+++ b/utils/metrics.go
@@ -45,3 +45,9 @@ func NewPrometheusHistogram(opts prometheus.HistogramOpts) *PrometheusHistogram 
 func (h *PrometheusHistogram) Update(v float64) {
 	h.histogram.Observe(v)
 }
+
+// MetricsCounterWrapper is an interface that wraps the methods of a
+// prometheus Counter to facilitate testing with mocks.
+type MetricsCounterWrapper interface {
+	Inc(int64)
+}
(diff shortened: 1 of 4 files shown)


utils/metrics.go
```diff
@@ -45,3 +45,9 @@ func NewPrometheusHistogram(opts prometheus.HistogramOpts) *PrometheusHistogram
 func (h *PrometheusHistogram) Update(v float64) {
 	h.histogram.Observe(v)
 }
+
+// MetricsCounterWrapper is an interface that wraps the methods of a
+// prometheus Counter to facilitate testing with mocks.
+type MetricsCounterWrapper interface {
```
HerbertJordan: nit: not a fan of the "Wrapper" suffix ... it is an interface, not a wrapper, and should have no suffix. Wdyt?

HerbertJordan (dismissed):

HerbertJordan (dismissed):

simonlechner (approved):

PR #1003, opened 2026-05-21
Sponsorship and Bundle metrics

simonlechner:
This PR adds counters for sponsored transactions and bundles as well as a histogram for effective gas of a bundle.
The summary execution gas had to be changed to hold information about the execution cost per transaction.

### DIFF
--- a/evmcore/bundle_precheck_test.go
+++ b/evmcore/bundle_precheck_test.go
@@ -1495,7 +1495,7 @@ func Test_trialRunBundleInternal_IncrementsMetrics(t *testing.T) {
 	).Build()
 
 	processedTx := ProcessedTransaction{} // tx after subgroup succeeds
-	expectedExecCost := core_types.ExecutionCost(50_000)
+	expectedExecCost := map[common.Hash]core_types.ExecutionCost{{0x01}: 50_000}
 
 	chainState := NewMockChainStateForBundleEval(ctrl)
 	chainState.EXPECT().GetLatestHeader().Return(&EvmHeader{
@@ -1520,7 +1520,7 @@ func Test_trialRunBundleInternal_IncrementsMetrics(t *testing.T) {
 	gasMock := utils.NewMockMetricsCounter(ctrl)
 
 	countMock.EXPECT().Inc(int64(1))
-	gasMock.EXPECT().Inc(int64(expectedExecCost))
+	gasMock.EXPECT().Inc(int64(expectedExecCost[common.Hash{0x01}]))
 
 	trialRunBundleInternal(
 		envelope, chainState, db, factory, rand.Read,
--- a/evmcore/bundle_precheck_test.go
+++ b/evmcore/bundle_precheck_test.go
@@ -723,7 +723,7 @@ func Test_trialRunBundleInternal_RejectsBundlesWhereEfficiencyIsBelowThreshold(t
 			processor.EXPECT().Run(any, any).Return(
 				ProcessSummary{
 					ProcessedTransactions: tc.processedTxs,
-					ExecutionCost:         tc.execCost,
+					ExecutionCosts:        tc.execCost,
 				},
 			)
 
@@ -1077,7 +1077,7 @@ func Test_trialRunBundleInternal_UsesPresentsOfReceiptToDecideResult(t *testing.
 			processor := NewMocktransactionProcessor(ctrl)
 			processor.EXPECT().Run(any, any).Return(ProcessSummary{
 				ProcessedTransactions: tc.processedTxs,
-				ExecutionCost:         map[common.Hash]core_types.ExecutionCost{{}: 1},
+				ExecutionCosts:        map[common.Hash]core_types.ExecutionCost{{}: 1},
 			})
 
 			factory := NewMocktransactionProcessorFactory(ctrl)
@@ -1405,7 +1405,7 @@ func Test_trialRunBundleInternal_ReturnsCorrectGasEfficiency(t *testing.T) {
 			processor.EXPECT().Run(any, any).Return(
 				ProcessSummary{
 					ProcessedTransactions: tc.processedTxs,
-					ExecutionCost:         tc.execCost,
+					ExecutionCosts:        tc.execCost,
 				},
 			)
 
@@ -1506,7 +1506,7 @@ func Test_trialRunBundleInternal_IncrementsMetrics(t *testing.T) {
 	processor := NewMocktransactionProcessor(ctrl)
 	processor.EXPECT().Run(any, any).Return(ProcessSummary{
 		ProcessedTransactions: []ProcessedTransaction{processedTx},
-		ExecutionCost:         expectedExecCost,
+		ExecutionCosts:        expectedExecCost,
 	})
 
 	factory := NewMocktransactionProcessorFactory(ctrl)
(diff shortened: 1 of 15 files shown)


gossip/emitter/scheduler/processor.go
```diff
@@ -99,8 +100,13 @@ func (p *evmProcessor) run(tx *types.Transaction) (bool, uint64) {
 		}
 	}
 
-	if summary.ExecutionCost == 0 ||
-		float64(gasUsed)/float64(summary.ExecutionCost) < evmcore.MinBundleEfficiency {
+	totalExecCost := core_types.ExecutionCost(0)
+	for _, cost := range summary.ExecutionCost {
+		totalExecCost += cost
+	}
```
LorenzSchueler: nit:
Since we sum them up in many places, at method `TotalExecutionCost` might be worth it.

gossip/c_block_callbacks.go
```diff
@@ -77,6 +78,17 @@ var (
...
 	spilledEventsMeter   = metrics.GetOrRegisterMeter("chain/events/spilled", nil)   // tx excluded because of MaxBlockGas
+
+	sponsoredCounter        = utils.MetricsCounter(metrics.GetOrRegisterCounter("chain/sponsored", nil))
+	skippedSponsoredCounter = utils.MetricsCounter(metrics.GetOrRegisterCounter("chain/sponsored/skipped", nil))
+	bundleCounter           = utils.MetricsCounter(metrics.GetOrRegisterCounter("chain/bundles", nil))
+	rolledBackBundleCounter = utils.MetricsCounter(metrics.GetOrRegisterCounter("chain/bundles/rolledBack", nil))
+
+	effectiveBundleGasHistogram = utils.MetricsHistogramWrapper(utils.NewPrometheusHistogram(prometheus.HistogramOpts{
+		Name:    "chain_bundle_gas_effective",
+		Help:    "Effective gas usage percentage for a bundle transactions",
```
LorenzSchueler:
```suggestion
		Name:    "chain_bundle_gas_efficiency",
		Help:    "Gas usage efficiency percentage for bundle transactions",
```

evmcore/state_processor.go
```diff
@@ -49,18 +50,22 @@ type StateProcessor struct {
 	bc        DummyChain          // Canonical block chain
 	upgrades  opera.Upgrades      // Enabled network upgrades
 	forReplay bool                // Whether the state processor is used for replaying transactions or for head state processing.
+
+	rolledBackBundleCounter utils.MetricsCounter
```
LorenzSchueler, line 54: Why not pass all 4 counters, so that they are all updated here and not some is the c block callbacks and some in the state processor?

evmcore/state_processor.go
```diff
@@ -84,9 +89,29 @@ func NewStateProcessorForReplay(
...
+
+	// ExecutionCost contains the raw execution cost for each transaction,
+	// this differs from gas used as it also includes the gas from rolled-back
+	// bundles.
+	// This is a map from the hash of the transaction that triggered execution to
+	// its execution cost.
+	// - For regular transactions and sponsored transactions, the key is the
+	// hash of the transaction itself.
+	// - For bundles, the key is the hash of the envelope transaction of the bundle.
+	ExecutionCost map[common.Hash]core_types.ExecutionCost
```
LorenzSchueler: nit:

```suggestion
	ExecutionCosts map[common.Hash]core_types.ExecutionCost
```

LorenzSchueler (dismissed):
The changes are fine but I have 2 concerns:
- The ratio reported in the histogram depends on 2 counters. Updating one of them in the c block callbacks and one in the state processor is not ideal because they might get out of sync more easily when new changes are introduced.
- we report the bundle efficiency, but this efficiency is not weighted by total cost. Therefore, if we have block1: 1000Gas 0.2Efficiency and block2: 1Gas 1.0Efficiency it would look like a 0.6Efficiency, although it is almost 0.2.

evmcore/state_processor.go
simonlechner: There are already lots of other metrics in the c block callbacks, this is the only one that requires special handling.

evmcore/bundle_precheck_test.go
```diff
@@ -745,13 +745,17 @@ func Test_trialRunBundleInternal_RejectsBundlesWhereEfficiencyIsBelowThreshold(t
 			)
 			require.Equal(t, tc.expectAccept, valid)
 
+			totalExecCost := core_types.ExecutionCost(0)
+			for _, cost := range tc.execCost {
+				totalExecCost += cost
+			}
```
LorenzSchueler, line 751: nit: use the `TotalExecutionCost` method

gossip/c_block_callbacks.go
```diff
@@ -77,6 +79,17 @@ var (
...
 	spilledEventsMeter   = metrics.GetOrRegisterMeter("chain/events/spilled", nil)   // tx excluded because of MaxBlockGas
+
+	sponsoredCounter        = utils.MetricsCounter(metrics.GetOrRegisterCounter("chain/sponsored", nil))
+	skippedSponsoredCounter = utils.MetricsCounter(metrics.GetOrRegisterCounter("chain/sponsored/skipped", nil))
+	bundleCounter           = utils.MetricsCounter(metrics.GetOrRegisterCounter("chain/bundles", nil))
+	rolledBackBundleCounter = utils.MetricsCounter(metrics.GetOrRegisterCounter("chain/bundles/rolledBack", nil))
+
+	effectiveBundleGasHistogram = utils.MetricsHistogramWrapper(utils.NewPrometheusHistogram(prometheus.HistogramOpts{
+		Name:    "chain_bundle_gas_effective",
+		Help:    "Effective gas usage ratio for a bundle transactions",
```
LorenzSchueler:
```suggestion
		Help:    "Effective gas usage ratio for a bundle transaction",
```

LorenzSchueler (dismissed):

evmcore/bundle_precheck_test.go
simonlechner: This is not a summary type.

evmcore/state_processor.go
```diff
@@ -502,6 +538,12 @@ func (r *transactionRunner) runTransactionBundleInternal(
 		sizeLimit:    sizeLimit,
 	}
 	if !bundle.RunBundle(txBundle, &runner) {
+
+		// Track unsuccessful bundle execution in metrics.
+		if ctxt.rolledBackBundleCounter != nil {
+			ctxt.rolledBackBundleCounter.Inc(1)
```
HerbertJordan, line 544: This will also count nested bundles being rolled back, not just top-level bundles.

gossip/c_block_callbacks.go
```diff
@@ -646,9 +660,60 @@ func processUserTransactions(
 		}
 	}
 
+	updateTransactionMetrics(orderedTxs, summary, upgrades, sponsoredCounter, skippedSponsoredCounter, bundleCounter)
+
 	return summary.CausedBy
 }
 
+func updateTransactionMetrics(inputTransactions []*types.Transaction, summary evmcore.ProcessSummary, upgrades opera.Upgrades,
```
HerbertJordan, line 668: It might be more maintainable if we push the passing of monitoring metrics into be state processor, e.g. through a struct
```
type Metrics struct {
    BundleGasHistogram Metric
    SkippedSponsoredTxCounter Metric
}
```
and call those metrics at the place where those values are actually produced, instead of encoding all of these values into the `ProcessingSummary` and then try to figure out what has happened after the fact.

simonlechner:
Superseded by #1006.

PR #1005, opened 2026-05-22
Disable acceptance of temporary blocked transactions in TxPool

HerbertJordan:
This PR disables bundles from being accepted when they can not be immediately processed on the chain.


LuisPH3 (commented):
We have to check norma, it could be that this change invalidates the current work generators. 
other than that looks good.

LuisPH3 (approved):

PR #1006, opened 2026-05-22
Add sonic features metrics to block processor

LuisPH3:
This is an alternative implementation to #1003,

The main idea is to allocate monitoring tools in c_block_callbacks and forward them into the state processor to be used, this allows to account events on location instead of collecting data and returning it. 

Notes: 

- Only top level bundles if nested is accounted for, including efficiency. IMO this makes sense as this is the one transaction unit submitted by the user. 
- Skipped sponsorships and executed are exclusive, sum is the total number of sponsorships accepted into the processor.


evmcore/block_metrics.go
```diff
@@ -0,0 +1,67 @@
...
+		m.ExecutedBundles.Inc(int64(1))
+	}
+}
+func (m *BlockExecutionMetrics) IncRolledBackBundle() {
+	if m.RolledBackBundles != nil {
+		m.RolledBackBundles.Inc(int64(1))
+	}
+}
+
+func (m *BlockExecutionMetrics) UpdateBundleEfficiencyHistogram(usedGas, totalGas uint64) {
```
HerbertJordan: Nit: `Update` implies something is replaced with a new value; the correct terminology for histograms would be [Observe](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Observer);

I guess this should also be updated in the `utils.MetricsHistogramWrapper`, which should be renamed to `utils.MetricsHistogram`; 

If you think it would be also the right time for it, it could also be a `metrics.Histogram` in the `utils/metrics` directory.

evmcore/state_processor.go
```diff
@@ -246,14 +269,38 @@ func runTransactions(
...
+				gasUsed += processedTx.Receipt.GasUsed
+			}
+
+			if subsidies.IsSponsorshipRequest(processedTx.Transaction) {
+				switch processedTx.Receipt {
+				case nil:
+					context.metrics.IncSkippedSponsoredTx()
+				default:
+					context.metrics.IncSponsoredTx()
+				}
```
HerbertJordan:
```suggestion
				if processedTx.Receipt == nil {
					context.metrics.IncSkippedSponsoredTx()
				} else {
					context.metrics.IncSponsoredTx()
				}
```

evmcore/block_metrics.go
```diff
@@ -0,0 +1,67 @@
...
+	if m.BundleEfficiency != nil && totalGas > 0 {
+		efficiency := float64(usedGas) / float64(totalGas)
+		m.BundleEfficiency.Update(efficiency)
+	}
+}
+
+// NoOpBlockExecutionMetrics ignores metrics collection, used by tools running
+// block processor functions outside of the context of a block processor
+// (e.g. pre-check, single proposer executor)
+var NoOpBlockExecutionMetrics BlockExecutionMetrics
```
HerbertJordan: nit: be declarative, not implementation detail specific; `NoOpBlockExecutionMetrics` => `NoBlockExecutionMetrics`

evmcore/state_processor.go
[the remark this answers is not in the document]
LuisPH3: upgrade guards added

evmcore/state_processor.go
HerbertJordan, line 1: Idea: using a interface matching the signature of the `BlockExecutionMetrics` in this file would make testing simpler; mocks could mock the `IncSPonsoredTx` and `UpdateBundleEfficiencyHistogram` directly, instead of having to understand the implementation details of `BlockExecutionMetrics`;

HerbertJordan (commented):
First thoughts, will continue reviewing after the stand-up.

evmcore/state_processor.go
LuisPH3: That was my first version, I discarded it because of two reasons:
- it forced me to check for nil on call as interfaces can be nil, the object is now passed by value and takes care of its own nill-ness. Having to check if an object is nil every time pollutes the code quite a lot. 
- Testing is more robust, as I do not only test that the method is called but that the right counter is incremented. For example the code would require dedicated testing:
```
func (s *SonicFeaturesMetrics) UpdateBundleEfficiencyHistogram(usedGas uint64, totalGas uint64) {
	if totalGas > 0 {
		efficiency := float64(usedGas) / float64(totalGas)
		s.BundleEfficiency.Update(efficiency)
	}
}
```
IMO adding the interface only adds boilerplate, because I have to initialize it anyway with the 5 variables.

evmcore/state_processor_test.go
```diff
@@ -4431,3 +4437,324 @@ func TestRunTransactions_RemainingSizeIsSetToZeroWhenProcessedTxExceedsIt(t *tes
...
+			[]ProcessedTransaction{{Transaction: txs[1], Receipt: &types.Receipt{GasUsed: 200}}},
+			core_types.TransactionResultSuccessful,
+			core_types.ExecutionCost(200),
+		)
+		executedBundles.EXPECT().Inc(int64(1)).Times(2)
+
+		runTransactions(context, txs, 0, math.MaxUint64)
+	})
+
+	t.Run("rolled back bundles are counted when yielding zero txs", func(t *testing.T) {
```
HerbertJordan: The property indicated by this test name is actually not satisfied by the implementation. The implementation does not use the number of accepted transactions as the signal for a rolled-back transaction, but the execution result being "TransactionResultFailed";

Please clarify whether the test of the implementation is what we want.

evmcore/state_processor_test.go
```diff
@@ -4431,3 +4437,324 @@ func TestRunTransactions_RemainingSizeIsSetToZeroWhenProcessedTxExceedsIt(t *tes
...
+		}
+
+		txs := []*types.Transaction{getTransactionBundle(t)}
+
+		runner.EXPECT().runTransactionBundle(context, txs[0], 0, gomock.Any()).Return(
+			[]ProcessedTransaction{{Transaction: txs[0], Receipt: nil}},
+			core_types.TransactionResultFailed,
+			core_types.ExecutionCost(0),
+		)
+		// histogram.Update must NOT be called (division by zero guard)
```
HerbertJordan: From a unit test perspective, this should be a test targeting the BlockExecutionMetrics' `ObserveBundleEfficiency` function. The guard is not part of the `runTransactions` function.

HerbertJordan (dismissed):

LuisPH3 (commented):
👍

evmcore/block_metrics.go
```diff
@@ -0,0 +1,87 @@
...
+
+// NewBlockExecutionMetrics creates a new BlockExecutionMetrics instance with
+// the given metric counters and histogram.
+func NewBlockExecutionMetrics(
+	sponsoredTxs utils.MetricsCounter,
+	skippedSponsoredTxs utils.MetricsCounter,
+	executedBundles utils.MetricsCounter,
+	rolledBackBundles utils.MetricsCounter,
+	bundleEfficiency utils.MetricsHistogram,
+) BlockExecutionMetrics {
```
HerbertJordan: nit: Luis' golden rule - accept interfaces, return concrete types

```suggestion
) *defaultBlockExecutionMetrics {
```

evmcore/block_metrics.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+type defaultBlockExecutionMetrics struct {
+	sponsoredTxs        utils.MetricsCounter
+	skippedSponsoredTxs utils.MetricsCounter
+	executedBundles     utils.MetricsCounter
+	rolledBackBundles   utils.MetricsCounter
+	bundleEfficiency    utils.MetricsHistogram
+}
```
HerbertJordan: In such a case, I think it is more readable if you make the struct and the fields public and just let the user create the struct as needed. This results in:
- named fields during initialization, so you can see which metric gets assigned to which instance; the order dependent factory function above, using the same type for most fields, would not make it easy to detect if arguments get mixed up
- in this specific case - even ignoring a few metrics is a valid setup, since all `nil` metrics are valid; there are no invariants to protect.

evmcore/state_processor.go
```diff
@@ -246,14 +269,37 @@ func runTransactions(
...
 		}
 		processedTxs = append(processedTxs, txs...)
 
+		if context.upgrades.Brio && bundle.IsEnvelope(tx) && context.metrics != nil {
+			// update metrics for bundles
+			switch txResult {
+			case core_types.TransactionResultSuccessful:
+				context.metrics.IncExecutedBundle()
+			case core_types.TransactionResultFailed:
+				context.metrics.IncRolledBackBundle()
```
HerbertJordan, line 292: nit: This switch basically calls for another metric tracking invalid bundles ..

evmcore/state_processor.go
```diff
@@ -758,7 +809,7 @@ func (tp *TransactionProcessor) Run(i int, tx *types.Transaction) ProcessSummary
 	return runTransactions(newRunContext(
 		tp.signer, tp.header.BaseFee, tp.stateDb, tp.gp, tp.blockNumber, tp.blockTime,
 		&tp.usedGas, tp.onNewLog, tp.upgrades, &transactionRunner{evm{tp.vmEnvironment}},
-		false,
+		false, nil,
```
HerbertJordan: I am surprised to see this working ... are we never running any sponsored transactions or bundles through this function?

HerbertJordan (dismissed):

facuMH (approved):

PR #1009, opened 2026-05-25
Fix pre-Brio fee attribution for subsidized transactions

thaarok:
Fixes incompatibility of the main branch with v2.1.6 discovered in Norma testing.

  After #988 merged processUserTransactionsNoLimits into the unified
  processUserTransactions, the causedBy map began propagating in non-Brio
  mode. This caused payment transactions generated by GasSubsidies
  sponsorships to be attributed to the originating validator, whereas the
  old code left them unattributed (txCausedBy was nil).

### DIFF
--- a/gossip/c_block_callbacks.go
+++ b/gossip/c_block_callbacks.go
@@ -646,6 +646,10 @@ func processUserTransactions(
 		}
 	}
 
+	// Pre-Brio: return nil to preserve the pre-Brio fee attribution behavior.
+	if !upgrades.Brio {
+		return nil
+	}
 	return summary.CausedBy
 }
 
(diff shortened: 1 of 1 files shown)


gossip/c_block_callbacks.go
```diff
@@ -487,7 +487,7 @@ func consensusCallbackBeginBlockFn(
 					// call OnNewReceipt
 					for i, r := range allReceipts {
 						originTx := r.TxHash
-						if origin, found := txCausedBy[r.TxHash]; found {
+						if origin, found := txCausedBy[r.TxHash]; found && es.Rules.Upgrades.Brio {
 							originTx = origin
 						}
```
HerbertJordan: nit: for readability:

```suggestion
						if es.Rules.Upgrades.Brio {
							if origin, found := txCausedBy[r.TxHash]; found {
								originTx = origin
							}
						}
```

HerbertJordan (dismissed):
Thanks for identifying the issue and adding the fix. I would appreciate if we could also get a unit test with the Sonic repository as a regression test.

facuMH (dismissed):
This PR adds a filter so that `txCauseBy` is only used when Brio is enabled.

LuisPH3:
Same fix with test #1011

PR #1013, opened 2026-05-26
Add own `HexToBytes` and `BytesToHex`

facuMH:
Implements https://github.com/0xsoniclabs/sonic-admin/issues/765

Introduces a dedicated `hexutils` package under `utils` providing two utility functions for bytes and strings conversions, and migrates existing usages across the codebase to use them.
- `HexToBytes`: decodes a hex string to bytes, stripping spaces between byte pairs; calls `log.Fatal` on invalid input.
- `BytesToHex`: encodes a byte slice to an uppercase hex string.

### DIFF
--- a/utils/hexutils/hexutils.go
+++ b/utils/hexutils/hexutils.go
@@ -20,13 +20,13 @@ import (
 	"encoding/hex"
 	"fmt"
 	"log"
-	"regexp"
+	"strings"
 )
 
 // HexToBytes converts a hex string to a byte sequence.
 // The hex string can have spaces between bytes.
 func HexToBytes(s string) []byte {
-	s = regexp.MustCompile(" ").ReplaceAllString(s, "")
+	s = strings.ReplaceAll(s, " ", "")
 	b := make([]byte, hex.DecodedLen(len(s)))
 	_, err := hex.Decode(b, []byte(s))
 	if err != nil {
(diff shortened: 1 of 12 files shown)


utils/hexutils/hexutils.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// HexToBytes converts a hex string to a byte sequence.
+// The hex string can have spaces between bytes.
+func HexToBytes(s string) []byte {
+	s = strings.ReplaceAll(s, " ", "")
+	b := make([]byte, hex.DecodedLen(len(s)))
+	_, err := hex.Decode(b, []byte(s))
+	if err != nil {
+		log.Fatal(err)
```
LuisPH3: This is unfortunately, not enough.

A successful solution for this problem would a sensible error propagation. Logging is just not very useful, and errors may take days util identified. 

A successful solution for this problem requires a study of the contexts where this function is used and sensible error propagation if needed.

LuisPH3 (changes requested):

utils/hexutils/hexutils.go
facuMH: All uses of `HexToBytes` are in declaration of global variables with either constant strings or embedded string, so error propagation via `([]byte, error)` would get lost. Instead I think using the `Must` pattern and panicking in cases of invalid string is more sensible approach. Look at https://github.com/0xsoniclabs/sonic/pull/1013/commits/26e663d8eab223f8316e5bcfd1363af80c9b964a

LuisPH3:
we wont be needing this now

PR #1015, opened 2026-05-27
Add network sponsorship modes to subsidies package

HerbertJordan:
This PR introduces support for two new sponsorship modes:
- network-sponsored transactions ... where transaction costs are absorbed by the network
- network-sponsored transactions with tracking ... where transaction costs are absorbed by the network, but tracking data for an automated on-chain traffic control may be implemented.

The change includes:
- updates to the subsidies registry interface, in particular
   - an update of the return values delivered by the `chooseFund` function, which now returns a sponsorship mode and an optional ID
   - the addition of a `track` function, for delivering tracking calls if requested
   - the extension of the `getGasConfig` function's list of return values to also include the gas limit for the tracking calls

**For Reviewers**
To enable the new processing model, the PR introduces a new `Sponsorship` struct returned from the `IsCovered` call. This struct replaces a list of individual fields returned before with a type abstracting all internal details.

To keep the scope of this PR limited to the subsidies package, the public interface is kept stable. However, in a follow-up PR #1016, the old legacy signatures are going to be removed and replaced with the new signatures.


HerbertJordan:
Replaced by #1017

PR #1016, opened 2026-05-27
Migrate callers to Sponsorship API and remove backward-compat shims

HerbertJordan:
This PR updates all code in the Sonic client interacting with the subsidies module to use the new `Sponsorship` object based API and removes the legacy API.


HerbertJordan:
Replaced by #1017

PR #1017, opened 2026-05-27
Add network sponsorship modes to subsidies

HerbertJordan:
This PR introduces support for two new sponsorship modes:
- network-sponsored transactions ... where transaction costs are absorbed by the network
- network-sponsored transactions with tracking ... where transaction costs are absorbed by the network, but tracking data for an automated on-chain traffic control may be implemented.

The change includes:
- updates to the subsidies registry interface, in particular
   - an update of the return values delivered by the `chooseFund` function, which now returns a sponsorship mode and an optional ID, either referencing a fund-ID, a tracking-ID, or being ignored
   - the addition of a `track` function, for delivering tracking calls if requested
   - the extension of the `getGasConfig` function's list of return values to also include the gas limit for the tracking calls
- the introduction of a new `Sponsorship` type abstracting the implementation details and providing everything needed by other packages for dealing with sponsorships; the instance is created by the `IsCovered` call and can be queried for everything sponsorship related during processing

Integration tests are going to be added in a follow-up PR.

**For Reviewers**
The core contributions of this change are limited to three files, which should be studied in the given order:
- the updated interface of the subsidies registry  [here](https://github.com/0xsoniclabs/sonic/blob/herbert/extended_sponsorship_update/gossip/blockproc/subsidies/registry/subsidies_registry.sol)
- the update of the subsidies utility library API [here](https://github.com/0xsoniclabs/sonic/blob/herbert/extended_sponsorship_update/gossip/blockproc/subsidies/subsidies.go)
- the update of the processing of sponsored transactions in the state processor [here](https://github.com/0xsoniclabs/sonic/blob/herbert/extended_sponsorship_update/evmcore/state_processor.go)

The rest of the PR is generated code, tests, documentation, and updated tests.

Overall, the following review order is suggested:
- start by studying the changes in the registry code [here](https://github.com/0xsoniclabs/sonic/blob/herbert/extended_sponsorship_update/gossip/blockproc/subsidies/registry/subsidies_registry.sol); if you like, you can also read the feature description [here](https://github.com/0xsoniclabs/sonic/blob/herbert/extended_sponsorship_update/gossip/blockproc/subsidies/network_sponsorships.md)
- review the changes in subsidies.go
- review the changes in the state_processor.go
- review the remaining changes to code files (not tests)
- check the tests added for the changes of production code
- check the adaptations to existing tests (which all the rest should be)

There will be a follow-up PR with integration tests.


evmcore/state_processor_test.go
```diff
@@ -2109,18 +2109,20 @@ func TestRunSponsoredTransaction_CoveredTransaction_ProcessesTwoTransactionsSucc
 	callData := processedTransactions[1].Transaction.Data()
 	require.Equal(4+2*32, len(callData)) // chooseFund + deductFees
```
LuisPH3, line 2388: test `TestRunSponsoredTransaction_CoveredTransaction_ProcessesTwoTransactionsSuccessfully` could be renamed.

I think we should add a test which checks post-txs being zero, one, or more txs. I guess these txs could be of any form.

evmcore/state_processor_test.go
LuisPH3, line 1: this file contains a family of tests where funds deduction is checked for failed, skipped txs.
- TestRunSponsoredTransaction_FeeDeductionTxFails_TransactionIsAcceptedWithoutFeeDeduction
- TestRunSponsoredTransaction_FeeDeductionTxIsSkipped_TransactionIsAcceptedWithoutFeeDeduction
- TestRunSponsoredTransaction_FailingCreationOfFeeDeduction_TransactionIsAcceptedWithoutFeeDeductio

There is a new expected scenario where fees do not change, it should be added to the list:
- TestRunSponsoredTransaction_TxIsNetworkSponsored_TransactionIsAcceptedWithoutFeeDeduction

tests/gas_subsidies/insufficient_funds_test.go
```diff
@@ -53,16 +53,19 @@ func TestGasSubsidies_RequestIsRejectedInCaseOfInsufficientFunds(t *testing.T) {
 	require.NoError(t, err)
 
 	// The cost of the sponsored transaction is the gas used by the tx
-	// plus the overhead of the sponsorship itself
-	cost := tx.Gas + gasConfig.OverheadCharge.Uint64()
+	// plus the maximum overhead charged, depending on the sponsorship type.
+	limit := tx.Gas + max(
+		gasConfig.OverheadChargeForFundBackedSponsorships.Uint64(),
+		gasConfig.OverheadChargeForNetworkSponsorshipsWithTracking.Uint64(),
+	)
```
LuisPH3, line 60: nit: test seems to use OverheadChargeForFundBackedSponsorships

LuisPH3 (dismissed):
This PR extends the subsidies feature to allow for network-sponsored transactions with or without a tailing tracking call to the contract.

The changes are focused on the rewrite of the subsidies' module. 
- add new signatures, 
- refactor tooling to abstract details
- contemplate backwards compatibility

The integration happens in the state processor, where a simple rewrite allows using the new abstraction of tailing internal transactions instead of a single payment tx. 
Testing is adequate, but multiples are allowed, although not really needed, this is not tested. 

Additionally, a new builder tool to create subsidies tailing txs is added, and tests are updated. 

This PR could use some integration tests, but that can be added as a follow-up PR.

tests/gas_subsidies/insufficient_funds_test.go
HerbertJordan: The `chooseFund` function needs to be called with an upper bound of the fees that are going to be charged for a transaction. This limit is computed [here](https://github.com/0xsoniclabs/sonic/blob/01691793247c838d5b6f56f9b4eda198707e3be4/gossip/blockproc/subsidies/subsidies.go#L84-L85), and needs to consider both possible sponsorship cases -- the network funded and the fund backed option -- since before running the `chooseFund` function, the mode is unclear.

For this test to pass, the funds made available for the sponsored transaction must match the transaction's gas limit + the maximum of these two values.

evmcore/state_processor_test.go
HerbertJordan: Renamed to what?

evmcore/state_processor_test.go
HerbertJordan: Added in ec2caf40181e88c7df2a63c9b54e4c7c2009b93e

evmcore/state_processor_test.go
HerbertJordan: A test case for zero or more post-transactions got added in 95bd08bd27e1957456057e71212479c6d4613d74

simonlechner (approved):
Both Claude and I could not find any issues that are worth mentioning.

PR #1018, opened 2026-05-28
[SONIC-003] Preserve group-level TolerateFailures flag in prepareBundle

thaarok:
convertProposalToPlanInternal built the execution plan group via bundle.NewGroupStep without applying EF_TolerateFailed when the proposal group had TolerateFailures=true. The symmetric path in execution_plan.go correctly called group.WithFlags(EF_TolerateFailed), but the proposal conversion path was missing the same step.

Result: a prepared bundle was stricter than requested — any failed transaction in the group would invalidate the whole bundle even though the caller explicitly set tolerateFailures=true.

Fix: mirror the execution_plan.go pattern after NewGroupStep:
  group = group.WithFlags(bundle.EF_TolerateFailed)

Add test case 'group TolerateFailures flag is preserved' to Test_convertProposalToPlan, which would have caught this regression.

### DIFF
--- a/api/sonicapi/execution_proposal.go
+++ b/api/sonicapi/execution_proposal.go
@@ -386,10 +386,11 @@ func convertProposalToPlanInternal(signer types.Signer, proposalStep any) (bundl
 			return steps[0], nil
 		}
 
-		return bundle.NewGroupStep(
-			step.OneOf,
-			steps...,
-		), nil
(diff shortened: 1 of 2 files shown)


LuisPH3 (approved):
Thanks!

jenikd (approved):

PR #1019, opened 2026-05-28
Add support for handling failed fee charge transactions

HerbertJordan:
This PR adds support for handling failed fee-payment transactions for sponsorships to the fee-tracking infrastructure.

Unlike regular transactions, where fees are charged via the gas bill, fee-payment transactions cover fees through an explicit fee-burn transaction. If this transaction fails, no actual fees are consumed, and no rewards should be attributed to validators, to avoid inflating the token supply.

This is a defensive mechanism, as fee payments should never fail in regular operation.


LuisPH3 (approved):
The changes add a guard around failing sponsored txs payments, to prevent validators from counting this fee for their emissions.
As stated, this should not happen.

PR #1021, opened 2026-05-28
[SONIC-001] Gas power accounting bypass via uint64 overflow

thaarok:
CalcGasPowerUsed accumulates tx.Gas() values using unchecked uint64 addition. A validator can craft an event whose metered transaction gas values sum past math.MaxUint64, wrapping around to a small number.


eventcheck/epochcheck/epoch_check_test.go
```diff
@@ -18,17 +18,86 @@ package epochcheck
...
+		e.SetEpochVote(inter.LlrEpochVote{Epoch: idx.Epoch(1)})
+	}
+	return e
+}
+
+func txWithGas(gas uint64) *types.Transaction {
+	return types.NewTx(&types.LegacyTx{Gas: gas})
+}
+
+func TestCalcGasPowerUsed(t *testing.T) {
```
LuisPH3: Missing tests:
- overflow on `* gasCfg.ParentGas`
- overflow on `* gasCfg.BlockVoteGas`
- overflow on `+ BlockVotesBaseGas` 
- overflow on combining all.
- overflow on `+ tx.Gas` (this one is tested) 

The following expectations should be tested:
-  cases with values equal to Max + 1 (shall clamp)
- cases with values equal to Max  (shall not clamp) 

Case names should describe the expectation:  is clamped or not.

LuisPH3 (commented):
I think that the changes proposed make sense, 
before brio we preserve the existing behavior. We enforce new constraints afterward. 

I have some cosmetic concerns, related to the easiness of reviewing the code:
something like: 
- the existing code should not be pushed into an `if`. Instead, the new behavior should be in the `if`. It is very easy for me to oversee a change in the legacy code, which must be kept verbatim to avoid accidental changes. 
- safeAdd and safeMul could be located in utils package. Name should be clampedAdd and clampedMul (dont reuse gomath names with different behavior).
- math functions shall be tested for boundary conditions. 
- We usually write testing tooling at the end of the file, and we start with the cases. We try to name test files with the pattern `TestFunction_Expectation_WhenScenario`. Focus on reviewing the features and expectations.

simonlechner (approved):
Nice find.

LuisPH3 (approved):

PR #1023, opened 2026-05-28
Flaky testInternalTransactionsDoNotChangeReceiptIndex

facuMH:
This PR fixes https://github.com/0xsoniclabs/sonic-admin/issues/790

**The Problem**
The original approach is inherently "racy" and exits the iterative loop as soon as an epoch is sealed, instead of retrying. 

**The Fix**
Use the same approach as [receipt_tx_index](https://github.com/0xsoniclabs/sonic/blob/main/tests/gas_subsidies/receipt_tx_index_test.go#L174), making it a more robust "racy" attempt to inject the extra transaction into the sealing epoch block.


simonlechner (approved):
I think this should be more robust than the original version.

PR #1024, opened 2026-05-29
Allow blob transactions with BlobHashes to enable ethereum history replay

LuisPH3:
This Change allows the state processor running ethereum history blob txs by enabling two components:

- forward blob hashes to state processor
- use blobBaseFee from block is available

## BlobBaseFee

BlobBase fee is used if available in the block. This PR targets history replays, Sonic does not have the capability of deciding how is this value computed, but it can be reused if the block is already known. 

The code would use two different BlobBasefees  depending on if it is being called via RPC or block processing, blobbasefee is now hardcoded to 1, which is the value used in the block processor and therefore committed to the history. 

## Blob Hashes
Sonic does not implement Blobs, but allows blob transactions as long as they contain no blobs. 
Ethereum does not allow bob transactions without blobs, but geth does not check nil list in the transaction pre-checks. This is used by sonic to execute empty blob txs. 

To enable history runs blob semantics have to be unified.  The system will now produce `BlobHashes` with hashes or nil to enable both semantics. 
The conversion was previously done in `applyTransaction` which is a function with unclear responsibilities and difficult to test. This PR minimizes changes by moving the patching of the blob transaction to  `TxAsMessage`, which is used by both the block processor and RPC calls and traces, therefore enabling the change for `ApplyTransactionWithEVM` as well (the entry point to the state processor from the RPC) 

## Checks
- [x] run bertha on eth
- [x] run bertha on sonic

### DIFF
--- a/evmcore/dummy_block.go
+++ b/evmcore/dummy_block.go
@@ -70,8 +70,8 @@ func GetCoinbase() common.Address {
 }
 
 // GetBlobBaseFee returns the blob base fee to be used by blocks on Sonic networks.
-func GetBlobBaseFee() uint256.Int {
-	return uint256.Int{}
+func GetBlobBaseFee() *uint256.Int {
+	return uint256.NewInt(1)
 }
 
 // NewEvmBlock constructor.
--- a/evmcore/dummy_block.go
+++ b/evmcore/dummy_block.go
@@ -71,6 +71,14 @@ func GetCoinbase() common.Address {
 
 // GetBlobBaseFee returns the blob base fee to be used by blocks on Sonic networks.
 func GetBlobBaseFee() *uint256.Int {
+	// The reason is that https://eips.ethereum.org/EIPS/eip-4844 defines a formula for the computation of the blob-base-fee:
+	//
+	//    the function get_base_fee_per_blob_gas is defines as:
+	//    - BLOB_BASE_FEE = MIN_BASE_FEE_PER_BLOB_GAS * e ^ (header.excess_blob_gas / BLOB_BASE_FEE_UPDATE_FRACTION)
+	//    - MIN_BASE_FEE_PER_BLOB_GAS = 1
+	//    - BLOB_BASE_FEE_UPDATE_FRACTION = 3338477
+	//
+	// In sonic the excess_blob_gas is always 0, since Sonic does not offer blob gas this reduces to 1 * e^0 = 1.
 	return uint256.NewInt(1)
 }
 
--- a/evmcore/evm.go
+++ b/evmcore/evm.go
@@ -77,7 +77,8 @@ func NewEVMBlockContextWithDifficulty(
 		difficulty = big.NewInt(0)
 	}
 
-	blobBaseFee := big.NewInt(1) // default for Sonic networks, overridden by header if present
+	// default for Sonic networks, overridden by header if present
+	blobBaseFee := GetBlobBaseFee().ToBig()
 	if header.BlobBaseFee != nil {
 		blobBaseFee = new(big.Int).Set(header.BlobBaseFee)
 	}
--- a/evmcore/evm.go
+++ b/evmcore/evm.go
@@ -74,7 +74,7 @@ func NewEVMBlockContextWithDifficulty(
 	if header.PrevRandao.Cmp(common.Hash{}) != 0 {
 		random = &header.PrevRandao
 		// Difficulty must be set to 0 when PREVRANDAO is enabled.
-		difficulty = big.NewInt(0)
+		difficulty.SetUint64(0)
 	}
 
 	// default for Sonic networks, overridden by header if present
--- a/evmcore/evm.go
+++ b/evmcore/evm.go
@@ -60,6 +60,7 @@ func NewEVMBlockContextWithDifficulty(
 		baseFee     *big.Int
 		random      *common.Hash
 	)
+
 	// If we don't have an explicit author (i.e. not mining), extract from the header
 	if author == nil {
 		beneficiary = header.Coinbase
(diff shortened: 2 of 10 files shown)


evmcore/state_processor_test.go
```diff
@@ -4769,3 +4746,166 @@ func TestStateProcessor_MetricsAreEnabledInSingleBlockProposerMode(t *testing.T)
...
+func TestTxAsMessage_ConvertsInternalTransactionToMessage(t *testing.T) {
+
+	tx := types.NewTx(&types.LegacyTx{
+		Nonce:    0,
+		To:       &common.Address{1},
+		Gas:      21_000,
+		GasPrice: big.NewInt(1),
+		Value:    big.NewInt(100),
+		Data:     []byte{0x01, 0x02},
+	})
```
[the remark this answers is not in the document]
LuisPH3: emtpy signaure is deliverate

evmcore/evm.go
```diff
@@ -73,7 +74,13 @@ func NewEVMBlockContextWithDifficulty(
 	if header.PrevRandao.Cmp(common.Hash{}) != 0 {
 		random = &header.PrevRandao
 		// Difficulty must be set to 0 when PREVRANDAO is enabled.
-		difficulty.SetUint64(0)
+		difficulty = big.NewInt(0)
```
LorenzSchueler: Why did you touch the difficulty? This should have noting to do with the PR, but if I'm not mistaken, it changes the semantics. Previously the value behind the passed in pointer was changed, while now we create a new pointer, so the original value is no longer updated.

LorenzSchueler (dismissed):

evmcore/evm.go
HerbertJordan: Same question, please clarify why this is changed.

evmcore/state_processor_test.go
```diff
@@ -4769,3 +4746,164 @@ func TestStateProcessor_MetricsAreEnabledInSingleBlockProposerMode(t *testing.T)
...
+				Data:       []byte{0x01, 0x02},
+				BlobHashes: []common.Hash{},
+			},
+			expectations: func(t *testing.T, msg core.Message) {
+				require.Equal(t, big.NewInt(1), msg.GasFeeCap)
+				require.Equal(t, big.NewInt(2), msg.GasTipCap)
+				require.Nil(t, msg.BlobHashes, "Sonic allows empty blob txs, go-ethereum allows processing nil")
+			},
+		},
+	}
```
HerbertJordan, line 5229: nit: set-code transactions are missing.

evmcore/state_processor.go
```diff
@@ -820,7 +820,20 @@ func (tp *TransactionProcessor) Run(i int, tx *types.Transaction) ProcessSummary
 // ApplyTransactionWithEVM attempts to apply a transaction to the given state database
 // and uses the input parameters for its environment similar to ApplyTransaction. However,
 // this method takes an already created EVM instance as input.
-func ApplyTransactionWithEVM(msg *core.Message, config *params.ChainConfig, gp *core.GasPool, statedb state.StateDB, blockNumber *big.Int, blockHash common.Hash, tx *types.Transaction, usedGas *uint64, evm *vm.EVM) (receipt *types.Receipt, err error) {
```
HerbertJordan, line 831: There is one call path to this function where the `msg` is not produced by `TxAsMessage`:
- called in `traceTx` [here](https://github.com/0xsoniclabs/sonic/blob/37c7a9238458e95000061e93baed7c1d0f4a9017/api/ethapi/api.go#L2553)
- called in `TraceCall` [here](https://github.com/0xsoniclabs/sonic/blob/37c7a9238458e95000061e93baed7c1d0f4a9017/api/ethapi/api.go#L2790)

The message is derived [here](https://github.com/0xsoniclabs/sonic/blob/37c7a9238458e95000061e93baed7c1d0f4a9017/api/ethapi/api.go#L2780) from user provided transaction parameters [here](https://github.com/0xsoniclabs/sonic/blob/37c7a9238458e95000061e93baed7c1d0f4a9017/api/ethapi/api.go#L2818) and [here](https://github.com/0xsoniclabs/sonic/blob/37c7a9238458e95000061e93baed7c1d0f4a9017/api/ethapi/transaction_args.go#L284)

So maybe, defensive coding should be applied, and the reset to `nil` for empty `BlobHashes` should be kept.

HerbertJordan (approved):

evmcore/dummy_block.go
```diff
@@ -70,8 +70,8 @@ func GetCoinbase() common.Address {
 }
 
 // GetBlobBaseFee returns the blob base fee to be used by blocks on Sonic networks.
-func GetBlobBaseFee() uint256.Int {
-	return uint256.Int{}
+func GetBlobBaseFee() *uint256.Int {
+	return uint256.NewInt(1)
```
HerbertJordan, line 82: We should document why this is 1 and not 0.

The reason is that [EIP-4844](https://eips.ethereum.org/EIPS/eip-4844) defines a formula for the computation of the blob-base-fee:
- the function `get_base_fee_per_blob_gas` is defines as `MIN_BASE_FEE_PER_BLOB_GAS * e ^ (header.excess_blob_gas / BLOB_BASE_FEE_UPDATE_FRACTION)`
- MIN_BASE_FEE_PER_BLOB_GAS = 1
- BLOB_BASE_FEE_UPDATE_FRACTION = 3338477

Since in sonic the `excess_blob_gas` is always 0, since we do not offer blob gas, this reduces to `1 * e^0 = 1`.

HerbertJordan (dismissed):

evmcore/state_processor.go
LuisPH3: I agree that defensive programming could help, but the code location is notoriously difficult to test. Added the missing code to the args->msg conversion and added a test that checks that args->tx->msg is equivalent to args->msg (which are the two code paths found)

evmcore/evm.go
LuisPH3: reverted. it could and should be done, but out of context.

LorenzSchueler (approved):

PR #1025, opened 2026-06-01
Add devcontainer config.

facuMH:
This PR adds a VS Code [`devcontainer`](https://code.visualstudio.com/docs/devcontainers/create-dev-container) configuration and restructures the `Dockerfile` into named multi-stage targets so the same file serves both development and production workflows.


LuisPH3:
We wont be needing this now.

PR #1034, opened 2026-06-05
Run sponsored tx without post tx is forReplay is enabled

LorenzSchueler:
This PR fixes a bug introduced in #939. When the `forReplay` flag is set, the state processor should process transactions in a way that allows replaying the history. This means it can not introduce additional transactions that are included in the block, because they would then be processed twice (once when they get reintroduced again, and once when replaying the originally introduced tx). In #939 sponsored transactions were therefore treated as regualr transactions to avoid processing the payment tx twice. However, executing the sponsored tx as a normal tx will fail. This PR instead processes them as a sponsored tx, but skips all post transactions.

### DIFF
--- a/evmcore/state_processor.go
+++ b/evmcore/state_processor.go
@@ -326,7 +326,7 @@ func runTransaction(
 			return []ProcessedTransaction{{Transaction: tx}}, core_types.TransactionResultInvalid, 0
 		}
 	}
-	if !context.forReplay && context.upgrades.GasSubsidies && subsidies.IsSponsorshipRequest(tx) {
+	if context.upgrades.GasSubsidies && subsidies.IsSponsorshipRequest(tx) {
 		res, result := context.runner.runSponsoredTransaction(context, tx, trueTxIndexOffset, sizeLimit)
 		execCost := core_types.ExecutionCost(0)
 		for _, r := range res {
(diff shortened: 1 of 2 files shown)


facuMH (approved):

evmcore/state_processor.go
```diff
@@ -472,6 +472,11 @@ func (r *transactionRunner) runSponsoredTransactionInternal(
 
 	gasUsed := processed.Receipt.GasUsed
 
+	// Skip post-execution transactions when replaying.
+	if ctxt.forReplay {
```
HerbertJordan, line 476: For a follow-up: Do you think there would be a way to verify that during the replay the same post-transactions would be proposed? This would extend replay test coverage to the sponsorship mechanism.

HerbertJordan (approved):
The change looks good, thanks for identifying and fixing this.

LuisPH3 (approved):

evmcore/state_processor.go
LorenzSchueler: I guess we could:
- on the Bertha side filter out the post transactions and build an artificial block without them
- replay this artificial block (without skipping post transactions)
- record the state root
- revert the DB to before the block
- replay the real block (and skip post transactions)
- check that the state root matches the recorded one

But that means repay is at least 2x slower (for blocks containing sponsored transactions)

PR #1035, opened 2026-06-08
Delay on-new-log notification until the transaction is confirmed

LuisPH3:
This PR delays the notification of new log messages until the transaction causing the log messages got confirmed to be included in a block.

Before this change, log messages of transactions that later get rolled back are reported to listeners. Among others, this could trigger changes in the validator's stake not reflected in the block history.

This PR is a revision on https://github.com/0xsoniclabs/sonic/pull/1032 with 2 contributions: 

- It moves the log emission code outside of code paths used by trial-run and single block proposal: 

The `onNewLog` callback is an integration between the block processor and the driver contract, I believe that the natural location for this code would be c_block_callbacks, but because of the difficulty of testing it in that location the code is moved to the first immediate location called from the consensus. 
The callback is not propagated into internal tools used by other facilities which execute blocks or part of blocks, as it is not needed. 

- It adds a test testing the original issue which triggered the refactor:
rolled back transactions shall not call the `onNewLog` callback.


evmcore/state_processor_test.go
```diff
@@ -863,63 +1225,6 @@ func TestApplyTransaction_SetsEffectiveGasPriceInReceipt(t *testing.T) {
 	}
 }
 
-func TestApplyTransaction_CollectsLogsFromStateDbIntoReceipt(t *testing.T) {
```
HerbertJordan: This test is not relevant for the change in this PR and should be preserved.

evmcore/state_processor_test.go
```diff
@@ -509,6 +514,363 @@ func TestProcessWithDifficulty_UsesProvidedDifficulty(t *testing.T) {
...
+
+	currentTxIndex := 0
+	stateDb := state.NewMockStateDB(ctrl)
+	stateDb.EXPECT().SetTxContext(gomock.Any(), gomock.Any()).Do(
+		func(_ common.Hash, index int) {
+			currentTxIndex = index
+		},
+	).AnyTimes()
+	stateDb.EXPECT().TxIndex().DoAndReturn(func() int { return currentTxIndex }).AnyTimes()
+	stateDb.EXPECT().GetLogs(gomock.Any(), gomock.Any()).Return([]*types.Log{{Address: common.Address{1}, TxIndex: 0}}).AnyTimes()
```
HerbertJordan: nit: AnyTimes() => MinTimes(1) or nothing? 

Just to make sure the test reaches the log emission.
Alternatively, check that the receipts has a log after line 793.

evmcore/state_processor_test.go
```diff
@@ -509,6 +514,363 @@ func TestProcessWithDifficulty_UsesProvidedDifficulty(t *testing.T) {
...
+	ctrl := gomock.NewController(t)
+
+	key, err := crypto.GenerateKey()
+	require.NoError(t, err)
+	signer := types.FrontierSigner{}
+	signed := func(tx *types.Transaction) *types.Transaction {
+		signedTx, err := types.SignTx(tx, signer, key)
+		require.NoError(t, err)
+		return signedTx
+	}
```
HerbertJordan: nit: while correct, types.MustSignNewTx(...) is the typically used for this, which I think is more readable. Wdyt?

evmcore/state_processor_test.go
```diff
@@ -509,6 +514,363 @@ func TestProcessWithDifficulty_UsesProvidedDifficulty(t *testing.T) {
...
+		},
+		big.NewInt(1),
+		math.MaxUint64,
+	)
+
+	require.Len(t, summary.ProcessedTransactions, len(transactions))
+	require.Equal(t, []*core_types.Log{
+		core_types.CoreLogFromGethLog(&types.Log{Address: common.Address{10}, TxIndex: 0}),
+		core_types.CoreLogFromGethLog(&types.Log{Address: common.Address{20}, TxIndex: 1}),
+		core_types.CoreLogFromGethLog(&types.Log{Address: common.Address{30}, TxIndex: 1}),
```
HerbertJordan: One of the main reasons for `core_types.CoreLogFromGethLog` is that the transaction index is stripped out. So the TxIndex values here are not actually tested and irrelevant for the test setup. These details could be stripped out.

evmcore/state_processor_test.go
```diff
@@ -509,6 +514,363 @@ func TestProcessWithDifficulty_UsesProvidedDifficulty(t *testing.T) {
...
+			},
+			logsByTxIndex: map[common.Hash][]*types.Log{
+				tx1.Hash(): {
+					{Address: common.Address{2}, TxIndex: 2},
+					{Address: common.Address{3}, TxIndex: 2},
+					{Address: common.Address{4}, TxIndex: 2},
+				},
+			},
+			expectedCallbacks: []*core_types.Log{
+				core_types.CoreLogFromGethLog(&types.Log{Address: common.Address{2}, TxIndex: 2}),
```
HerbertJordan: Transaction indices are stripped away by `core_types.CoreLogFromGethLog`. So adding those here is not really effective.

HerbertJordan (dismissed):
Looks good, just some minor things in the tests. Feel free to ignore.

evmcore/state_processor_test.go
LuisPH3: reverted

evmcore/state_processor_test.go
LuisPH3: replaced

evmcore/state_processor_test.go
LuisPH3: logs simplified

evmcore/state_processor_test.go
LuisPH3: replaced index with other log fields, simplified log expectations.

LorenzSchueler (approved):

HerbertJordan (approved):

PR #1037, opened 2026-06-08
Fix Service/Handler race condition

facuMH:
This PR fixes https://github.com/0xsoniclabs/sonic-admin/issues/407

**The problem:**
Given the right heavy load conditions, a services spawned by [p2p protocol](https://github.com/0xsoniclabs/sonic/blob/main/gossip/service.go#L508) could call `wg.Add(1)` after the handler already started shutting down (aka, call [wg.Wait](https://github.com/0xsoniclabs/sonic/blob/main/gossip/handler.go#L500)). To the runtime race detection, this is interpreted as a race condition.
It is also an anti-pattern or misuse of the [`WaitGroup`](https://pkg.go.dev/sync#WaitGroup).

**The proposed solution:**
Since there is no real control of when the task will be started, instead of using a wait group, this PR proposes to use a [Condition variable](https://pkg.go.dev/sync#Cond) with a mutex.

### DIFF
--- a/gossip/handler.go
+++ b/gossip/handler.go
@@ -496,6 +496,11 @@ func (h *handler) Stop() {
 	// After this send has completed, no new peers will be accepted.
 	close(h.quitSync)
 
+	// Disconnect existing sessions.
+	// This also closes the gate for any new registrations on the peer set.
+	// sessions which are already established but not added to h.peers yet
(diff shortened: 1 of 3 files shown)


LuisPH3:
overseen by https://github.com/0xsoniclabs/sonic/pull/1113

PR #1039, opened 2026-06-09
Update max code size to eth Amsterdam values

simonlechner:
[EIP-7954](https://eips.ethereum.org/EIPS/eip-7954) has increased the max contract code size and max init code size again. To already support ethereum's next update these values are adjusted.

### DIFF
--- a/opera/vm_config.go
+++ b/opera/vm_config.go
@@ -26,8 +26,8 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


simonlechner:
This update will not be performed for the current version due to interference with the maximum transaction size (128KiB) and the missing updated gas calculation for state data of [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037).

PR #1044, opened 2026-06-11
Protect parsing of execution plan from stack overflow

HerbertJordan:
When loading a bundle, the execution plan is first parsed and then validated. The validation limits the nesting depth of execution plan steps. However, since the validation happens after the parsing, this limit is not enforced for the parser.

An attacker could create a serialized version of a deeply nested execution plan. The RLP decoding, being a recursive procedure, would thus start building up a very deep stack, which can lead to the exceeding of Go's maximum stack size, causing a panic. 

This change introduces a pre-check for the nesting depth of a serialized execution plan before performing the actual parsing. Anything significantly deeper than a valid execution plan will fail the parsing step -- and is thus classified as an invalid bundle.

### DIFF
--- a/gossip/blockproc/bundle/execution_plan_test.go
+++ b/gossip/blockproc/bundle/execution_plan_test.go
@@ -18,11 +18,13 @@ package bundle
 
 import (
 	"bytes"
+	"encoding/binary"
 	"fmt"
 	"testing"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
 )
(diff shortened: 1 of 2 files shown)


gossip/blockproc/bundle/execution_plan_test.go
```diff
@@ -460,6 +462,130 @@ func TestExecutionStep_decode_FailsOnInvalidInput(t *testing.T) {
...
+	deep := wrapStepInAllOfGroups(leaf, 1_000)
+
+	var s ExecutionStep
+	require.ErrorContains(s.decode(bytes.NewReader(deep)), "nesting depth")
+
+	// Build a chain of nested group steps that would cause a stack overflow if
+	// the depth guard were not in place, and verify that it is rejected with an
+	// error.
+	deep = wrapStepInAllOfGroups(leaf, 1_000_000)
+	require.ErrorContains(s.decode(bytes.NewReader(deep)), "nesting depth")
```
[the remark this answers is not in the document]
HerbertJordan: It only uses less than 10 MB and it is deliberately introduced as a proof of concept.

gossip/blockproc/bundle/execution_plan_test.go
```diff
@@ -460,6 +462,130 @@ func TestExecutionStep_decode_FailsOnInvalidInput(t *testing.T) {
...
```
LorenzSchueler: Maybe I'm missing something, but what do we gain from having a manual group wrapper that has the same output as the real encoder? Why not use the real encoder in the first place? The manual one seems a bit faster, but for me this below runs in <1s.

```suggestion

	// Build a chain of nested group steps far deeper than the permitted limit.
	// A few thousand levels is orders of magnitude beyond maxStepEncodingRlpDepth.
	nested := leafStep
	for range 1_000 {
		nested = NewAllOfStep(nested)
	}
	var nestedBuf bytes.Buffer
	require.NoError(nested.encode(&nestedBuf))

	var s ExecutionStep
	require.ErrorContains(s.decode(bytes.NewReader(nestedBuf.Bytes())), "nesting depth")

	// Build a chain of nested group steps that would cause a stack overflow if
	// the depth guard were not in place, and verify that it is rejected with an
	// error.
	nested = leafStep
	for range 1_000_000 {
		nested = NewAllOfStep(nested)
	}
	require.NoError(nested.encode(&nestedBuf))
	require.ErrorContains(s.decode(bytes.NewReader(nestedBuf.Bytes())), "nesting depth")
```

LorenzSchueler (dismissed):
The fix looks good. For the tests I don't understand the benefit of the handwritten encoder.

gossip/blockproc/bundle/execution_plan_test.go
HerbertJordan: When I used the existing one, the encoder also ran into a stack overflow ...  that's why the custom one was required. I will check again ..

HerbertJordan: You are right, the encoder is able to handle the 1M case, while the decoder fails.

I simplified the test accordingly in 23cc21c808f52225acdff30d2f4868af66b0f625

LorenzSchueler (approved):

LuisPH3 (approved):

PR #1051, opened 2026-07-03
Prepare release branch for new release cycle

LuisPH3:
This PR prepares 2.2 branch for next releases

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -38,13 +38,13 @@ const (
 	// The patch version, which must only be non-zero for release candidates
 	// and official releases. All development versions must have a patch
 	// version of 0.
-	Patch = 0
+	Patch = 1
 
 	// The pre-release version. This is set to "dev" for development versions
 	// on the main branch and should be updated to "rc.X" for release candidates
 	// on release branches only. For a final release, this must be set to an
 	// empty string. All other values are invalid.
-	PreRelease = ""
+	PreRelease = "rc.1"
 )
 
 // Get returns the complete version information.


version/version.go
```diff
@@ -38,13 +38,13 @@ const (
...
 	// version of 0.
-	Patch = 0
+	Patch = 1
 
 	// The pre-release version. This is set to "dev" for development versions
 	// on the main branch and should be updated to "rc.X" for release candidates
 	// on release branches only. For a final release, this must be set to an
 	// empty string. All other values are invalid.
-	PreRelease = ""
+	PreRelease = "rc.1"
```
HerbertJordan, line 47: We do have a bit of an inconsistency here .. this version of the code would identify itself as `2.2.1-rc.1` -- which technically is not correct, since it is not tagged as such. In practice, we will cherry-pick fixes on this branch, and only if we have a set of those, and tagged it, it would be `2.2.1-rc.1`. It is thus not easy to deduce what is running as part of which release candidate. If we want to be accurate here, we would have to version the in-between codes as `2.2.1-rc.1.dev`. So the order would go

2.2.0 -> 2.2.1-rc.1.dev -> 2.2.1-rc.1 -> 2.2.1-rc.2.dev -> 2.2.1-rc.2 -> 2.2.1

This might be a bit annoying, since it requires 2 PRs per step. One to remove the .dev or rc.2 suffix, which is then tagged, and the another to start the new dev-version.

The only place where this really shows up, that I know of, is in our network scanner that tries to detect the client version composition. So it might not be an issue at all. Just wanted to mention it.

Wdyt?

LuisPH3: I am aware of this situation, to be honest i do not see much value on this file and it most an inconvenience that needs to be maintained for now.
having the release branch tagged as dev requires us to patch this file again before the tag is done, and the risk of forgetting doing that is high.

I would consider fetching the "PreRelease" from the tags in a build step and automatize this in the future.

PR #1054, opened 2026-07-08
Recover panics in P2P message handling

thaarok:
P2P hardening - if panic in P2P occurs, disconnect the peer, don't crash the node.

### DIFF
--- a/gossip/handler.go
+++ b/gossip/handler.go
@@ -23,6 +23,7 @@ import (
 	"log/slog"
 	"math"
 	"math/rand/v2"
+	"runtime/debug"
 	"strings"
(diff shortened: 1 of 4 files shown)


LuisPH3 (dismissed):
The proposed changes do no modify existing logic, but it is very difficult to judge if they prevent crash or system malfunctioning. 

Without testing, it is not guaranteed that the node does not crash after continuing execution, or that clean-ups succeed. 
This does not introduce any new fail mode to the original reported panic, Therefore, I think we can merge these changes. The code could use testing though.

LuisPH3 (approved):

PR #1062, opened 2026-07-13
Fail with distinct error code

simonlechner:
In case permanent lock can not be written, sonic should fail with a distinct error code to ensure scripted runners do not restart automatically.

Part of the fixes for the internal audit.

### DIFF
--- a/utils/errlock/errlock.go
+++ b/utils/errlock/errlock.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"runtime/debug"
 
 	"github.com/ethereum/go-ethereum/log"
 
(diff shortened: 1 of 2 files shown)


LuisPH3 (changes requested):
This early return prevents the panic with the backlog presenting the location of the crash. Returning 74 is a good idea, both code paths must do it, but the stack trace must be printed as well, independently of the result of the write operation

thaarok (dismissed):
Note: we have to also update documentation, to include following into systemd unit files:
```
[Service]
Restart=on-failure
RestartPreventExitStatus=74
```

utils/errlock/errlock.go
```diff
@@ -59,12 +62,25 @@ func (l *ErrorLock) Check() error {
...
+	eLockPath, writeErr := write(l.dataDir, err.Error())
+	if writeErr != nil {
+		log.Error("errlock: failed to write halt-lock file; terminating"+
+			" with distinct exit code so the node is not silently restarted"+
+			" into the same unsafe state that triggered Permanent()",
+			"path", eLockPath,
+			"originalErr", err,
+			"writeErr", writeErr,
+		)
+	} else {
```
LuisPH3: I would like to get rid of this else, the reason:
Something when wrong, so that this function is called, this justifies the error in line 76.
during error handling, something new goes wrong, this justifies an extra error, but it does not replace the original error.

LuisPH3 (changes requested):

utils/errlock/errlock.go
simonlechner: Good call, removed the else check.

LuisPH3 (approved):

PR #1064, opened 2026-07-13
Add http.Servers timeouts

simonlechner:
This PR adds timeouts to the standalone pprof and metrics http.Servers.

### DIFF
--- a/cmd/sonicd/metrics/flags.go
+++ b/cmd/sonicd/metrics/flags.go
@@ -207,10 +207,11 @@ func newMetricsServer(address string) *http.Server {
 	m.Handle("/debug/metrics/prometheus", gethprometheus.Handler(metrics.DefaultRegistry))
 	m.Handle("/debug/metrics/prometheus/native", promhttp.Handler())
 	return &http.Server{
-		Addr:         address,
-		Handler:      m,
-		ReadTimeout:  30 * time.Second,
-		WriteTimeout: 30 * time.Second,
-		IdleTimeout:  120 * time.Second,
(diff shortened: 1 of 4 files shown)


cmd/sonicd/metrics/flags.go
```diff
@@ -192,14 +192,25 @@ func SplitTagsFlag(tagsFlag string) map[string]string {
...
+	m.Handle("/debug/metrics", exp.ExpHandler(metrics.DefaultRegistry))
+	m.Handle("/debug/metrics/prometheus", gethprometheus.Handler(metrics.DefaultRegistry))
+	m.Handle("/debug/metrics/prometheus/native", promhttp.Handler())
+	return &http.Server{
+		Addr:         address,
+		Handler:      m,
+		ReadTimeout:  30 * time.Second,
+		WriteTimeout: 30 * time.Second,
+		IdleTimeout:  120 * time.Second,
+	}
```
thaarok, line 216: Why not `ReadHeaderTimeout: 5 * time.Second,` to be consistent with the pprof timeouts bellow?

thaarok (dismissed):
I am not sure if tests TestNewPProfServer_HasTimeouts and TestNewMetricsServer_SetsDefaultValues are meaningful, since they only tests assigning into a structure, but if it is required by the coverage requirements...

cmd/sonicd/metrics/flags.go
simonlechner: Good catch, fixed.

LuisPH3 (approved):

PR #1066, opened 2026-07-15
Stop filter event loop before closing the store

simonlechner:
This PR cleans up the shutdown process of the client. The filter system loop is now guaranteed to stop before the database caches (that are read from) are set to nil.


LuisPH3 (approved):
The PR adds the missing shutdown procedure for the filter RPC backend, which is an active component and needs to stop during shutdown.

PR #1072, opened 2026-07-20
Move proxy contract out of subsidies

LorenzSchueler:
This PR moves the proxy contract out of the subsidies directory.
The proxy delegates to another contract, but the contract it delegates to can be changed. This essentially implements an updatable contract.

It is currently only used for the subsidies, but in the future this contract will be also used for transaction priorities (GBS)


LuisPH3 (approved):
👍

PR #1075, opened 2026-07-20
Add basic types for transaction priorities

LorenzSchueler:
This PR adds the `Priority` and `Config` types and related methods. These two types are the return types of the `getPriority` and `getPriorityConfig` contract calls which will be added in the next PR. The `Priority` defines which priority should be applied to a given transaction and the entity under which the priority is registered. The `Config` contains the limits that are applied during event emission and block formation.

### DIFF
--- a/gossip/blockproc/priorities/types_test.go
+++ b/gossip/blockproc/priorities/types_test.go
@@ -1,0 +1,74 @@
+// Copyright 2026 Sonic Operations Ltd
+// This file is part of the Sonic Client
+//
+// Sonic is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Sonic is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with Sonic. If not, see <http://www.gnu.org/licenses/>.
+
+package priorities
+
+import (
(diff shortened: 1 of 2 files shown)


gossip/blockproc/priorities/types_test.go
```diff
@@ -0,0 +1,66 @@
...
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			require.Equal(t, tc.want, tc.a.Cmp(tc.b))
+		})
+	}
+}
+
+// Prio builds a Priority for tests. The id is limited to a single byte for
+// simplicity.
+func Prio(level uint64, weight uint64, id byte) Priority {
```
LuisPH3, line 64:
```suggestion
func NewPriority(level uint64, weight uint64, id byte) Priority {
```
because of go conventions

LuisPH3: I think that this would work though: 
```
  {{2, 0, 0}, {1, 1, 0}, 1},
```

LuisPH3 (dismissed):
The PR introduces basic types to sort transactions in priority categories, it adds other advanced tooling related to other features, the scope does not allow to

gossip/blockproc/priorities/types_test.go
LorenzSchueler: Any change just adds a lot of verbosity. I would prefer to keep it as is.
```
"higher level wins":                        {Priority{2, 0, [16]byte{0}}, Priority{1, 1, [16]byte{0}}, 1},
"lower level loses":                        {NewPriority(1, 1, 0), NewPriority(2, 0, 0), -1},
"equal non-zero level, higher weight wins": {Prio(1, 2, 0), Prio(1, 1, 0), 1},
```

LorenzSchueler:
@thaarok I changed the types to u64/u128. Is this fine now?

thaarok (approved):
yes, thanks!

PR #1076, opened 2026-07-20
Add placeholder priority contract and functions to call it

LorenzSchueler:
This PR adds a stand-in `PriorityRegistry` contract and `GetPriority`/`GetConfig` calls gated on the `TransactionPriorities` flag, with hand-rolled ABI encoding, strict validation, and safe degradation to `FallbackConfig`.


gossip/blockproc/priorities/registry/priorities_registry.sol
```diff
@@ -0,0 +1,93 @@
...
+    // --- interface consumed by the Sonic client ---
+
+    function getPriority(
+        address from,
+        address /*to*/,
+        uint256 /*value*/,
+        uint256 /*nonce*/,
+        bytes calldata /*data*/,
+        uint256 gas
+    ) external view returns (uint256 level, uint256 weight, bytes32 id) {
```
thaarok: I think `uint256` for level and for weight is unnecessary big type for this purpose.

I would suggest to use a shorter type, at least `uint128`, which would allow to store level+weight in a single storage slot. (Store Priority in 2 slots instead of 3 slots)

LorenzSchueler: I agree - a smaller integer should be fine. Even `uint64` is probably enough. Which one should we go with?

thaarok: We may fit even the id into the same slot and save performace for these calls - uint64 should be sufficient even for the entity id.
What about `uint64 level + uint64 weight + uint128 entityId` ?

LuisPH3 (dismissed):

thaarok (dismissed):

LuisPH3 (approved):

PR #1082, opened 2026-07-22
Deploy priority registry in fake genesis and report as system contract

LorenzSchueler:
This PR deploys the priority registry in the fake genesis when the priorities feature is enables and report as it as a system contract.


LuisPH3 (dismissed):
The PR adds the new registry to the fake net and updates rpc methods to publish the address

simonlechner (approved):

PR #1083, opened 2026-07-23
Add priority ordering core

LorenzSchueler:
This PR adds the pure ordering core of the transaction-priorities feature.
This is used in a follow-up to order the transactions in the `c_block_callbacks.go` after the scrambler run.

`Prioritize` reorders a base-ordered, pre-filtered transaction list so that registry-prioritized transactions form a prefix in (level desc, weight desc, hash asc) order, subject to two coupled constraints:
- Per-sender nonce ordering — a transaction keeps its priority only while it extends its sender's contiguous run of prioritized nonces from the block-start account nonce; stale nonces are demoted and the first gap demotes the rest of the sender's txs (they would be nonce-too-high in the prefix).
- Per-entity gas budget — each entity may spend at most `Config.MaxGasPerEntityPerBlock` gas (by gas limit) on prioritized transactions. Selection walks the per-sender frontiers greedily; a frontier that doesn't fit its entity's remaining budget blocks that sender's run, so budget is only spent on transactions that can actually execute in the prefix.
Everything not selected (non-prioritized, nonce-blocked, over budget) keeps its relative base order after the prefix. The function is a pure, deterministic total function of its inputs. 

Classification is abstracted behind a Classifier interface (the EVM-backed implementation and the block-formation call site land in a follow-up).

Unit tests cover each helper (classification, nonce-run, budgeted prefix selection, packing) plus an end-to-end scenario exercising all constraints together.

### DIFF
--- a/gossip/blockproc/priorities/ordering.go
+++ b/gossip/blockproc/priorities/ordering.go
@@ -40,16 +40,17 @@ type NonceReader interface {
 	GetNonce(common.Address) uint64
 }
 
-// txWithPrio pairs a transaction with its classified priority.
-type txWithPrio struct {
+// transactionWithPriority pairs a transaction with its classified priority.
+type transactionWithPriority struct {
 	tx       *types.Transaction
 	priority Priority
 }
 
-// cmpLevelWeightHash orders two txWithPrio for the prioritized prefix, returning
-// >0 when x precedes y, i.e. by (level desc, weight desc, hash asc). Note this is
-// the opposite sign convention to the ascending cmpNonceHash.
-func (x txWithPrio) cmpLevelWeightHash(y txWithPrio) int {
+// cmpLevelWeightHash orders two transactionWithPriority for the prioritized
+// prefix, returning >0 when x precedes y, i.e. by (level desc, weight desc,
+// hash asc). Note this is the opposite sign convention to the ascending
+// cmpNonceHash.
+func (x transactionWithPriority) cmpLevelWeightHash(y transactionWithPriority) int {
 	if c := cmp.Compare(x.priority.Level, y.priority.Level); c != 0 {
 		return c
 	}
@@ -59,8 +60,8 @@ func (x txWithPrio) cmpLevelWeightHash(y txWithPrio) int {
 	return bytes.Compare(y.tx.Hash().Bytes(), x.tx.Hash().Bytes())
 }
 
-// cmpNonceHash orders two txWithPrio by (nonce asc, hash asc).
-func (x txWithPrio) cmpNonceHash(y txWithPrio) int {
+// cmpNonceHash orders two transactionWithPriority by (nonce asc, hash asc).
+func (x transactionWithPriority) cmpNonceHash(y transactionWithPriority) int {
 	if c := cmp.Compare(x.tx.Nonce(), y.tx.Nonce()); c != 0 {
 		return c
 	}
@@ -72,7 +73,7 @@ func (x txWithPrio) cmpNonceHash(y txWithPrio) int {
 // subject to two coupled constraints:
 //
 //   - Per-sender nonce ordering: a transaction keeps its priority only while it
-//     extends its sender's contiguous run of prioritized nonces from the
+//     extends its sender's contiguous sequence of prioritized nonces from the
 //     block-start account nonce. This keeps each sender in nonce order: a later
 //     nonce hoisted ahead of a lower-nonce predecessor left behind would be
 //     nonce-too-high and skipped.
@@ -81,7 +82,7 @@ func (x txWithPrio) cmpNonceHash(y txWithPrio) int {
 //
 // Transactions that are not selected — non-prioritized, nonce-blocked, or over
 // budget — are pushed back by the hoisted prioritized prefix but keep their
-// relative base order among themselves.
+// original relative order among themselves.
 //
 // The base order must already be the mode-specific order (scrambler output in
 // legacy mode, proposal order in single-proposer mode) and must already be
@@ -105,11 +106,11 @@ func Prioritize(
 	txsWithPrio := classify(base, classifier)
 
 	// Collect the prioritized transactions per sender ordered by nonce which
-	// form a continuous run starting at the start-of-block sender nonce.
-	prioSenderRuns := prioritizedSenderRuns(txsWithPrio, signer, state)
+	// form a continuous sequence starting at the start-of-block sender nonce.
+	prioSenderSequences := prioritizedSenderSequences(txsWithPrio, signer, state)
 
 	// Collect the transaction indices which should form the block txs prefix.
-	prioPrefixIndices := prioritizedTxsPrefix(txsWithPrio, prioSenderRuns, cfg.MaxGasPerEntityPerBlock)
+	prioPrefixIndices := computePrioritizedTxsPrefix(txsWithPrio, prioSenderSequences, cfg.MaxGasPerEntityPerBlock)
 
 	return combinePrioPrefixWithRemainder(txsWithPrio, prioPrefixIndices)
 }
@@ -117,27 +118,28 @@ func Prioritize(
 // classify pairs each transaction with its priority, preserving order. A
 // classifier error is treated as "not prioritized" (zero Priority), which
 // keeps Prioritize a total function of its inputs.
-func classify(base types.Transactions, classifier Classifier) []txWithPrio {
-	txsWithPrio := make([]txWithPrio, len(base))
+func classify(base types.Transactions, classifier Classifier) []transactionWithPriority {
+	txsWithPrio := make([]transactionWithPriority, len(base))
 	for i, tx := range base {
 		p, err := classifier.Priority(tx)
 		if err != nil {
 			p = Priority{} // deterministic failure rule: errors => not prioritized
 		}
-		txsWithPrio[i] = txWithPrio{tx: tx, priority: p}
+		txsWithPrio[i] = transactionWithPriority{tx: tx, priority: p}
 	}
 	return txsWithPrio
 }
 
-// prioritizedSenderRuns groups the prioritized entries by sender and reduces
-// each sender to its run: its prioritized transactions in nonce order forming a
-// contiguous run from the block-start account nonce. Stale nonces (below the
-// account nonce) are skipped and the first gap ends the run, as later nonces are
-// unreachable. A transaction whose sender cannot be recovered is left
-// non-prioritized. It returns, per sender, the entry indices of the run in
-// nonce order; senders left with an empty run are omitted.
-func prioritizedSenderRuns(
-	txsWithPrio []txWithPrio,
+// prioritizedSenderSequences groups the prioritized entries by sender and
+// reduces each sender to its sequence: its prioritized transactions in nonce
+// order forming a contiguous sequence from the block-start account nonce. Stale
+// nonces (below the account nonce) are skipped and the first gap ends the
+// sequence, as later nonces are unreachable. A transaction whose sender cannot
+// be recovered is left non-prioritized. It returns, per sender, the entry
+// indices of the sequence in nonce order; senders left with an empty sequence
+// are omitted.
+func prioritizedSenderSequences(
+	txsWithPrio []transactionWithPriority,
 	signer types.Signer,
 	state NonceReader,
 ) map[common.Address][]int {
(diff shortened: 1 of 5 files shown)


gossip/blockproc/priorities/types.go
```diff
@@ -60,9 +60,9 @@ type Config struct {
 	// MaxGasPerEntityPerBlock bounds the total gas limit of prioritized
 	// transactions of one entity in a single block (authoritative, enforced at
 	// block formation).
-	// Transactions are packed in (level desc, weight desc, hash asc) order
-	// until the next transaction would exceed the budget; the remainder is
-	// demoted.
+	// Transactions are packed in (level desc, weight desc, hash asc) order until
+	// one would exceed the budget; that transaction and the later nonces of its
+	// sender are demoted, while other senders of the same entity keep packing.
```
LuisPH3: packing?

gossip/blockproc/priorities/ordering.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+// NonceReader reads block-start account nonces; it is the subset of
+// state.StateDB that Prioritize needs to enforce per-sender nonce ordering.
+type NonceReader interface {
+	GetNonce(common.Address) uint64
+}
+
+// txWithPrio pairs a transaction with its classified priority.
+type txWithPrio struct {
```
LuisPH3: lets try to keep our good habits, do not use abreviations like Prio, tx is good for a variable, not for a type
maybe we need agent instructions

gossip/blockproc/priorities/ordering.go
```diff
@@ -0,0 +1,249 @@
...
+//     extends its sender's contiguous run of prioritized nonces from the
+//     block-start account nonce. This keeps each sender in nonce order: a later
+//     nonce hoisted ahead of a lower-nonce predecessor left behind would be
+//     nonce-too-high and skipped.
+//   - Per-entity gas budget: an entity may spend at most
+//     cfg.MaxGasPerEntityPerBlock gas (gas limit) on prioritized transactions.
+//
+// Transactions that are not selected — non-prioritized, nonce-blocked, or over
+// budget — are pushed back by the hoisted prioritized prefix but keep their
+// relative base order among themselves.
```
LuisPH3: I understand: in the original relative order?

gossip/blockproc/priorities/ordering.go
```diff
@@ -0,0 +1,249 @@
...
+		return c
+	}
+	if c := cmp.Compare(x.priority.Weight, y.priority.Weight); c != 0 {
+		return c
+	}
+	return bytes.Compare(y.tx.Hash().Bytes(), x.tx.Hash().Bytes())
+}
+
+// cmpNonceHash orders two txWithPrio by (nonce asc, hash asc).
+func (x txWithPrio) cmpNonceHash(y txWithPrio) int {
```
LuisPH3: here, or in Prioritize: add a note about priorities, why nonce and then hash

gossip/blockproc/priorities/ordering.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+// prioritizedSenderRuns groups the prioritized entries by sender and reduces
+// each sender to its run: its prioritized transactions in nonce order forming a
+// contiguous run from the block-start account nonce. Stale nonces (below the
+// account nonce) are skipped and the first gap ends the run, as later nonces are
+// unreachable. A transaction whose sender cannot be recovered is left
+// non-prioritized. It returns, per sender, the entry indices of the run in
+// nonce order; senders left with an empty run are omitted.
+func prioritizedSenderRuns(
```
LuisPH3: nit: sequence instead of run, no verb in function.

`extractPrioritizedSenderSequences`

gossip/blockproc/priorities/ordering.go
```diff
@@ -0,0 +1,249 @@
...
+
+// prioritizedTxsPrefix walks the per-sender runs greedily, returning the
+// entry indices in prioritized-prefix order. Each step takes the highest-priority
+// eligible frontier transaction (a sender's lowest un-selected nonce) and
+// advances that sender. A frontier that does not fit its entity budget removes
+// the sender's remaining transactions (its later nonces depend on it),
+// so a budget is only ever spent on transactions that can actually execute in
+// the prioritized prefix. It consumes bySender, mutating it as senders are
+// advanced and exhausted.
+func prioritizedTxsPrefix(
```
LuisPH3: `computePrioritizedTxsPrefix`

gossip/blockproc/priorities/ordering.go
```diff
@@ -0,0 +1,249 @@
...
+	budgetOf := func(id [16]byte) uint64 {
+		if r, ok := remaining[id]; ok {
+			return r
+		}
+		return perEntityBudget
+	}
+	for len(bySender) > 0 {
+		best := -1
+		var bestSender common.Address
+		for sender, run := range bySender {
```
LuisPH3: I wonder if building a heap would be a better algorithm than this search.

gossip/blockproc/priorities/ordering_test.go
```diff
@@ -0,0 +1,478 @@
...
+
+func (fakeSigner) Equal(types.Signer) bool { return false }
+
+// fakeNonceReader is a NonceReader backed by per-address block-start account nonces.
+type fakeNonceReader map[common.Address]uint64
+
+func (f fakeNonceReader) GetNonce(sender common.Address) uint64 { return f[sender] }
+
+// makeTxN creates a transaction with the given nonce and 21k gas.
+func makeTxN(nonce uint64) *types.Transaction {
```
LuisPH3: lets call it makeTxsNonce at least. I had to come down here to understand what was i reading

gossip/blockproc/priorities/ordering_test.go
```diff
@@ -0,0 +1,478 @@
...
+			entries: []txWithPrio{
+				{makeTxN(0), Prio(1, 20, 100)},
+				{makeTxN(1), Prio(1, 40, 100)},
+				{makeTxN(0), Prio(1, 30, 101)},
+				{makeTxN(1), Prio(1, 10, 101)},
+			},
+			bySender: map[common.Address][]int{{0xa}: {0, 1}, {0xb}: {2, 3}},
+			budget:   1_000_000,
+			want:     []int{2, 0, 1, 3},
+		},
```
LuisPH3, line 379: Add a test case where a sender has transactions in nonce sequence but the priorities are in the inverse sequence. what should happen then?

LuisPH3 (dismissed):
The pr proposes an algorithm which sorts a list of transactions using an order:
the order prepends transactions based in their priority level/weight. 
It respects the nonce order of prioritised transactions. tiebreakers with hash make replicated resolution deterministic.

gossip/blockproc/priorities/ordering.go
LorenzSchueler: Runtime complexity wise yes - I'm just not sure if this matters and I thought instead of using just another data structure lets keep this simple for now and benchmark later.

gossip/blockproc/priorities/ordering_test.go
LorenzSchueler: This is technically already covered by the existing cases, but I'll add one just to make it explicit.

LuisPH3 (dismissed):

gossip/blockproc/priorities/ordering.go
```diff
@@ -0,0 +1,255 @@
...
+
+// cmpLevelWeightHash orders two transactionWithPriority for the prioritized
+// prefix, returning >0 when x precedes y, i.e. by (level desc, weight desc,
+// hash asc). Note this is the opposite sign convention to the ascending
+// cmpNonceHash.
+func (x transactionWithPriority) cmpLevelWeightHash(y transactionWithPriority) int {
+	if c := cmp.Compare(x.priority.Level, y.priority.Level); c != 0 {
+		return c
+	}
+	if c := cmp.Compare(x.priority.Weight, y.priority.Weight); c != 0 {
```
simonlechner, line 57:
```suggestion
	if c := x.priority.Cmp(y.priority); c != 0 {
```
No need to reimplement the priority comparison.

gossip/blockproc/priorities/ordering.go
```diff
@@ -0,0 +1,255 @@
...
+			delete(bySender, bestSender)
+		}
+	}
+	return selected
+}
+
+// combinePrioPrefixWithRemainder builds the final transaction order: the
+// prioritized entries in prefix order, followed by the remaining entries
+// (demoted + non-prioritized) in their original base order.
+func combinePrioPrefixWithRemainder(entries []transactionWithPriority, prioPrefixIndices []int) types.Transactions {
```
simonlechner: I know the function name is already very long, but I am not a fan of using `Prio` instead of `Priority` especially when we have a `Brio` flag.

gossip/blockproc/priorities/ordering.go
```diff
@@ -0,0 +1,255 @@
...
+// on it), so a budget is only ever spent on transactions that can actually
+// execute in the prioritized prefix. It consumes bySender, mutating it as
+// senders are advanced and exhausted.
+func computePrioritizedTxsPrefix(
+	txsWithPrio []transactionWithPriority,
+	bySender map[common.Address][]int,
+	perEntityBudget uint64,
+) []int {
+	selected := make([]int, 0, len(txsWithPrio))
+	remaining := make(map[[16]byte]uint64)
```
HerbertJordan: nit: pre-existing setup issue - it might be nice to introduce an `EntityID` type or something similar, instead of using the raw `[16]byte` in the code.

simonlechner (dismissed):
Just some minor comments.
The amount of code comments will be hard to keep up to date in future changes.

gossip/blockproc/priorities/ordering.go
```diff
@@ -0,0 +1,255 @@
...
+	budgetOf := func(id [16]byte) uint64 {
+		if r, ok := remaining[id]; ok {
+			return r
+		}
+		return perEntityBudget
+	}
+	for len(bySender) > 0 {
+		best := -1
+		var bestSender common.Address
+		for sender, sequence := range bySender {
```
HerbertJordan, line 214: An algorithmic comment: this loop nest is O(T*N) where T is the number of transactions, and N the number of senders, which in the worst case is O(T²); as long as the priority feature is not a heavily used feature, this should be fine. But if traffic would spike, let's say 1000 senders in one block, this loop can become expensive.

A potential ideas to reduce this: organize the candidates in a heap, sorted by `cmpLevelWeightHash`, to make locating the best next candidate cheap (O(1)); should get you in total to O(TlogN); but building a heap is a bit annoying when only using Go standard library features; Serse had a nice [Heap](https://github.com/0xsoniclabs/serse/blob/main/lib/heap/heap.go#L22) implementation, but maybe there is also some widely used library out there;

gossip/blockproc/priorities/ordering.go
HerbertJordan, line 1: Very nice code organization and scoping of isolated sub-problems!

gossip/blockproc/priorities/ordering.go
```diff
@@ -0,0 +1,255 @@
...
+			if n < expected {
+				continue // stale nonce: do not prioritize this transaction
+			}
+			if n > expected {
+				// nonce gap: do not prioritize this or any later transaction
+				// from this sender
+				break
+			}
+			sequence = append(sequence, idx)
+			expected++
```
HerbertJordan, line 179: An exotic case: This function might have a problem if the same nonce shows up more than once. If, for instance, you get for whatever reason this:
```
tx := []Transactions{
    Tx{Sender: A, Nonce: 1}, 
    Tx{Sender: A, Nonce: 2}, 
    Tx{Sender: A, Nonce: 2}, 
    Tx{Sender: A, Nonce: 3}, 
}
```
and all of them are prioritized, with an initial nonce of 1 in the store, then the second nonce 2 and the nonce 3 transaction is demoted to a regular transaction (by line 176). However, the second nonce 2 transaction is a no-op, and the Nonce: 3 transaction could be executed following it, enjoying priority.

Does the input to `Prioritize` already have guaranteed nonce-deduplication? Or is this something that should be considered here, for defensive coding?

gossip/blockproc/priorities/ordering.go
LorenzSchueler: Not quite. Priority.Cmp returns equal if level is 0 no matter the weight. If we use it here we would have a situation where if the level is 0 we ignore the weight and break ties only using the hash. This would be weird since this breaks the order level→weight→hash

HerbertJordan (dismissed):
Very nice and clean code, for the feature as well as for the tests. Great!

I added one suggestion for a named type and two potential issues that should be addressed. Please let me know what you think

gossip/blockproc/priorities/ordering.go
simonlechner: Right, then a unit test case is missing. I tried it locally before the review and no test failed.

gossip/blockproc/priorities/ordering.go
LorenzSchueler: Yesterday I also thought about this problem, and it can happen that we get duplicated nonces here. However, I came to the conclusion that this is NOT a problem.

When we process the second transaction with nonce 2, it is removed, however this nonce is smaller than expected so we just skip this one and not all remaining ones.
So we get Tx1, Tx2, Tx4 with prio and Tx3 demoted.

I'll add a test case for it.

gossip/blockproc/priorities/ordering.go
LorenzSchueler: This is true. If this is actually a problem depends on how many prioritized senders and transactions we allow. But I agree that it would be safer to have a better runtime bound.
The (not yet merged) emitter part already uses a heap, but the API is bad.
If you are fine with it, I would suggest changing this in a follow-up and using shared utilities between emitter and block processing.

gossip/blockproc/priorities/ordering.go
HerbertJordan: Right, I missed the continue in line 171.

HerbertJordan (dismissed):

gossip/blockproc/priorities/ordering.go
LorenzSchueler: Added.

simonlechner (approved):

PR #1089, opened 2026-07-28
Add counters for registry query failure

LorenzSchueler:
This PR adds counters to be able to monitor when queries to the priority registry fail.


gossip/c_block_callbacks.go
```diff
@@ -79,6 +79,11 @@ var (
 	skippedTxsMeter   = metrics.GetOrRegisterMeter("chain/txs/skipped", nil)
 	invalidTxsMeter   = metrics.GetOrRegisterMeter("chain/txs/invalid", nil)
 
+	priorityFailures = priorityFailureMeters{
```
LuisPH3, line 82: I need some context, can we add oneliners about what is being measured? just like in the chain/events counters

LuisPH3 (changes requested):

simonlechner (approved):

LuisPH3 (approved):

PR #1091, opened 2026-07-29
Add FrontierHeap and use it for priority ordering

LorenzSchueler:
This PR introduces `utils/frontier_heap`, a generic max-heap over the frontier (heads) of a set of sequences. Each sequence's next element joins the frontier only once the sequence is advanced.

API:
- `AddSequence` — insert a sequence, O(log s)
- `Peek` — greatest frontier head, O(1)
- `Shift` — take the greatest head and advance its sequence, O(log s)
- `PopSequence` — drop the whole sequence owning the greatest head, O(log s)

The priority ordering (`computePrioritizedTxsPrefix`) now uses this heap. This reduces the time complexity from O(n*s) to O(n log s) where n is the total number of transactions and s the number of distinct senders. For the worst case where all senders are distinct (s = n) this means going from O(n²) to O(n log n).

The selection order is unchanged.

### DIFF
--- a/gossip/blockproc/priorities/ordering.go
+++ b/gossip/blockproc/priorities/ordering.go
@@ -22,6 +22,7 @@ import (
 	"slices"
 
 	"github.com/0xsoniclabs/sonic/opera"
+	"github.com/0xsoniclabs/sonic/utils/frontierheap"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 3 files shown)


LuisPH3 (approved):
The changes replace search algorithm with heap structure, code is lean, testing is adequate.

PR #1092, opened 2026-07-29
Handle sigterm during node start/password reading

thaarok:
It is possible to use a unix pipe as a source of the validator keystore password.
The issue is, if the node receives SIGTERM while the node is waiting for the password from the pipe, the node terminates improperly and breaks the database. The same applies even without the pipe - when the SIGTERM is received during the startup phase of the node.

**This PR introduce SIGTERM handling by context for the whole duration for which the node keeps the database open.**

As a side effect, this removes the forced-termination by repeated Ctrl+C - we are not going to support termination without proper closing of the database.

Note: startFreeDiskSpaceMonitor moved from startNode() into lachesisMainInternal() - it makes canceling the context from it more straightforward and there is no logical reason for it to be part of the startNode() method.

Edit: The same fix was applied for reading a password from the console (GetPassPhrase).

### DIFF
--- a/cmd/sonicd/diskusage/monitor.go
+++ b/cmd/sonicd/diskusage/monitor.go
@@ -17,15 +17,14 @@
 package diskusage
 
 import (
-	"os"
-	"syscall"
+	"context"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
 )
 
-func MonitorFreeDiskSpace(stopNodeSig chan os.Signal, path string, freeDiskSpaceCritical uint64) {
+func MonitorFreeDiskSpace(ctx context.Context, stopNode context.CancelFunc, path string, freeDiskSpaceCritical uint64) {
 	ticker := time.NewTicker(60 * time.Second)
 	defer ticker.Stop()
 	for {
@@ -36,11 +35,15 @@ func MonitorFreeDiskSpace(stopNodeSig chan os.Signal, path string, freeDiskSpace
 		}
 		if freeSpace < freeDiskSpaceCritical {
 			log.Error("Low disk space. Gracefully shutting down Sonic to prevent database corruption.", "available", common.StorageSize(freeSpace))
-			stopNodeSig <- syscall.SIGTERM
+			stopNode()
 			break
 		} else if freeSpace < 2*freeDiskSpaceCritical {
 			log.Warn("Disk space is running low. Sonic will shutdown if disk space runs below critical level.", "available", common.StorageSize(freeSpace), "critical_level", common.StorageSize(freeDiskSpaceCritical))
 		}
-		<-ticker.C
+		select {
+		case <-ticker.C:
+		case <-ctx.Done():
+			return
+		}
 	}
 }
--- a/cmd/sonictool/app/account.go
+++ b/cmd/sonictool/app/account.go
@@ -70,7 +70,7 @@ func accountCreate(ctx *cli.Context) error {
 		scryptP = keystore.LightScryptP
 	}
 
-	passwordList, err := config.MakePasswordList(ctx)
+	passwordList, err := config.MakePasswordList(context.Background(), ctx)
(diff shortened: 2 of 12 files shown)


thaarok:
Updates:
* fixed sigterm handling during reading tx-signing keystore password file
* logging improvement - log before reading password file (it is not visible why is the sonic start stuck - that it is waiting for the password form the pipe), also removed misleading `Shutdown complete.` log message, after which the user still have to wait for the carmen database closing

cmd/sonictool/app/account.go
```diff
@@ -69,11 +70,11 @@ func accountCreate(ctx *cli.Context) error {
 		scryptP = keystore.LightScryptP
 	}
 
-	passwordList, err := config.MakePasswordList(ctx)
+	passwordList, err := config.MakePasswordList(context.Background(), ctx)
```
LuisPH3, line 73: this is only called on sonic tool, control+c is not that important here but it could be added.

cmd/sonictool/app/validator.go
```diff
@@ -46,11 +47,11 @@ func validatorKeyCreate(ctx *cli.Context) error {
 		return err
 	}
 
-	passwordList, err := config.MakePasswordList(ctx)
+	passwordList, err := config.MakePasswordList(context.Background(), ctx)
```
LuisPH3, line 50: same here, this could have ctl-c support now

cmd/sonictool/chain/import_events.go
```diff
@@ -65,7 +66,10 @@ func EventsImport(ctx *cli.Context, files ...string) error {
 	cfg.Node.P2P.StaticNodes = nil
 	cfg.Node.P2P.TrustedNodes = nil
 
-	node, svc, nodeClose, err := config.MakeNode(ctx, cfg)
+	sigCtx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
```
LuisPH3, line 69: nit: this context could be declared in the top level and shared with all subcommands.

cmd/sonicd/diskusage/monitor.go
```diff
@@ -36,11 +35,15 @@ func MonitorFreeDiskSpace(stopNodeSig chan os.Signal, path string, freeDiskSpace
...
-			stopNodeSig <- syscall.SIGTERM
+			stopNode()
 			break
 		} else if freeSpace < 2*freeDiskSpaceCritical {
 			log.Warn("Disk space is running low. Sonic will shutdown if disk space runs below critical level.", "available", common.StorageSize(freeSpace), "critical_level", common.StorageSize(freeDiskSpaceCritical))
 		}
-		<-ticker.C
+		select {
+		case <-ticker.C:
+		case <-ctx.Done():
```
LuisPH3, line 45: 💯  we have always wanted to change the code to stop this routine. thanks

LuisPH3 (approved):
This PR introduces a better way of handling signals which can reach more parts of the code.

The changes affect the signal handling during the whole sonic lifetime, this needs to be tested by killing running clients.

cmd/sonictool/app/account.go
thaarok: The account tool already supports Ctrl+C - because the database is not open there, standard Ctrl+C handling is sufficient. (There is no need to catch the signal.)

cmd/sonictool/app/validator.go
thaarok: The validator tool already supports Ctrl+C - because the database is not open there, standard Ctrl+C handling is sufficient. (There is no need to catch the signal.)

cmd/sonictool/app/account.go
thaarok: Keeping for a follow-up PR.

cmd/sonictool/app/validator.go
thaarok: Keeping for a follow-up PR.

PR #1094, opened 2026-07-31
Add priorityContext for emitter

LorenzSchueler:
This PR adds the `priorityContext` for the emitter. This context will be used in the next PRs to query the priority of transactions to be able to get them into events faster.
It caches the priority config so that the limits looked up when ever needed.


LuisPH3 (approved):
The priority context creates a transient object with evm and access to the state to be reused by all evaluations done within this one emission, and therefore using the same state.

PR #1095, opened 2026-07-31
Use newer version of lru cache which supports generics

LorenzSchueler:
This PR uses the `v2` version of the `golang-lru` package. Unlike the previous version is supports generics. This has two advantages:
- type safety - no more casting which can fail at runtime
- no interface boxing - fewer allocations

We already depended on the new version transitively, now it is a direct dependency. The old version is now only an indirect dependency.


LuisPH3 (approved):
The pr updates a dependency to reduce heap allocations.

3 code paths use this feature:
- Bundles:  tested in integration tests
- txs ring, part of the validator role, tested in integration tests + norma
- gasprice, part of the suggest gas price infrastructure, tested in integration tests.

PR #1097, opened 2026-07-31
Use newer version of lru cache which supports generics (#1095) [v2.2]

LorenzSchueler:


LuisPH3 (approved):

LuisPH3:
testing:
- bundles via integration (sonic, norma) 
- tx ring buffer, validator behavior:  integration sonic norma
- gasprice: suggest gas price is used during integration tests.

PR #1099, opened 2026-07-31
Make emitter candidate ordering priority aware

LorenzSchueler:
This PR makes the emitter's candidate ordering priority-aware and lets a validator eagerly include prioritized transactions that are not its own turn, so that prioritized transactions reach a block without waiting for their turn-owner to emit.

`transactionsByPriceAndNonce` becomes `transactionsByPriorityAndPriceAndNonce` and orders candidates by

    (stage asc, priority level desc, priority weight desc, effective tip desc, first-seen time asc)

where `stage` partitions candidates into prioritized-my-turn, prioritized-not-my-turn and not-prioritized. Per-sender nonce sequencing is still honored by the underlying `FrontierHeap`, so the emitted order is not monotonic: a promoted nonce is staged on its own metadata, not its predecessor's.

### DIFF
--- a/gossip/emitter/emitter.go
+++ b/gossip/emitter/emitter.go
@@ -43,6 +43,7 @@ import (
 	"github.com/ethereum/go-ethereum/metrics"
 
 	"github.com/0xsoniclabs/sonic/gossip/blockproc/bundle"
+	"github.com/0xsoniclabs/sonic/gossip/blockproc/priorities"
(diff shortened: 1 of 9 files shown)


LuisPH3 (approved):
The changes promote emission of prioritized transactions:
- emitters will emit transactions outside their turn if they are prioritized. 
- Throttler logic is respected by remaining silent if no own txs are to be emitted.

PR #1108, opened 2026-08-03
Assert the subsidy deduction is actually observed

LuisPH3:
Fixes https://github.com/0xsoniclabs/sonic-admin/issues/779

## Root cause: the test failed while the product was correct

The reported failure looks like a fund-accounting bug:

```
expected: 0xde0b6b3a7640000
actual  : 0xdd80977ee98ae0e
Messages: the sponsorship fund should be deducted by the expected amount
```

Decoding it says otherwise. `0xde0b6b3a7640000` is `1e18`, which is **exactly** `donation`. The assertion is

```go
require.Equal(t, donation.Uint64()-fundsDelta, fundsAfter, ...)
```

so an expected value equal to the full donation means `fundsDelta` was **zero**. The actual balance was lower, i.e. the on-chain deduction had happened correctly — the test simply never observed it. The subtest completed in 0.19s, consistent with a scan loop whose body never ran.

The scan window is

```go
for blockNumber := blockBefore.NumberU64() + 1; blockNumber <= blockAfter.NumberU64(); blockNumber++ {
```

where `blockAfter` is read from the `"latest"` block immediately after the scenario returns. The scenario only waits for a *receipt*, not for the head pointer to catch up. If the window therefore excludes the block holding the sponsored transaction, `subsidies.IsSponsorshipRequest` never matches, `fundsDelta` stays zero, and the final assertion degrades into comparing the donation against itself while reporting a misleading balance mismatch.

## The fix

**1. Require that a sponsorship request was actually observed.** Every one of the eight scenarios in this test sponsors at least one transaction, so observing none means the test did not measure the thing it asserts about. (The "contract creation cannot be sponsored" cases belong to a different test, so the guard is valid for all cases here.)

This also closes the matching **false negative**: as written, the assertion would have passed vacuously had no funds been deducted at all. The guard fixes both directions.

**2. Extend and re-read the window** while no sponsorship request has been seen, up to ten times at 100ms, so a lagging head pointer no longer decides the outcome. The blocks already exist by this point — the scenario has receipts — so this only waits for the head pointer to catch up, not for new blocks.

## Verification

Forcing an empty scan window replaces the misleading message with an accurate one:

```
no sponsorship request observed in blocks 6..5 after 0 window extension(s);
the fund assertion below would not be measuring anything
```

The balance mismatch from the report no longer appears, because the guard fires first.

```
go test ./tests/gas_subsidies -run TestGasSubsidies_SubsidizedTransaction_DeductsSubsidyFunds -count=1
  ok   github.com/0xsoniclabs/sonic/tests/gas_subsidies   19.986s
```

## Confidence, stated honestly

The two halves do not carry equal certainty, and reviewers should weight them differently:

- **The guard is certainly correct.** Whatever the cause, this test can no longer report a fund-accounting error when the real problem is that it measured nothing — nor pass when nothing was deducted.
- **The window re-read is a fix for an inferred cause.** The lagging-head mechanism follows from the arithmetic (`fundsDelta == 0`) and the 0.19s duration, but I did not observe it directly. If the flake recurs, the guard's message will say so precisely, and the window-extension count in that message will indicate whether re-reading helped.

## Scope

Test-only. One file: `tests/gas_subsidies/deducts_funds_test.go`. No product code, no shared test helpers.

---
*Prepared with Claude Code. The `fundsDelta == 0` conclusion is derived from the reported values; the underlying cause is inferred and labelled as such above.*

### DIFF
--- a/tests/gas_subsidies/deducts_funds_test.go
+++ b/tests/gas_subsidies/deducts_funds_test.go
@@ -19,6 +19,7 @@ package gas_subsidies
 import (
 	"math/big"
 	"testing"
+	"time"
 
 	"github.com/0xsoniclabs/sonic/evmcore"
 	"github.com/0xsoniclabs/sonic/gossip/blockproc/subsidies"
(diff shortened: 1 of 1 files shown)


simonlechner (approved):
Thanks for keeping comments at bay

PR #1114, opened 2026-08-05
Test that priority feature can be enabled and disabled

LorenzSchueler:
This PR adds a test ensuring the priority feature can be enabled and disabled. It also adds test utils which are used by this test. These utilities will be used by more test added in upcoming PRs.

### DIFF
--- a/tests/priorities/test_assertion_utils.go
+++ b/tests/priorities/test_assertion_utils.go
@@ -1,0 +1,155 @@
+// Copyright 2026 Sonic Operations Ltd
+// This file is part of the Sonic Client
+//
+// Sonic is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Sonic is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with Sonic. If not, see <http://www.gnu.org/licenses/>.
+
+package priorities
+
+import (
+	"math/big"
+	"testing"
+
+	"github.com/0xsoniclabs/sonic/tests"
+	"github.com/0xsoniclabs/sonic/utils/signers/internaltx"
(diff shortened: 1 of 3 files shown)


tests/priorities/test_assertion_utils.go
```diff
@@ -0,0 +1,155 @@
...
+// head observed just before the batch was submitted; that block is excluded
+// because it still holds the test's setup transactions (e.g. the funding of the
+// ordinary senders), which would otherwise count as ordinary traffic.
+//
+//   - expectPrioritized == true: all prioritized txs come before all ordinary
+//     ones (no ordinary tx precedes a prioritized one).
+//   - expectPrioritized == false: at least one ordinary tx precedes a
+//     prioritized one (proof that no reordering ran).
+//
+// Callers should submit the whole batch via SendAllToPool, so this global
```
LuisPH3, line 113: this is not true, it happens to be because the network only has one single emitter, it is always its turn and it can add up to 32 txs from every account into the events. 
the whole explanation is a little loose, it is true that the pool will delay things if txs are submitted one at the time, but it does not provide any guarantees neither.

tests/priorities/test_assertion_utils.go
```diff
@@ -0,0 +1,155 @@
...
+				}
+			} else {
+				sawOrdinary = true
+			}
+		}
+	}
+
+	allPrioritizedFirst := !ordinaryBeforePriority
+	require.True(sawPriority && sawOrdinary)
+	require.Equal(expectPrioritized, allPrioritizedFirst)
```
LuisPH3: this is true every time if the feature is enabled and working correctly. 
But it is stadistical if the feature is not enabled. I am afraid this could turn flaky very easily. there is is 1% (1 tx vs 10*10 ordinary txs)  chance to land up front. I do not know if this chance can be even less because the 10 consecutive nonces per sender .

Can we decrease the chance to land up front to something like 0.1 % ? 
is there any other way of testing this?

LuisPH3 (changes requested):

tests/priorities/test_assertion_utils.go
LorenzSchueler: In an ideal case you don't put in a single prio tx but multiple so the chance of all of them coming first is significantly lower. I could put a check that enough prio and non-prio transactions are used to have an extremely low chance of a false positive. And since the number of prio/ non prio tx depends on the test setup, this will fail immediately if the setup is broken, but will never be flaky.
WDYT?

LuisPH3 (approved):

PR #1118, opened 2026-08-05
Test that priority contract can be swapped

LorenzSchueler:
This PR adds a test ensuring the on-chain priority-registry implementation can be swapped out behind its proxy. The test verifies that the prioritization behavior changes accordingly.


LuisPH3 (approved):
kudos for the naming

PR #1121, opened 2026-08-06
Test validators are not slashed for emitting tx that are not their turn

LorenzSchueler:
This PR adds a test ensuring that if validators emit events that are not their turn, they do not get slashed for it. This is required for the eager emission of prioritized transactions. The test check that this is independent of whether the priority feature is enabled and whether the transaction is actually prioritized.


LuisPH3 (approved):
👍

PR #1122, opened 2026-08-06
Fuzz daily from this repository, rotating over its entry points

LuisPH3:
Moves Sonic's fuzzing out of LaScala and into this repository, where it is easier to maintain: the pipeline sits next to the code it fuzzes, and builds on the same `CI/Dockerfile.jenkins` the PR job already uses.

That is also why it has to live here. A Jenkins `dockerfile` agent is resolved against the *pipeline's own* SCM, so a pipeline in LaScala cannot use this image — the Go version the fuzzer ran on was the agent's, not the one this repository pins. Tosca's fuzzer broke that way when its toolchain moved into the image.

`CI/fuzzing.jenkinsfile` runs once a day and fuzzes one entry point, rotating over every `func FuzzXxx(f *testing.F)` it finds in its own checkout — so adding a fuzzer needs no change to this pipeline. `EntryPoint` is for running one by hand; empty means whichever has gone longest without a turn. The rotation and the fuzzing steps come from LaScala's shared library (0xsoniclabs/lascala#312, merged). There is no build stage: no cgo and no submodules, so `go test` compiles from a clean checkout.

One line worth not deleting: `GOCACHE` points into the workspace because `go test -fuzz` keeps its corpus in `$GOCACHE/fuzz`, and the container does not survive the build.

Unrelated, noticed while doing this: `FUZZING.md` is stale — it points at a `make fuzz` target that does not exist and describes the old `dvyukov/go-fuzz` workflow. Happy to fix here or separately.

Needs a Jenkins job pointed at this file; nothing runs until that exists.

### DIFF
--- a/CI/fuzzing.jenkinsfile
+++ b/CI/fuzzing.jenkinsfile
@@ -1,0 +1,166 @@
+// Copyright 2026 Sonic Operations Ltd
+// This file is part of the Sonic Client
+//
+// Sonic is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Sonic is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


simonlechner (approved):
Automatically picking up all fuzzers in a round-robin is a nice addition.

PR #1125, opened 2026-08-07
Test interaction between priority lanes and sponsorship and bundles

LorenzSchueler:
This PR adds 2 tests for the interacting between priority lanes and sponsorship and bundles.
- ensure a bundle inherits the priority of its envelope sender, not of the senders it encloses.
- ensure prioritizing a sponsee keeps each sponsored transaction paired with the internal transaction that settles its gas fee.

Alongside, the shared batch submission is extracted into `sendShuffledToPool`, which now shuffles before submitting so the observed ordering cannot depend on the order transactions were built in.

### DIFF
--- a/tests/priorities/feature_combinations_test.go
+++ b/tests/priorities/feature_combinations_test.go
@@ -1,0 +1,183 @@
+// Copyright 2026 Sonic Operations Ltd
+// This file is part of the Sonic Client
+//
+// Sonic is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Sonic is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with Sonic. If not, see <http://www.gnu.org/licenses/>.
+
+package priorities
+
+import (
+	"math/big"
+	"slices"
+	"testing"
+
+	"github.com/0xsoniclabs/sonic/gossip/blockproc/bundle"
+	"github.com/0xsoniclabs/sonic/gossip/blockproc/subsidies"
+	"github.com/0xsoniclabs/sonic/opera"
+	"github.com/0xsoniclabs/sonic/tests"
+	testbundles "github.com/0xsoniclabs/sonic/tests/bundles"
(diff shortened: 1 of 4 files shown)


tests/priorities/feature_combinations_test.go
```diff
@@ -0,0 +1,183 @@
...
+	const numBundles = 5
+	const txsPerBundle = 2
+
+	cases := map[string]struct {
+		prioritizeEnvelopeSender bool
+		expectPrioritized        bool
+	}{
+		"prioritized envelope sender": {prioritizeEnvelopeSender: true, expectPrioritized: true},
+		"prioritized bundle senders":  {prioritizeEnvelopeSender: false, expectPrioritized: false},
+	}
```
LuisPH3, line 54: this table test feels like trying to reuse code more than testing properties. please be descriptive on what is being tested: properties with expectations, this test is declarative on the domains being tested not the properties expected.

tests/priorities/feature_combinations_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// Verify that each sponsored tx is followed by its payment tx.
+	for _, tx := range sponsoredTxs {
+		receipt := receiptByHash[tx.Hash()]
+		block, err := client.BlockByNumber(t.Context(), receipt.BlockNumber)
+		require.NoError(err)
+		txs := block.Transactions()
+		require.Less(int(receipt.TransactionIndex)+1, len(txs))
+
+		payment := txs[receipt.TransactionIndex+1]
```
LuisPH3, line 177: 👍

tests/priorities/test_assertion_utils.go
```diff
@@ -97,6 +89,33 @@ func buildOrdinaryTraffic(
...
+// in the order given.
+func sendShuffledToPool(
+	t *testing.T,
+	net *tests.IntegrationTestNet,
+	txs []*types.Transaction,
+) []common.Hash {
+	t.Helper()
+
+	batch := slices.Clone(txs)
+	rand.Shuffle(len(batch), func(i, j int) {
```
LuisPH3: Random interactions in tests need to be seeded, otherwise test failures cannot be reproduced.

tests/priorities/test_assertion_utils.go
```diff
@@ -97,6 +89,33 @@ func buildOrdinaryTraffic(
 	return txs
 }
 
+// sendShuffledToPool adds all given transactions to the pool in a single batch
+// and in random order, so they all become available before the emitter builds
+// its first event and their relative ordering depends neither on the order they
+// were built in nor on their arrival times in the pool. It returns their hashes
+// in the order given.
+func sendShuffledToPool(
```
LuisPH3, line 97: i like this addition, but please consider message bellow

LuisPH3 (changes requested):

LuisPH3 (approved):

PR #1130, opened 2026-08-10
Test that txs are non-prioritized when registry misbehaves

LorenzSchueler:
This PR adds a test ensuring that a registry which misbehaves in different ways just causes all transactions to be non-prioritized, but does not crash the system.


LuisPH3 (approved):
good idea

PR #1131, opened 2026-08-11
Test that entity gas limit gets applied

LorenzSchueler:
This PR adds a test ensuring that the per entity per block gas budget gets applied. It submits one more prioritized transactions that fit into this budget, and checks that only the highest priority ones actually get the priority, and the lowest priority one is demoted and treated as non-prioritized.

### DIFF
--- a/tests/priorities/entity_gas_limit_test.go
+++ b/tests/priorities/entity_gas_limit_test.go
@@ -1,0 +1,110 @@
+// Copyright 2026 Sonic Operations Ltd
+// This file is part of the Sonic Client
+//
+// Sonic is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Sonic is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


LuisPH3 (commented):
demoted txs join the ordinary ones to be scrambled, what are the probabilities of this test being flacky because the tx end up first among ordinaries?

LorenzSchueler:
The test fails only if, in every block, all demoted txs sit before the first ordinary tx in the scrambler order.
The whole batch fits one event and one block. We have 30 prioritized txs from 30 senders, 6 per entity across 5 entities, of which 25 keep priority and 5 — the lowest-weight one of each entity — are demoted. Additionally, we have 100 ordinary txs from 20 senders.
For the test to fail, those 5 demoted txs would need to sit before all 100 ordinary txs. The scrambler respecting nonce order does not reduce this to the 20 heads of the distinct ordinary senders: it assigns a slot to every tx by hash XOR salt and then only reorders a sender's txs within the slots that sender already holds, so the first ordinary tx sits at the minimum over all 100 ordinary slots, not over 20.
For d demoted and m ordinary txs in a block the probability is 1/binomial(d+m, d), here 1/binomial(105, 5) = 1/96,560,646 ≈ 1.0e-8, i.e. 0.000001%.

LuisPH3 (approved):

PR #1132, opened 2026-08-11
Fix enr shutdown race condition

simonlechner:
A nightly race detection [run](https://scala.fantom.network/job/Sonic/job/Unit-test-race-detection/551) failed due to a race condition during node shutdown.

`Service.Stop` is strictly sequential. The concurrency is the background
goroutines started by `Service.Start`, still running alongside it. Each step
either *joins* its goroutine or merely *signals* it.

Before the fix, the ENR updater was signalled but never joined:

```
 Service.Stop()  ── one goroutine, top to bottom ──┐
                                                   │
 step                          │ feeder │ ENR updater │
 ────────────────────────────── ─────── ───────────── ┤
 txpool.Stop()                 │  ███   │     ███     │
 verWatcher.Stop()             │  ███   │     ███     │
 emitters[i].Stop()            │  ███   │     ███     │
 operaDialCandidates.Close()   │  ███   │     ███     │
 handler.Stop()                │  ███   │     ███     │
 filterAPI.Stop()              │  ███   │     ███     │
 feed.Stop():                  │        │             │
   close(stopFeeder) ─────────▶│ signal │     ███     │
   <-feederDone  ═══ JOIN ═════╡ exits  │     ███     │
   scope.Close() ──────────────────────▶ sub.Err() closes
 gpo.Stop()                    │        │     ███  ⚠  │
 tflusher.Stop()               │        │     ███  ⚠  │
 engineMu.Lock()               │        │     ███  ⚠  │
 blockProcWg.Wait()            │        │     ███  ⚠  │
 dagIndexer.Close()            │        │     ███  ⚠  │
 store.Commit()                │        │     ███  ⚠  │
 ── Stop returns ──────────────│        │     ███  ⚠  │
 p2p.Server.Stop()             │        │     ███  ⚠  │
 node.doClose()                │        │     ███  ⚠  │
 nodeClose() → Store.Close()   │        │     ███  ⚠  │
   MigrateCaches ✗WRITE ◀── DATA RACE ──▶ ✗READ GetBlock
```

The feeder gets a real join (`<-feederDone`). The ENR updater got only a signal
(`scope.Close()`), fired at the very end of `feed.Stop()`, with nothing waiting
for it afterwards. Its lifetime was unbounded — the `⚠` band was luck.

After the fix:

```
 filterAPI.Stop()              │  ███   │     ███     │
 stopENRUpdater():        ◀NEW │        │             │
   close(quit) ────────────────────────▶│   signal    │
   <-done  ═════════ JOIN ══════════════╡   exits     │
 feed.Stop()                   │  ███   │   (gone)    │
   ...                         │        │             │
 store.Commit()                │        │             │
 ── Stop returns ──────────────│        │             │
 nodeClose() → Store.Close()   │        │             │
   MigrateCaches ✗WRITE          nothing left to read ✓
```

## Why the join, not the signal

```
 BEFORE                               AFTER
 ──────                               ─────
 newHead: [b][b][b]…  cap 10          newHead: [b][b][b]…  cap 10

 select {                             select {
   case <-newHead:                      case <-newHead:
       store.GetBlock()   ⚠                 store.GetBlock()
   case <-sub.Err():                    case <-quit:     ◀ closed by Stop
       return                               return
 }                                      case <-sub.Err():
                                            return
 Both cases ready ⇒ Go picks a       }
 ready case at random. A queued
 block can win *after* sub.Err()     Still random — but Stop blocks
 closed, and nothing waits for       on <-done, so the loop is
 the loop to finish.                 provably finished before Stop
                                     moves on.
```

The select stays non-deterministic. Correctness comes from `stop()` blocking on
`<-done`, not from controlling which branch wins.

### DIFF
--- a/gossip/discover.go
+++ b/gossip/discover.go
@@ -17,10 +17,12 @@
 package gossip
 
 import (
+	"sync"
+
 	"github.com/Fantom-foundation/lachesis-base/inter/idx"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/forkid"
-	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/ethereum/go-ethereum/p2p/enr"
 	"github.com/ethereum/go-ethereum/rlp"
 
 	"github.com/0xsoniclabs/sonic/evmcore"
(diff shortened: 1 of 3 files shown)


LorenzSchueler:
Duplicate of #1129 and #1126

LuisPH3:
Overseen by #1129 

Reason, wg is much cleaner than channels.

PR #1133, opened 2026-08-11
Run the Ethereum execution-spec tests in this repo's build image

LuisPH3:
Moves the EEST (Ethereum execution-spec tests) pipeline out of [lascala](https://github.com/0xsoniclabs/lascala) and into this repository, as `CI/eth-exec-spec-tests.jenkinsfile`, so it can run inside `CI/Dockerfile.jenkins` — the same image the PR pipeline builds — instead of on a bare Jenkins agent with whatever toolchain that machine happens to carry.

Companion PR removing the old script: 0xsoniclabs/lascala (`luis/move-eest-pipeline-to-sonic`).

## What the pipeline does differently

- **`agent { dockerfile { filename 'CI/Dockerfile.jenkins' } }`** instead of `agent { label ... }`.
- **No checkout stage.** The job is configured as *Pipeline script from SCM* against `SonicRepository` / `SonicVersion`, so the checkout Jenkins performs to build the docker agent is already the revision under test — build image and code under test always come from one commit. `SonicRepository` and `SonicVersion` are still declared as parameters because the job configuration references them and the release-testing coordinator passes them.
- **`wget` → `curl -sSfL -O`.** The build image ships `git`, `curl` and `make`; there is no `wget` in it, so the download stage would have failed as-is.
- **`rm -rf fixtures` before extracting.** The workspace is reused and no longer gets a cleaning checkout, so a run that switched `TargetFork` would otherwise also run the previous fork's cases.

Stages, timeout, Slack channel, `GOMEMLIMIT` and the fixture versions per fork are unchanged.

## Jenkins job configuration required before merge takes effect

`/Sonic/Ethereum-exec-spec-tests` keeps its name, path and parameters — only its Pipeline definition changes:

- SCM: git, Repository URL `${SonicRepository}`, Branch specifier `${SonicVersion}`,
- Script path: `CI/eth-exec-spec-tests.jenkinsfile`
- Define all four parameters (`AgentLabel`, `SonicRepository`, `SonicVersion`, `TargetFork`) **in the job config too** — the SCM checkout happens before the pipeline's `parameters {}` block is parsed, so on the first build `${SonicVersion}` would otherwise expand to nothing.
- **Uncheck "Lightweight checkout"** — it is unreliable at expanding a parameterized branch specifier.

Worth a manual build with `SonicVersion` set to a 12-char commit before the coordinator points at it.


LuisPH3:
test: https://scala.fantom.network/blue/organizations/jenkins/Sonic%2FEthereum%20execution-spec%20tests/detail/Ethereum%20execution-spec%20tests/4/pipeline

simonlechner (approved):

PR #1142, opened 2026-08-14
Size the genesis state caches to the accounts a test net is given

LuisPH3:
This PR is needed by testing tools which create large batches of accounts during genesis.

A genesis is written as a single block, and Carmen panics rather than writing out a node whose hash it has not computed yet, so importing more accounts than the state caches hold takes the process down with `unable to store account node with dirty hash` at around 1536 accounts. Two caches were hardcoded to the minimum: the one the datadir is initialized with, and the genesis builder's own, which no flag reached.

`StartIntegrationTestNet` now derives all three sizes from the accounts it was handed, and the existing options still override it:

| accounts | live / archive | statedb.cache |
|---------:|---------------:|--------------:|
|      ≤ ~500 | 1 (minimum) | 1024 |
|     4096 | 4 nodes/account × 1200 B | 16384 |
| override | `LiveCacheSize` / `ArchiveCacheSize` | `CacheSize` |

A test with a handful of accounts runs on exactly the values it ran on before, so nothing existing pays for this. 32768 accounts import in about 1.5 s.

The builder takes the sizes as `makegenesis.CacheSizes`, which sonictool fills in from the flags that already size the import.


tests/integration_test_net.go
```diff
@@ -1565,6 +1560,41 @@ func validateAndSanitizeOptions(options ...IntegrationTestNetOptions) (Integrati
...
+// for a size only a few of them need. It is not enough for a genesis of thousands of accounts, though:
+// a genesis is written as one block, and Carmen panics rather than writing out a node whose hash it has
+// not computed yet, so the caches have to hold the whole genesis trie.
+func (o IntegrationTestNetOptions) cacheSizes() (live, archive int64, elements int) {
+	const (
+		// A node per account, plus headroom for the branch nodes above them.
+		nodesPerAccount = 4
+		// mpt.EstimatePerNodeMemoryUsage, which Carmen does not export. The byte sizes are divided by
+		// it to get a node count, so this is how a node count is turned back into bytes.
+		bytesPerNode = 1200
```
simonlechner, line 1576: Carmen exports this variable, it just can't be used as a constant.

tests/integration_test_net.go
```diff
@@ -1565,6 +1560,41 @@ func validateAndSanitizeOptions(options ...IntegrationTestNetOptions) (Integrati
...
+		// Carmen's own minimum node cache, which is what the minimum below really buys.
+		minNodes = 2000
+	)
+
+	live, archive, elements = 1, 1, 1024
+	if nodes := nodesPerAccount * len(o.Accounts); nodes > minNodes {
```
simonlechner, line 1582: Should storage slots also be accounted for here?

tests/integration_test_net.go
```diff
@@ -1565,6 +1560,41 @@ func validateAndSanitizeOptions(options ...IntegrationTestNetOptions) (Integrati
...
 }
 
+// cacheSizes are the state database cache sizes to run with: the caller's where given, otherwise the
+// minimum, raised to hold the genesis this net was handed.
+//
+// The minimum is the default because these caches are preallocated, so every test would otherwise pay
+// for a size only a few of them need. It is not enough for a genesis of thousands of accounts, though:
+// a genesis is written as one block, and Carmen panics rather than writing out a node whose hash it has
+// not computed yet, so the caches have to hold the whole genesis trie.
+func (o IntegrationTestNetOptions) cacheSizes() (live, archive int64, elements int) {
```
simonlechner, line 1570: Why not return a `makegenesis.CacheSizes` directly?

simonlechner (commented):
There are some cleanups that should be looked into. It would have been nice to see a test that requires these changes, to understand which changes are required.

tests/integration_test_net.go
ihtefyw: no, Carmen caches the node which contains a reference to the storage trie, not the actual slots. The byte per node should already include the storage trie pointer size

integration/makefakegenesis/json.go
```diff
@@ -238,7 +238,11 @@ func GenerateFakeJsonGenesis(
 	return jsonGenesis
 }
 
-func GetGenesisIdFromJson(json *GenesisJson, tmpDir string) (_ common.Hash, retErr error) {
+func GetGenesisIdFromJson(
+	json *GenesisJson,
+	tmpDir string,
+	caches ...makegenesis.CacheSizes,
```
ihtefyw, line 244: why is this a slice? Is it only to have an optional value?

tests/integration_test_net.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/1142/changes/27221f5b25a8e47e73218a8fdbebebf8b2b672e7

tests/integration_test_net.go
LuisPH3: https://github.com/0xsoniclabs/sonic/pull/1142/changes/93f09cc942401e7ee3561322124b1ef3a264ed60

integration/makefakegenesis/json.go
LuisPH3: this is a pattern used in this code region representing an optional parameter. It is used in other functions of the package.. I am not a big fan, but ill keep it for coherence

LuisPH3:
> There are some cleanups that should be looked into. It would have been nice to see a test that requires these changes, to understand which changes are required.

https://github.com/0xsoniclabs/sonic/pull/1142/changes/d617eec54d5c11c1c8bfe2fc5a29a7ec8155391c adds an integration test to check that genesis can have large numbers of accounts. without the proposed changes carmen just panics processing the update

simonlechner (dismissed):
Looks good from my side.

ihtefyw (dismissed):

simonlechner (approved):

PR #1143, opened 2026-08-14
Let a chain be replayed with each block's own base fee

LuisPH3:
`VerifyBlocks` replays every block with a base fee of zero, which lets sponsored transactions through checks they would otherwise fail. That is not faithful for a dynamic-fee transaction: `TransactionToMessage` prices one at `min(tip + baseFee, feeCap)`, so at zero the replay charges the tip alone, the sender keeps a different balance than the node computed, and the state roots diverge.

```
                     replay base fee  |  effective gas price
  sponsored tx            0 (default) |  passes the checks the node skipped
  dynamic-fee tx          0 (default) |  tip only        -> roots diverge
  dynamic-fee tx    UseBlockBaseFee   |  min(tip+bf,cap) -> roots match
```

`VerifyBlocksWithOptions` takes the header's own base fee instead. `VerifyBlocks` keeps its signature and its behaviour, so both existing callers are untouched; no in-tree caller sets the option yet.

### DIFF
--- a/tests/block_verifiability_utils.go
+++ b/tests/block_verifiability_utils.go
@@ -38,22 +38,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// VerifyBlocksOptions adjusts how a chain is replayed.
-type VerifyBlocksOptions struct {
-	// UseBlockBaseFee replays each block with the base fee recorded in its own
-	// header, instead of with a base fee of zero.
-	//
-	// Replaying with zero lets sponsored transactions through checks they would
-	// otherwise fail, which is why it is the default. It is not faithful for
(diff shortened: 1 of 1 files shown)


LorenzSchueler (changes requested):
When the live execution can run with a non-zero base fee, the replay must also be able to do so. The current fix does not work for a workload containing both ordinary dynamic fee tx and sponsored tx.

simonlechner (approved):
Originally, this test util was only used for sponsored transactions, which have a gasprice of 0. With bundles, it was moved to be used with all transactions types, but has only ever been used with legacy priced transactions (gasPrice == tipCap == feeCap).
The new version allows using it with modern transaction types.

LorenzSchueler (approved):

PR #1148, opened 2026-09-02
Drop the dead statedb argument from TxListenerModule.Start

LorenzSchueler:
DriverTxListenerModule.Start stored the state database but never read it, so removing the argument from the interface, its mock, the implementation, and all callers is behavior-preserving.


LuisPH3 (approved):
good cleanup

PR #1155, opened 2026-09-07
Build with debug symbols by default [v2.2]

LuisPH3:
Backport of #1154 to `v2.2`.

Removes `-s -w` from the `sonicd` and `sonictool` link flags, so the binaries we
build and ship keep their symbol table and DWARF debug information.

## Binary size

Measured on this branch (linux/amd64):

| binary | stripped (`-s -w`) | with symbols | delta |
|---|---|---|---|
| `sonicd` | 47.6 MB | 69.2 MB | +21.7 MB (+46%) |
| `sonictool` | 46.5 MB | 67.7 MB | +21.2 MB (+46%) |

The extra ~21 MB is metadata only: the symbol table and DWARF sections sit
alongside the code, they are not loaded into resident memory at run time and
they do not affect the generated machine code. Runtime performance is
unchanged; only the artifact on disk (and in the container image layer) grows.

## What this buys us

- **Readable stack traces.** `-w` strips DWARF, `-s` strips the symbol table.
  Without them, a crash dump or a core file from a node in the field gives us
  addresses instead of function names, and we cannot map them back without
  reproducing the exact build.
- **Working debuggers.** `dlv attach` / `dlv core` and `gdb` on a running or
  crashed node become useful: real frames, arguments, and local variables
  instead of `??`.
- **Usable profiles and traces.** `pprof` and the execution tracer resolve
  symbols from the binary, so CPU, heap, mutex and block profiles taken from a
  production node come back with function and line attribution rather than raw
  addresses. This is the difference between a profile we can act on and one we
  have to guess at.
- **Post-mortem on the machine that failed.** For an incident on a validator we
  do not control, being able to symbolize whatever the operator sends us,
  without asking them to rebuild, shortens diagnosis considerably.

Given that a node binary is deployed once and runs for weeks, trading ~21 MB of
disk for symbolized traces and attachable debuggers is worth it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,8 @@ sonicd:
 	GIT_DATE=`git log -1 --date=short --pretty=format:%ct 2>/dev/null || echo ""` && \
 	GOPROXY=$(GOPROXY) \
 	go build \
-	    -ldflags "-s -w -X github.com/0xsoniclabs/sonic/version.gitCommit=$${GIT_COMMIT} \
-	                    -X github.com/0xsoniclabs/sonic/version.gitDate=$${GIT_DATE}" \
+	    -ldflags "-X github.com/0xsoniclabs/sonic/version.gitCommit=$${GIT_COMMIT} \
+	              -X github.com/0xsoniclabs/sonic/version.gitDate=$${GIT_DATE}" \
 	    -o build/sonicd \
 	    ./cmd/sonicd && \
 	    ./build/sonicd version
@@ -19,8 +19,8 @@ sonictool:
 	GIT_DATE=`git log -1 --date=short --pretty=format:%ct 2>/dev/null || echo ""` && \
 	GOPROXY=$(GOPROXY) \
 	go build \
-	    -ldflags "-s -w -X github.com/0xsoniclabs/sonic/version.gitCommit=$${GIT_COMMIT} \
-	                    -X github.com/0xsoniclabs/sonic/version.gitDate=$${GIT_DATE}" \
+	    -ldflags "-X github.com/0xsoniclabs/sonic/version.gitCommit=$${GIT_COMMIT} \
+	              -X github.com/0xsoniclabs/sonic/version.gitDate=$${GIT_DATE}" \
 	    -o build/sonictool \
 	    ./cmd/sonictool && \
 	    ./build/sonictool --version


LorenzSchueler:
### CI failure here is unrelated to this PR — `v2.2` is missing the #1105 backport

*(This comment was generated by Claude, on request from @LorenzSchueler — findings below are from a code/history investigation, see the verified/hypothesis labels.)*

The build failed with `panic: test timed out after 30m0s`, blaming `TestPendingTransactionSubscription_ReturnsFullTransaction (25m21s)`. Nothing to do with the link flags in this PR.

**What actually happened (verified)**

The hung goroutine was parked here:

```
goroutine 146023 [chan receive, 25 minutes]:
github.com/0xsoniclabs/sonic/tests.TestPendingTransactionSubscription_ReturnsFullTransaction(...)
    tests/pending_transactions_subscription_test.go:59
```

On `v2.2`, line 59 is the unbounded `got := <-pendingTxs`. On `main` this wait is bounded by a 60s `select`, added in #1105 ("Bound the wait for pending-transaction notifications", merge `67c221666`). That commit is **not** reachable from `origin/v2.2`:

```
$ git branch -r --contains 67c221666 | grep v2.2
(no output)
```

So on `v2.2` a missed notification kills the entire `tests` package with a 30-minute panic instead of failing one test in 60s. Every other test in the dump is parked in `t.Parallel()` behind it, because this test runs serially.

Notably, execution got *past* `EthSubscribe` and `SendTransaction` — both returned no error. The subscription was live and the transaction was accepted by the node. Only the `newPendingTransactions` notification never arrived.

**Why the notification went missing (hypothesis)**

There is a race in `NewPendingTransactions` ([`gossip/filters/api.go:204-208`](https://github.com/0xsoniclabs/sonic/blob/v2.2/gossip/filters/api.go#L204-L208), identical on `main`):

```go
rpcSub := notifier.CreateSubscription()   // returns to the client here

go func() {
    incomingTxs := make(chan []*types.Transaction, 128)
    pendingTxSub := api.events.SubscribePendingTxs(incomingTxs)  // ...but only registers here
```

`SubscribePendingTxs` does block on `<-sub.installed`, but only once that goroutine gets scheduled. Between the client's `EthSubscribe` returning and the install completing, transactions entering the pool are dropped for this subscriber. The test sends its single transaction in exactly that window, and one missed event is unrecoverable — nothing else is ever sent on that channel.

This is labelled hypothesis: the window is real by inspection, but I have no server-side log proving that ordering for this specific run. It would be settled by node logs showing the tx entering the pool before `SubscribePendingTxs` installed. It fits a loaded 30-minute CI run, where scheduling delay for a fresh goroutine is far likelier than on an idle machine.

**Relation to existing issues**

- 0xsoniclabs/sonic-admin#752 (*Flaky TestPendingTransactionSubscription_ReturnsHashes*) is the same symptom on the sibling test — closed one second after #1105 merged, i.e. closed on the symptom, not the cause.
- 0xsoniclabs/sonic-admin#741 names this test under "possibly related", but is the wrong bucket: its mechanism is consensus failing to produce a block, whereas this notification fires on **pool acceptance**. Its mitigation #968 *is* on `v2.2` and could never have prevented this.

**Suggested follow-ups**

1. Backport `67c221666` (#1105) to `v2.2` — bounds the damage from 30 min to 60 s.
2. Move the `SubscribePendingTxs` call out of the goroutine in `gossip/filters/api.go` so the feed is installed before the RPC returns. This is the actual fix, and it applies to `main` too — where #1105 currently only converts the hang into a flaky 60s failure.

A re-run should get this PR green.

simonlechner (approved):

PR #1160, opened 2026-09-07
Fix HistoryStorage pre-block system calls in tracing

thaarok:
Fix initialization of the block state for all tracing methods:
* `debug_traceBlock*` and `trace_block` missed HistoryStorage pre-block system calls
* `trace_transaction` missed HistoryStorage pre-block system calls **only when tracing the first tx** of a block

Affects only transactions/blocks which reads hash of the block, who is parent of the current block, from the HistoryStorage contract - the read returns a different value while tracing.

Also fixed following minor issues:
* Redundant statedb.EndTransaction() calls in tracing (not affecting the result)
* Handling TxAsMessage err in traceBlock (should not occur in prod, but better safe)


api/ethapi/api.go
```diff
@@ -2714,7 +2741,7 @@ func (api *PublicDebugAPI) traceBlock(ctx context.Context, block *evmcore.EvmBlo
 		results[i] = &txTraceResult{TxHash: tx.Hash(), Result: res}
 		resultsLength += len(res)
 
-		statedb.EndTransaction()
+		// Note: statedb.EndTransaction() already included in ApplyTransactionWithEVM
```
jenikd: Instead of having this comment, I suggest deleting this

api/ethapi/tx_trace.go
```diff
@@ -567,8 +567,8 @@ func (s *PublicTxTraceAPI) traceTx(
 	chainConfig := s.b.ChainConfig(idx.Block(block.Number.Uint64()))
 	resultReceipt, err := evmcore.ApplyTransactionWithEVM(msg, chainConfig, core.NewGasPool(msg.GasLimit), statedb, block.Number, block.Hash, tx, &index, tracedEVM.vmenv)
 
+	// Note: statedb.EndTransaction() already included in ApplyTransactionWithEVM
```
jenikd: Also this note is not needed

jenikd (commented):
Thanks for this fix

LuisPH3 (dismissed):
I appreciate the testing effort. 
@jenikd  candidate for v2.2.3 (with other rpc fixes)?

jenikd:
> I appreciate the testing effort. @jenikd candidate for v2.2.3 (with other rpc fixes)?

Yes

jenikd (dismissed):

LuisPH3 (dismissed):

PR #1162, opened 2026-09-09
Install subscription feeds before the subscribe RPC returns

simonlechner:
`NewPendingTransactions` and `NewHeads` told the client it was subscribed before registering the feed with the event system, so every event in that window was silently dropped — unrecoverably for `newPendingTransactions`, where a client
that subscribes and sends a single transaction waits forever, which is the cause of the intermittent hangs in the pending-transaction subscription tests. This installs the subscription before returning, the same change go-ethereum has since made upstream in both methods.

### DIFF
--- a/gossip/filters/api.go
+++ b/gossip/filters/api.go
@@ -201,12 +201,17 @@ func (api *PublicFilterAPI) NewPendingTransactions(ctx context.Context, fullTx *
 		return &rpc.Subscription{}, rpc.ErrNotificationsUnsupported
 	}
 
-	rpcSub := notifier.CreateSubscription()
+	// The subscription must be installed in the event system before this
+	// function returns. Otherwise the client's subscription request completes
+	// while the feed is not yet registered, and transactions entering the pool
+	// in that window are never reported to this subscriber.
+	var (
+		rpcSub       = notifier.CreateSubscription()
+		incomingTxs  = make(chan []*types.Transaction, 128)
+		pendingTxSub = api.events.SubscribePendingTxs(incomingTxs)
+	)
 
 	go func() {
-		incomingTxs := make(chan []*types.Transaction, 128)
-		pendingTxSub := api.events.SubscribePendingTxs(incomingTxs)
-
 		for {
 			select {
 			case txs := <-incomingTxs:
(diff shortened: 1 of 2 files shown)


gossip/filters/api.go
```diff
@@ -201,12 +201,17 @@ func (api *PublicFilterAPI) NewPendingTransactions(ctx context.Context, fullTx *
...
-	rpcSub := notifier.CreateSubscription()
+	// The subscription must be installed in the event system before this
+	// function returns. Otherwise the client's subscription request completes
+	// while the feed is not yet registered, and transactions entering the pool
+	// in that window are never reported to this subscriber.
+	var (
+		rpcSub       = notifier.CreateSubscription()
+		incomingTxs  = make(chan []*types.Transaction, 128)
+		pendingTxSub = api.events.SubscribePendingTxs(incomingTxs)
+	)
```
LorenzSchueler: nit: I thought the convention is to use `:=` when assigning and only `var` when using the zero value?
Feel free to ignore this.

```suggestion
	rpcSub := notifier.CreateSubscription()
	incomingTxs := make(chan []*types.Transaction, 128)
	pendingTxSub := api.events.SubscribePendingTxs(incomingTxs)
```

gossip/filters/api.go
```diff
@@ -272,12 +277,15 @@ func (api *PublicFilterAPI) NewHeads(ctx context.Context) (*rpc.Subscription, er
...
 	}
 
-	rpcSub := notifier.CreateSubscription()
+	// Installed before returning, for the same reason as in
+	// NewPendingTransactions.
+	var (
+		rpcSub     = notifier.CreateSubscription()
+		headers    = make(chan *evmcore.EvmHeaderJson)
+		headersSub = api.events.SubscribeNewHeads(headers)
+	)
```
LorenzSchueler: Same here.

LuisPH3:
> the same change go-ethereum has since made upstream in both methods.

can you reference the commit/pr?

LuisPH3 (approved):
candidate for 2.2.3 ? @jenikd wdyt?

simonlechner:
> > the same change go-ethereum has since made upstream in both methods.
> 
> can you reference the commit/pr?

https://github.com/ethereum/go-ethereum/pull/33990

jenikd:
> candidate for 2.2.3 ? @jenikd wdyt?

Yes, this should be included

jenikd (approved):
Looks good, thanks for this

PR #1175, opened 2026-09-16
Fix integration tests flakiness when waiting for receipt

jenikd:
## Problem

  `TestBundles_BundleOnlyTransactionsAreAcceptedByTxPoolButNeverExecutedStartingFromBrio` failed sporadically in CI with:  
  failed to estimate gas for transaction: insufficient funds for transfer

  The failing call is `eth_estimateGas` from an account that was just funded with `MakeAccountWithBalance`. DoEstimateGas` returns this error when `value >= balance`, which also holds for value 0 and balance 0, so the node saw the new account with a zero balance.

  ## Cause

  Receipt helpers in `tests` (`GetReceipt`, `Run`, `RunAll`, `EndowAccount`, `MakeAccountWithBalance`, ...) returned as soon as `eth_getTransactionReceipt` succeeded. In `gossip/c_block_callbacks.go`, block processing stores the transaction index, receipts and the block before it calls `SetBlockEpochState`, which is what `GetLatestBlockIndex` reads. Between these two steps a receipt is visible while `"latest"` still resolves to the previous block. Any state-dependent RPC call at `"latest"` issued in that window observes the state before the transaction.
  
  The window is small, which is why the failure only shows up on loaded CI machines. Inserting a short sleep between `SetBlockIndex` and `SetBlockEpochState` reproduces the failure deterministically.

  ## Fix

  `Session.TryGetReceipts` now waits, after each receipt, until the node reports the receipt's block (or a later one) as its latest block and serves the state of that block (`waitUntilBlockStateIsLatest`). All helpers returning receipts are built on this function, so they now guarantee that subsequent `"latest"` queries include the transaction.
  
  ## Tests
  
  - `TestWaitUntilBlockStateIsLatest_*`: table tests of the probe against an in-process fake RPC server, covering immediate availability, lagging latest block, missing state, both combined, and timeout / cancellation error reporting.
  - `TestIntegrationTestNet_StateIsAvailableWhenReceiptIsReturned`: table test over all receipt helpers, checking balance at the receipt block, balance at latest, and gas estimation from the new account. Fails with the original error when the wait is disabled and the node delays the latest block update.
  - `testIntegrationTestNet_CanEndowAccountsWithTokens`: removed the `WaitForProofOf` workaround, the test now verifies state visibility directly after the receipt.


LuisPH3 (approved):

LuisPH3 (approved):
nice tool

PR #1176, opened 2026-09-16
evmcore: skip invalid authorizations in txpool authority index

LuisPH3:
Supersedes #1150 by @0xrlawrence, whose commit is included unchanged with original authorship. This PR adds a reproducing test on top.

### Problem

`txLookup.addAuthorities` iterated the raw authorization list and discarded the error from `Authority()`, so an authorization with an unrecoverable signature was indexed under the zero address. `removeAuthorities` iterates `SetCodeAuthorities()`, which yields recovered authorities only, so that zero-address entry was never removed. Pool validation does not check authorization signatures, so such a transaction is accepted by the pool and the entry leaks for the lifetime of the pool once SetCode transactions are enabled.

### Changes

- Skip authorizations that do not recover to an authority in `addAuthorities`, making it symmetric with `removeAuthorities` (original fix from #1150).
- Add `TestSetCodeTransactions_InvalidAuthorityIsUntrackedOnRemove`: submits a SetCode tx with a forged authorization signature, removes it, and asserts the authority index is empty. Fails without the fix, passes with it.
- Make `validateTxPoolInternals` skip unrecoverable authorizations as well. It previously used the same discarded-error lookup, so with the fix alone it would report a false "authority not tracked" for any pooled tx carrying an invalid authorization.


simonlechner (approved):

0xrlawrence:
👍

PR #1178, opened 2026-09-17
Fix single block proposer stall

simonlechner:
In single-proposer mode, block height must only advance on a confirmed proposal — producing an empty block without one invalidates the proposal still travelling through consensus, which arrives a block too late and leaves the next block proposal-less too. Under sustained load this feeds back on itself and starves the chain of transactions; the fix skips such blocks (unless there are cheaters to deactivate), while empty-block pacing on an idle network is unaffected.


simonlechner:
This is not an issue in production, the issue is on the test side setting the `MaxEmptyBlockSkipPeriod` to 4 seconds.

PR #1179, opened 2026-09-17
Add the test tooling Faust builds against

LuisPH3:
[Faust](https://github.com/0xsoniclabs/Faust) is the transaction property test that moved out of this tree. It builds against the `tests` package of this repository, and two things it imports never left the branch it grew up on (`luis/rapid_testing_txs`). Every PR in the Faust stack currently fails build and lint on the first of them.

This adds both, each with a test:

- `tests/contracts/nonce_priority_registry`: the registry contract Faust's priority workload deploys and drives. A self-contained `.sol` with its generated binding and `gen.go`, same shape as the sibling `misbehaving_priority_registry`. `TestPriorities_NonceRegistry_PrioritizesTheRegisteredNonceSlotsOnly` installs it through the proxy, checks it answers slot by slot what was registered, and checks the effect: a sender's registered stretch of nonces is scheduled ahead of ordinary traffic and the same sender's next, unregistered stretch is not.
- `IntegrationTestNet.ExportGenesis(path)`: stops the net and writes the genesis it started from plus every block since to a genesis file through `sonictool genesis export`, so a failing Faust run can be replayed. `TestIntegrationTestNet_ExportGenesis_WritesAChainANodeCanStartFrom` exports after a few blocks, checks the net is left stopped, wipes the node's state, imports the file and checks the same blocks and balances come back.

Verified by building and vetting Faust against `main` plus this change, which passes; against `main` alone it fails on exactly these two symbols.

Not included: the state cache sizing Faust's 4096-account pool needs at runtime. That is #1142 and stays there.


LuisPH3:
faust adopted the changes, no work needed in sonic

PR #1185, opened 2026-09-21
Adapt Sonic to `Carmen::StagedBlock` interface

ihtefyw:
This PR updates Carmen to https://github.com/0xsoniclabs/carmen/commit/432872a40380b536e0635d3624a16f00b334bdf7.
It introduces the `StagedBlock`, a new system which allows each block to be either committed or rolled back.

The main change is the `EndBlock` interface, which is now returning `(StagedBlock, error)`. The main user is the block processor, which currently commits the block right away as no revert logic is implemented in Sonic.

Another important change is included here: the `EndBlock` adapter now returns an error in case the `StateDB` is non-committable, e.g. it cannot be written. Before this was simply ignored, which is now complicated by the handling of the `StagedBlock`. Because calling a state-changing operation on a non-committable stateDB is a misuse, it is now threated as an error.


gossip/evmstore/carmen.go
```diff
@@ -395,11 +395,15 @@ func (c *CarmenStateDB) BeginBlock(number uint64) {
 }
 
 func (c *CarmenStateDB) EndBlock(number uint64) (carmen.StagedBlock, error) {
+	if !c.committable {
+		// only the live, committable StateDB seals blocks; a non-committable
+		// view has nothing to finalize, so reaching this point is a misuse.
+		return nil, fmt.Errorf("called EndBlock on a non-committable StateDB")
```
LuisPH3, line 401: this makes a little more sense than before. we need to be careful now with the uses. type change should have exposed them

gossip/blockproc/evmmodule/evm.go
```diff
@@ -185,23 +185,41 @@ func (p *OperaEVMProcessor) Finalize() (evmBlock *evmcore.EvmBlock, numSkipped i
 
 	evmBlock = p.evmBlockWith(transactions)
 
+	// The state root is always taken from the live state, even if committing
+	// the block fails below, so the caller never receives a zero root.
+	defer func() {
```
LuisPH3: this is.... 😨 

modifying on a defer a captured  reference to the return of a member....

can  we just not do this cleaner?

gossip/blockproc/evmmodule/evm.go
```diff
@@ -185,23 +185,41 @@ func (p *OperaEVMProcessor) Finalize() (evmBlock *evmcore.EvmBlock, numSkipped i
...
+		evmBlock.Root = p.statedb.GetStateHash()
+	}()
+
 	// Commit block
-	done := p.statedb.EndBlock(evmBlock.Number.Uint64())
-	// Use asynchronous commit for blocks older than one hour to speed up catching up.
-	// For recent blocks (within the last hour), wait for the commit to complete
+	stagedBlock, err := p.statedb.EndBlock(evmBlock.Number.Uint64())
+	if err != nil {
+		log.Error("Failed to finalize block %v: %v", evmBlock.Number, err)
```
LuisPH3: this makes sense but it has implications, it changes semantics right?
in the core this may be alright. how does this behave in the rpc? is it used there?

gossip/evmstore/carmen.go
```diff
@@ -412,6 +412,32 @@ func (c *CarmenStateDB) EndBlock(number uint64) <-chan error {
...
+type blockEnder interface {
+	EndBlock(number uint64) (carmen.StagedBlock, error)
+}
+
+// EndBlockAndCommit ends the given block, commits it to the archive and waits
+// for the archive update to complete. A block that is ended but never committed
+// is rolled back when the state is closed, so every EndBlock needs a matching
+// Commit. This helper is meant for uses without a reason to pipeline blocks,
+// such as genesis construction and tests; block processing commits itself.
+func EndBlockAndCommit(db blockEnder, number uint64) error {
```
LuisPH3: i think db could just have been the original interface stateb,   blockEnder is not needed.
the function prototype would read better (no body knows what a block ender is)

LuisPH3 (commented):
thanks for the commits, we need to discuss a couple of details and check the rpc side of evm.go 
ping me again after the copilot review  is resolved

gossip/evmstore/carmen.go
ihtefyw: Unfortunately it is needed, sonic defines its own `StateDB` interface using Ethereum types, so they are not compatible. Only a couple of them are compatible, in this case `EndBlock`

gossip/blockproc/evmmodule/evm.go
ihtefyw: We can put it at the end, but the issue is another: Finalize() does not return any error, only logs the issue. Technically you should ignore everything that happened after an error (e.g. the GetStateHash call) but there is no error to return...
I am not sure how this works in sonic, is it common in the client?

PR #1186, opened 2026-09-22
Cap the genesis node caches that OOM the race-detection nightly

LuisPH3:
The race-detection nightly failed with `signal: killed` on the `gossip` package. That is the OOM killer, not a race report — no race was ever detected.

## Cause

Two node caches are opened with `Capacity: 0`, which Carmen reads as "use the default" and sizes for 10M nodes — an owners slice plus an index map, ~600MB per opened trie:

- `evmstore.Store.ImportArchiveWorldState` passes `Cache.StateDbCapacity` through as the node cache capacity, and `MemTestStoreConfig` never set it. Carmen opens both an archive and a live trie with it, so that is twice.
- `mptIo.ExportArchive` hardcodes an empty `NodeCacheConfig`. Carmen's own *live* export uses a capacity of 2000; the archive path looks like an oversight.

So every test store that applies a fake genesis allocated ~1.8GB. Under `-race` the detector shadows all of it, and with `-p`/`-parallel` both defaulting to `GOMAXPROCS` those runs multiply.

`GOMEMLIMIT` could not have caught this: it bounds the Go heap, while the shadow memory is mapped by the TSan runtime outside it.

## Measured, `-race`, peak RSS

| | before | after |
|---|---|---|
| `TestConsensusCallback_AppliesTransactionPriorities` | 3465 MB (3.16s) | 412 MB (0.12s) |
| the parallel group from the failing log | 19797 MB | 1578 MB |
| whole `gossip` package | — | 2096 MB, PASS |

## Notes

- The export fix is not test-only. Exporting is a read-only sequential scan, so the minimum capacity is enough wherever a genesis is built, `sonictool` included.
- `evmstore.DefaultStoreConfig` is deliberately untouched: a real node importing a genesis with a 1940MiB cache can afford the default slots, and `tests/` already passes `--statedb.cache`.
- The nightly's own parallelism is capped separately in lascala (`-p 2 -parallel 2`), matching what carmen's race pipeline already does.


ihtefyw (approved):

ihtefyw (commented):
This may change the archive export performance.
It's better to benchmark the archive export with and without those changes.

ihtefyw (changes requested):
This may change the archive export performance.
It's better to benchmark the archive export with and without those changes.

LuisPH3:
> This may change the archive export performance. It's better to benchmark the archive export with and without those changes.

i agree, the nighties seem to behave just by reducing parallelism, maybe we do not need this change.

PR #1187, opened 2026-09-22
Add check for updated mocks in Jenkinsfile

ihtefyw:
This adds a check to the Jenkinsfile that each mock has been updated to avoid cases where people forgets to run the mock generation before committing.

`mockgen` has been added as a tool in go.mod, which requires the generator strings to be updated.


LuisPH3 (changes requested):
please add tools to the go.mod file, https://go.dev/doc/modules/gomod-ref#tool

then use `go mod download` in the pipeline instead of in the dockerfile. 
this keeps versioning of tools in one place

sonicbuildbot (approved):

LuisPH3 (approved):

PR #1188, opened 2026-09-22
Reject `EndBlock` on non-committable StateDB

ihtefyw:
EndBlock is only meaningful for the live, committable `StateDB`. The non-committable views handed to the tx pool, the emitter and the RPC backend never seal blocks, so calling `EndBlock` on them is a misuse. 
This PR adds an early return in case a call to `EndBlock` in a non-committable state is detected.

The tests that exercised EndBlock on non-committable state are narrowed to the committable case, and the non-committable test now expects the error.

This is in preparation for the Carmen version update, which includes changes to the `EndBlock` interface which clashes with the behaviour of `EndBlock` in Sonic.


ihtefyw:
Close as we included the changes in #1185

PR #1189, opened 2026-09-23
Restore bundle history hash on genesis import without bundles

jmpike:
A newly created genesis can contain a non-zero bundle history but no retained bundles. `ApplyGenesis` skips the import whenever there were no bundles, so the history hash stays zero. The hash is part of the epoch state via `EpochEndExecutionPlanChainHash`. The node initialized without the history hash would synchronize blocks normally up to the epoch sealing and then diverges.


gossip/apply_genesis.go
```diff
@@ -128,81 +128,132 @@ func (s *Store) ApplyGenesis(g genesis.Genesis) (err error) {
...
-			s.Log.Info("No processed bundles in genesis, skipping import")
-			return nil
-		}
-		s.Log.Info("Importing processed bundles from genesis", "count", len(bundlesByBlock))
+	switch {
+	case len(bundlesByBlock) > 0 && hasHistory:
+		return s.replayProcessedBundles(historyHashes, bundlesByBlock)
+	case len(bundlesByBlock) > 0:
+		s.Log.Crit("Bundles were processed but no history hash was found in genesis")
+		return errors.New("bundles were processed but no history hash was found in genesis")
```
thaarok, line 154: `s.Log.Crit` terminates the processing, if you want to return the err, you should remove the `s.Log.Crit`

gossip/apply_genesis.go
```diff
@@ -128,81 +128,132 @@ func (s *Store) ApplyGenesis(g genesis.Genesis) (err error) {
...
+	case hasHistory:
+		// the history hash is updated for every block even if no bundles are retained,
+		// we need to restore it to produce correct epoch state hash at the epoch sealing
+		return s.restoreBundleHistoryHash(historyHashes.Latest)
+	default:
+		s.Log.Info("No processed bundles in genesis, skipping import")
```
thaarok, line 160: This log is not up-to-date, since it is "No processed bundles or bundle history hash in genesis" now.

gossip/apply_genesis.go
```diff
@@ -128,81 +128,132 @@ func (s *Store) ApplyGenesis(g genesis.Genesis) (err error) {
...
+		return nil
+	}
+}
```
thaarok, line 163: I don't like we need a switch with 4 branches for such trivial thing. I would prefer something like:
```
if len(bundlesByBlock) > 0 || hasHistory {
		if !hasHistory {
				return err...
		}
		return s.replayProcessedBundles(historyHashes, bundlesByBlock)
}
```
When bundlesByBlock are empty, replayProcessedBundles should be still able to handle it - initialize to the oldest hash, replay nothing (empty bundles slice), check that resulting hash matches with the newest one (since it is still the same value).
I think the special branch for hasHistory but no bundles is an unnecessary complication.

jmpike: "I think the special branch for hasHistory but no bundles is an unnecessary complication."
It's not. This is literally the reason why we have the bug.

I think four cases are way easier to read than nested branching across several IFs.

gossip/apply_genesis.go
LuisPH3: i would go with the return, i do not know what do we have locked in the context, but crit tends to interrupt concurrent operations or leave resources unreleased (lock files?)
since we return an error nevertheless, this is a no-brainer

gossip/apply_genesis.go
LuisPH3:
```suggestion
		s.Log.Info("No processed bundles in genesis, skipping bundle history import")
```
or no log at all

gossip/apply_genesis.go
```diff
@@ -128,81 +128,132 @@ func (s *Store) ApplyGenesis(g genesis.Genesis) (err error) {
...
+		"oldestBlockNum", historyHashes.Oldest.BlockNumber, "oldestHash", historyHashes.Oldest.Hash,
+		"latestBlockNum", historyHashes.Latest.BlockNumber, "latestHash", historyHashes.Latest.Hash)
+
+	// replay blocks to latest and add retained bundles
+	// the execution plan chain is updated block by block
+	startBlock := s.initBundleHistoryReplay(historyHashes)
+	for block := startBlock; block <= historyHashes.Latest.BlockNumber; block++ {
+		bundlesPerBlock, err := positionsByExecutionPlan(bundlesByBlock[block])
+		if err != nil {
+			s.Log.Crit("Invalid processed bundles in genesis", "block", block, "err", err)
```
LuisPH3: same as before. lets avoid crit if we can return an error.

LuisPH3 (commented):
The changes add support for loading a g-file where the accumulated bundles-history hash is not zero while the known history contains zero bundles. 
the changes add clear handling of this situation and refactors the bundle load code.

thanks for the refactor.

some comments:

gossip/apply_genesis.go
jmpike: The log message updated

jmpike: Wording updated + reflected another comment below.

gossip/apply_genesis.go
jmpike: I agree, Log.Crit() removed.

LuisPH3 (approved):

